PR #7, opened 2022-01-11
Fixed heztner cloud link

Julian-Louis:


mysticaltech:
Thanks @Julian-Louis, good catch!

PR #8, opened 2022-01-12
Write server ip instead of local ip

Julian-Louis:
[https://github.com/kube-hetzner/kube-hetzner/blob/master/master.tf#L47](https://github.com/kube-hetzner/kube-hetzner/blob/master/master.tf#L47) edit localhost ip by server ip, but [here](https://github.com/kube-hetzner/kube-hetzner/blob/master/master.tf#L54-L57), the ip is edited by localhost, so I need to overwrite it
image

### DIFF
--- a/master.tf
+++ b/master.tf
@@ -51,6 +51,7 @@ resource "hcloud_server" "first_control_plane" {
   # Install Hetzner CCM and CSI
   provisioner "local-exec" {
     command = <<-EOT
+      sed -i -e 's/127.0.0.1/${self.ipv4_address}/g' ${path.module}/kubeconfig.yaml
       kubectl -n kube-system create secret generic hcloud --from-literal=token=${var.hcloud_token} --from-literal=network=${hcloud_network.k3s.name} --kubeconfig ${path.module}/kubeconfig.yaml
       kubectl apply -k ${path.module}/hetzner/ccm --kubeconfig ${path.module}/kubeconfig.yaml
       kubectl -n kube-system create secret generic hcloud-csi --from-literal=token=${var.hcloud_token} --kubeconfig ${path.module}/kubeconfig.yaml


mysticaltech:
@Julian-Louis, this is already done at line 47 above in the masters.tf file.
ksnip_20220113-100405

Julian-Louis:
Yes, I know, and it works, but when [this command](https://github.com/kube-hetzner/kube-hetzner/blob/master/master.tf#L54) is running, the IP is edited back to 127.0.0.1 and the command failed, it's weird

mysticaltech:
I understand it may seem that way, but it wouldn't connect if it was. If I am really missing something, do not hesitate to try to prove it (loom.com?).

Julian-Louis:
I'll do it as soon as possible

dhoppe:
@Julian-Louis Do you use macOS? The command `sed` behaves differently at macOS than at Linux. 

At macOS the parameter `-i` should look like `-i ''`, otherwise it will create a backup file named `kubeconfig.yaml-e`.

Julian-Louis:
@dhoppe, I'm currently using windows, but the commands works, but just after, the IP come backs to localhost

master.tf
```diff
@@ -51,6 +51,7 @@ resource "hcloud_server" "first_control_plane" {
   # Install Hetzner CCM and CSI
   provisioner "local-exec" {
     command = <<-EOT
+      sed -i -e 's/127.0.0.1/${self.ipv4_address}/g' ${path.module}/kubeconfig.yaml
```
dhoppe, line 54: This should not be necessary, because the IP address has already been changed at line 47 and Terraform will keep and re-use this config.

Julian-Louis: I agree with you, the ip is edited to my server IP, but then the IP is edited back to 127.0.0.1 and it won't work @dhoppe

Julian-Louis:
I just retried, and i have the same issue:
image

mysticaltech:
Really weird @Julian-Louis, we will keep this open, as maybe it's a windows specific thing.

mysticaltech:
@Julian-Louis Fedora Linux.

PR #10, opened 2022-01-13
Add version constraint for Terraform provider (hcloud)

dhoppe:
The next major release might contain some changes that are not backwards compatible.


mysticaltech:
Thanks, @dhoppe!

PR #11, opened 2022-01-13
Reorganize structure of Terraform manifests

dhoppe:
This should lead to better readability.


mysticaltech:
Awesome. Makes sense @dhoppe! 🙏

PR #13, opened 2022-01-13
Update documentation (cp vs mv)

dhoppe:
You need to copy this file, otherwise git will complain about a deleted file which might end up in the next pull request.


mysticaltech:
Indeed!

PR #15, opened 2022-01-13
Make version of CCM / CSI configurable

dhoppe:
Users should be able to use a specific version of CCM / CSI. If no value has been provided, the latest available version will be fetched from the GitHub repository.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 .terraform*
 *.tfstate*
 crash.log
+hetzner/ccm/kustomization.yaml
+hetzner/csi/kustomization.yaml
 kubeconfig.yaml
+kubeconfig.yaml-e
 terraform.tfvars
 templates/rendered/traefik_config.yaml
--- a/data.tf
+++ b/data.tf
@@ -1,0 +1,15 @@
+data "github_release" "hetzner_ccm" {
+  repository  = "hcloud-cloud-controller-manager"
+  owner       = "hetznercloud"
+  retrieve_by = "latest"
+}
+
+data "github_release" "hetzner_csi" {
+  repository  = "csi-driver"
+  owner       = "hetznercloud"
+  retrieve_by = "latest"
+}
+
+data "hcloud_image" "linux" {
+  name = local.hcloud_image_name
+}
--- a/hetzner/ccm/kustomization.yaml
+++ b/hetzner/ccm/kustomization.yaml
@@ -1,8 +1,0 @@
-apiVersion: kustomize.config.k8s.io/v1beta1
-kind: Kustomization
-
-resources:
-- https://github.com/hetznercloud/hcloud-cloud-controller-manager/releases/latest/download/ccm-networks.yaml
-
-patchesStrategicMerge:
-- patch.yaml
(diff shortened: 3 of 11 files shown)


dhoppe:
Hm. I need to take care of the Docker images as well. 
https://github.com/kube-hetzner/kube-hetzner/blob/master/hetzner/ccm/patch.yaml#L10
https://github.com/kube-hetzner/kube-hetzner/blob/master/hetzner/csi/patch.yaml#L23

mysticaltech:
Thank you for that massive pull request. It looks good and I trust your good judgment.

dhoppe:
This pull request was not ready to be merged. We are able to download a specific version of *.yaml for CCM / CIS, but the patch.yaml will set the Docker image to latest. 

I was wondering if this could be handled via https://github.com/kbst/terraform-provider-kustomization

mysticaltech:
The reason I added the kustomization and changed the docker image to either latest or canary, with imagePullPolicy of Always, was to have them update automatically when the k3s system controller does upgrades both k3s and k3os. 

I realize it is risky, and indeed it would be nice if we could have that be optional.

mysticaltech:
@dhoppe Ideally, users should not touch Hetzner's own CCM and CSI. But I just wanted to make everything "auto-upgradable". 

So if we could add a switch to either have the default manifests with the version chosen, or the manifest customized with docker latest and canary like now, it would be great! Just to give users the choice.

mysticaltech:
Between, I have tested this PR, and it deploys the cluster just fine, so we are halfway there. I believe a switch to customize or not to customize would be great!

dhoppe:
I agree. At some point it gets pretty opinionated anyway. 

For example, I prefer ArgoCD to install the following tools:

- CCM
- CIS
- K8s Dashboard
- Prometheus Stack
- ...

I have created a few Terraform modules for `hcloud` and need to update / publish them sometime. Maybe with this we can get rid of the redundant code regarding server, master and agent.

mysticaltech:
It makes sense @dhoppe, only yesterday I was looking at ArgoCD, I realized it could do this, and probably do upgrades too. If that is the case, then the only customization needed is for the CCM to give it the right cluster CIDR like at https://github.com/kube-hetzner/kube-hetzner/blob/38219041ff6b55cdc00da28e8835c31e39f40f87/hetzner/ccm/patch.yaml#L19.

About the other tools you mention, if we could have them too, it would be awesome! Prometheus / Grafana and the k8s dashboard would really make this project complete!

mysticaltech:
@dhoppe Just removed the container images customizations, now it will use the vanilla manifest from Hetzner, just with the right version thanks to your work yesterday. Now it's cleaner indeed.

For my dream of the "automatic upgrade", probably it can be accomplished with ArgoCD at a later stage, to monitor new releases of the CCM and CSI (you tell me, I'm a beginner).

mysticaltech:
@dhoppe Finally also added the possibility to use the "latest" containers for the CCM and CSI via two new optional booleans `hetzner_ccm_container_latest` and `hetzner_csi_container_latest`, by default set to false.

PR #16, opened 2022-01-13
Add public IPv4 address of Traefik to output

dhoppe:
This output might be required by other Terraform modules to manage the DNS records.


mysticaltech:
Tried to merge this but it failed and had to revert it. Also, the type of LB is configured [here](https://github.com/kube-hetzner/kube-hetzner/blob/5a6b7779b91a4ee97eb5303bd2c5290e8841b3c7/terraform.tfvars.example#L12). So not always going to be lb11.

I understand what you are trying to achieve, maybe we can find a better way.

dhoppe:
This was just the name of the data source. The public IPv4 address will be fetched, based on the name `traefik`.

mysticaltech:
Ok, I understand, will just change `lb11` to `lb` and push it back then!

mysticaltech:
It did not work because the lb is automatically provisioned by the CCM, not through terraform. So there is no way I know of acquiring its IP unless we ask Kubernetes somehow.

When deploying I get the following error:

```
Error: no Load Balancer found with name traefik
 
   with data.hcloud_load_balancer.lb,
   on data.tf line 17, in data "hcloud_load_balancer" "lb":
   17: data "hcloud_load_balancer" "lb" {
```

mysticaltech:
@dhoppe If your concern is to have the lb IP to later configure DNS, this is easily solved by external-dns that can be installed in kube.

mysticaltech:
You seem pretty advanced :) But just in case, gave an example of using **external-dns** at the end of that issue https://github.com/kube-hetzner/kube-hetzner/issues/18

dhoppe:
Hm. To be honest, the loadbalancer already existed when I tested this. Unfortunately my DNS provider is not supported by external-dns, yet. 
https://github.com/kubernetes-sigs/external-dns/pull/1980

mysticaltech:
In that case, maybe there is a way to ask Kubernetes and grab the value, store it, and display that. Maybe with https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs

PR #19, opened 2022-01-13
Prevent secrets from being exposed

dhoppe:
The Hetzner API token should not be exposed in case of any errors, as you can see at #8.


mysticaltech:
Ah wow, learning a lot today! :) Perfect.

PR #20, opened 2022-01-13
Revert "Add public IPv4 address of Traefik to output"

mysticaltech:
Reverts kube-hetzner/kube-hetzner#16


mysticaltech:
It had failed the terraform checks, and even when I tried manually to add, it fails to run.

PR #22, opened 2022-01-23
make private key optional to support hardware tokens for SSH

phaer:
Hi & thanks for this fantastic project!

I'd like to use a yubikey for authentication with my cluster, so I don't have a private ssh key lying on my disk, only on the token. The way that normally works is that I just use my local ssh-agent and pass the needed ssh identity as a *public* key to the agent. The following PR supports that workflow if private_key is set to `null`.

It currently duplicates the logic in 3 files to keep it simple, but I could also introduce `local.private_key` and `local.agent_identity` to deduplicate if you wish so.
What do you think, is this a viable approach?

### DIFF
--- a/master.tf
+++ b/master.tf
@@ -24,7 +24,8 @@ resource "hcloud_server" "first_control_plane" {
 
     connection {
       user        = "root"
-      private_key = file(var.private_key)
+      private_key = var.private_key == null ? null : file(var.private_key)
+      agent_identity = var.private_key == null ? file(var.public_key) : null
       host        = self.ipv4_address
     }
   }
@@ -35,7 +36,8 @@ resource "hcloud_server" "first_control_plane" {
 
     connection {
       user        = "root"
-      private_key = file(var.private_key)
+      private_key = var.private_key == null ? null : file(var.private_key)
+      agent_identity = var.private_key == null ? file(var.public_key) : null
       host        = self.ipv4_address
     }
   }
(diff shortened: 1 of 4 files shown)


mysticaltech:
Thank you for this PR, it is really appreciated. I understand the use case, it's just that it adds some complexity to the code. Will leave the PR open for now and see if there's more support for that feature.

master.tf
```diff
@@ -35,15 +36,16 @@ resource "hcloud_server" "first_control_plane" {
...
+      agent_identity = var.private_key == null ? file(var.public_key) : null
       host        = self.ipv4_address
     }
   }
 
   # Wait for k3os to be ready and fetch kubeconfig.yaml
   provisioner "local-exec" {
     command = <<-EOT
-      sleep 60 && ping ${self.ipv4_address} | grep --line-buffered "bytes from" | head -1 && sleep 100 && scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i ${var.private_key} rancher@${self.ipv4_address}:/etc/rancher/k3s/k3s.yaml ${path.module}/kubeconfig.yaml
+      sleep 60 && ping ${self.ipv4_address} | grep --line-buffered "bytes from" | head -1 && sleep 100 && scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i ${var.private_key == null ? var.public_key : var.private_key} rancher@${self.ipv4_address}:/etc/rancher/k3s/k3s.yaml ${path.module}/kubeconfig.yaml
```
mysticaltech, line 48: I know it works, it's just a little confusing to me. And I do not want to add something I do not yet fully understand, otherwise, I can't maintain it later on.

phaer: What exactly is confusing? The mix of public and private keys? SSHs man page for -i: `Selects a file from which the identity (private key) for public key authentication is read.  You can also specify a public key file  to  use  the  corre‐
              sponding  private  key  that  is loaded in ssh-agent(1) when the private key file is not present locally.  ` 
And that's pretty much all there is to it; we don't have the private key locally present with yubikeys, other hardware tokens, or even SSH certificates and so on.

mysticaltech: Ok, thanks for the clarification! Will review and approve ASAP later today.

phaer: Great, thanks! I might try to simplify the code a bit by using locals later on. Thanks for considering this PR, I understand that all code is a liability, even more so code that one might not entirely understand at first. In my circles, yubikeys are quite common nowadays, but not too sure how popular they are in general. But this PR should enable other use cases as ssh certificates or just pre-loaded, password-protected private keys as well.

mysticaltech: Perfect! Yes, do not hesitate to simplify with locals, the simpler the better! :) And really appreciate your contribution, thank you. This will surely help others!

phaer:
Ah, I had not used terraform too much recently and wasn't aware of terraform fmt. Will fix later today

mysticaltech:
Ok, great! I will revert it for now (should've waited for the check to complete properly).

mysticaltech:
@phaer When you have the fixes (don't worry nothing urgent), don't hesitate to create a new PR! 🙏

mysticaltech:
Especially if it allows the use of passphrases like you've explained here https://github.com/kube-hetzner/kube-hetzner/pull/25#issuecomment-1021082091, it makes this PR even more important!

phaer:
@mysticaltech I think it should, but it might be a good idea to test it with a few setups more. I hope it still works with your setup in #27

PR #23, opened 2022-01-23
remove empty provider blocks

phaer:
terraform v1.1.4 (and at least v1.0.11) emits a warning for empty
provider blocks. Removing those still seems to work while avoiding
that warning.


mysticaltech (approved):

mysticaltech:
Thanks for letting me know @phaer, merged!

PR #25, opened 2022-01-24
adding passphrase notice to getting started

fraank:
as keys with passphrases are causing creation failure, this hint would have saved me 20min :-)


mysticaltech:
Thanks, @fraank! I guess it could work but would require adding `agent = true`, please have a look at https://stackoverflow.com/a/66867869

mysticaltech:
Am also pondering about closing the SSH port on the nodes as soon as the remote execution is done. That way, passphrase or not it becomes irrelevant. I also remember somewhere that ed25519 was secure enough to not need a passphrase, but can't find that anywhere now.

mysticaltech:
@phaer You seem to SSH well, what do you think about ed25519 and passphrases, are they important in your opinion? Shall we build logic to support them?

fraank:
maybe, a single `ssh-add` before executing wasn't enough.

mysticaltech:
@fraank  `ssh-add`, plus you have to edit the terraform files to and add `agent=true`, to the remote-exec connection params. Normally that should work! And maybe that [line](https://github.com/kube-hetzner/kube-hetzner/blob/5ffb366a93bd610ba8786e3e00f00518b6f097f6/master.tf#L46) too would need adaptation for the `scp` command, not sure.

phaer:
`agent` should be true by default, as I understand the [docs](https://www.terraform.io/language/resources/provisioners/connection) 

> Set to false to disable using ssh-agent to authenticate. On Windows the only supported SSH authentication agent is Pageant.

@fraank It should work via your agent, if you check out my [PR](https://github.com/kube-hetzner/kube-hetzner/pull/22) and set `private_key` to `null` so that terraform just asks your agent for the right identity using it's public key.
Please make sure that the right key is loaded via `ssh-add -L`, to list loaded ssh keys before.

PR #27, opened 2022-01-25
make private key optional to support SSH agent usage

phaer:
So this is just #22, but now using locals, correctly formatted and with a short mention in the README.

It works for me with my yubikey, i'd be especially interested if it solves the problem @fraank is having in #25.


mysticaltech:
Thank you for your hard work, I believe that is a very important PR.

PR #29, opened 2022-01-25
ingress: don't use private ip, make ipv6 optional

phaer:
The rationale for this PR has been documented in https://github.com/kube-hetzner/kube-hetzner/issues/18#issuecomment-1019993588. I believe that usage of kube-hetzner together with external-dns is broken without this settings - using cloudflare proxying just hides the issues, because cloudflare seems to filter the private IP automatically.
With this PR it should work both, with or without proxying as well as for other DNS providers supported by external-dns


mysticaltech:
Perfect, great improvements! I hesitated a little bit about turning off ipv6 by default, but we can always do that later. Thanks!

PR #33, opened 2022-01-29
Use variable for network region

TimHeckel:
Instead of hard-coding the network into `eu-central`, this allows the user to set the network region in the main config


mysticaltech (approved):

mysticaltech:
Thanks, @TimHeckel, appreciate the PR, I did not know this was needed for Ashburn, really appreciate it!

PR #34, opened 2022-01-29
Add Hetzner placement group and link servers to it

waammar:
it is important to use placement group for HA deployment, from Hetzner Cloud doc they are used to influence the distribution of interdependent virtual servers in their data centers.

In spread placement groups, all virtual servers are running on different physical servers. This decreases the probability that some instances might fail together.

### DIFF
--- a/agents.tf
+++ b/agents.tf
@@ -2,12 +2,13 @@ resource "hcloud_server" "agents" {
   count = var.agents_num
   name  = "k3s-agent-${count.index}"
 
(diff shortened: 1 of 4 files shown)


mysticaltech:
Thanks, @waammar, really appreciate it, indeed very useful, and solves one of the open issues https://github.com/kube-hetzner/kube-hetzner/issues/31.

waammar:
Oh I didn't see that there is already an issue! glad that I helped.

PR #37, opened 2022-02-01
added letsencrypt with http challenge configuration for traefik

Olivierwenger:
I added two optional variables `traefik_acme_tls` and `traefik_acme_email` to enable TLS with letsencrypt, this will directly create a prod acme. I chose the http Challenge as it isn't dependent on the DNS provider.

I noticed afterwards that some people already changed their config the same way after I finished the modification https://github.com/kube-hetzner/kube-hetzner/issues/18 so I don't know if my PR is really relevant, please remove it if it isn't how you want to do it.


templates/traefik_config.yaml.tpl
```diff
@@ -24,3 +24,9 @@ spec:
       - "--entryPoints.websecure.proxyProtocol.trustedIPs=127.0.0.1/32,10.0.0.0/8"
       - "--entryPoints.web.forwardedHeaders.trustedIPs=127.0.0.1/32,10.0.0.0/8"
       - "--entryPoints.websecure.forwardedHeaders.trustedIPs=127.0.0.1/32,10.0.0.0/8"
+%{ if traefik_acme_tls ~}
+      - "--certificatesresolvers.le.acme.httpchallenge=true"
+      - "--certificatesresolvers.le.acme.httpchallenge.entrypoint=web"
```
mnencia, line 29:
```suggestion
      - "--certificatesresolvers.le.acme.tlschallenge=true"
```

mnencia:
I think you should also allow overriding the `--certificatesresolvers.letsencrypt.acme.caServer` parameter because it's better to use the staging server when doing experiments. Otherwise, you could get temporarily suspended due to production rate limits.

templates/traefik_config.yaml.tpl
Olivierwenger: @mnencia True, I wasn't sure if Traefik supported TLS-ALPN-01 but it does, I'll try it this evening and change it accordingly if it works.

templates/traefik_config.yaml.tpl
```diff
@@ -24,3 +24,9 @@ spec:
...
+      - "--certificatesresolvers.le.acme.email=${traefik_acme_email}"
+      - "--certificatesresolvers.le.acme.storage=/data/acme.json"
```
mysticaltech, line 31: Maybe just use ` acme.json` instead of `/data/acme.json`, both will work definitely but the former is how they do it on one [Traefik example](https://doc.traefik.io/traefik/v2.0/user-guides/crd-acme/). 

Anyways, this file will not be saved permanently, so if the pod restarts it will reissue a certificate. But I believe that's not a problem.

I think a way to have the file saved permanently would be to let [cert-manager](https://cert-manager.io/)do the certificate fetching and storing as a Kubernetes TLS secret, then they can be used inside of an ingress definition like the certificates we add manually.

templates/traefik_config.yaml.tpl
mysticaltech: Good suggestion @mnencia, seems to be the best option indeed.

mysticaltech:
Love this proposal @Olivierwenger, thanks! It's definitely useful to have the option to configure it out of the box.

mysticaltech:
> I think you should also allow overriding the `--certificatesresolvers.letsencrypt.acme.caServer` parameter because it's better to use the staging server when doing experiments. Otherwise, you could get temporarily suspended due to production rate limits.

@mnencia Indeed, having the option to start with staging would be good. And later one can reapply the generated HelmChartConfig without the line in question:

`      - "--certificatesresolvers.letsencrypt.acme.caserver=https://acme-staging-v02.api.letsencrypt.org/directory"
`

mysticaltech:
@Olivierwenger I can add the staging option later no worries, as soon as you confirm that all works, we can merge this.

templates/traefik_config.yaml.tpl
Olivierwenger: @mysticaltech Actually I started with `acme.json` but this didn't work, I had the following erro on the traffic pod when I runned it:
`is skipped from the resolvers list because: unable to get ACME account: open acme.json: read-only file system`

I don't know really know why but storing it in the /data emptydir volume fixed it

Olivierwenger:
@mysticaltech I tested @mnencia proposed changes and it works. I just force pushed to change the `terraform.tfvars.example` accordingly.
I didn't add the the staging option.
Thanks for your help

mysticaltech:
@Olivierwenger The `terraform fmt -check -diff` command is failing. Probably you just need to run `terraform fmt` to align the formatting and style and push again, please.

Olivierwenger:
@mysticaltech done, it should work now

mysticaltech:
Awesome! Thanks again for this nice contribution @Olivierwenger! 🙏

TimHeckel:
Hey this is great - is there any way some documentation could be added to readme on how to use this though? Ideally, how do you create the Ingress after these options are enabled -- both for subdomains (e.g., blog.mysite.com) as well as the Ingress for the root (mysite.com)?

Olivierwenger:
@TimHeckel Yep I'll try to provide an example to test it. Would it be better to use Ingress or IngressRoute ?

TimHeckel:
Hi Oliver - for my money I think Ingress, just cause people may want to swap Traefik out for something, but if IngressRoute is what's working - then by all means just use that. Thanks!

PR #41, opened 2022-02-07
Reduce reboot time

mnencia:


locals.tf
```diff
@@ -17,17 +17,14 @@ locals {
     "aria2c --follow-metalink=mem https://download.opensuse.org/tumbleweed/appliances/openSUSE-MicroOS.x86_64-k3s-kvm-and-xen.qcow2.meta4",
     "qemu-img convert -p -f qcow2 -O host_device $(ls -a | grep -ie '^opensuse.*microos.*k3s.*qcow2$') /dev/sda",
     "sgdisk -e /dev/sda",
-    "partprobe /dev/sda",
     "parted -s /dev/sda resizepart 4 99%",
     "parted -s /dev/sda mkpart primary ext2 99% 100%",
+    "partprobe /dev/sda",
```
mnencia, line 22: Moving the partprobe here should be enough to make `udevadm settle` superfluous.

mysticaltech:
@mnencia Love it, it's just that for `head -1 && sleep 3` I actually had to change that last sleep, I had to change it from 30 to 60 because sometimes it was causing a waitForAction error, stopping the install. That's because sometimes the system takes time to finish booting and accept connections.

Now instead of only pinging, if we could make sure that ssh can connect, we could cut down that time.

locals.tf
mysticaltech: In my last commit, now pushed, I am using it, but please feel free to remove if you think `partprobe /dev/sda` is enough.

mnencia:
Now with a direct ssh check :-)

mysticaltech:
Wonderful! How did I not think of that before?! 😂 Thank you!

agents.tf
```diff
@@ -43,9 +43,14 @@ resource "hcloud_server" "agents" {
...
 
+  # Issue a reboot command
+  provisioner "local-exec" {
+    command = "ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -l root ${self.ipv4_address} '(sleep 2; reboot)&'; sleep 3"
+  }
+
   # Wait for MicroOS to reboot and be ready
   provisioner "local-exec" {
-    command = "sleep 60 && ping ${self.ipv4_address} | grep --line-buffered 'bytes from' | head -1 && sleep 60"
+    command = "until ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -l root -o ConnectTimeout=2 ${self.ipv4_address} true; do sleep 1; done"
```
mysticaltech, line 53: About this, it was causing lots of connection timeouts, like so:

ksnip_20220207-213050

Some people would have assumed that to be a problem, so I added back the ping, which waits silently for the network interface to come alive.

Then there was still this, the connection refused, because SSH is not yet ready:

ksnip_20220207-214249

To mitigate this, followed the ping with a sleep 10, and now we have something that looks good, that is faster, thanks to your ssh magic, and without waitForAction error, so I believe good compromise there! Everything looks ok, without any additional time, it's optimized to the maximum:

ksnip_20220207-220056

Just wanted to share those final tweaks to that line.

mnencia:
IMHO, there is a cleaner solution. Given that the timeout is expected, I would just redirect stderr and print a message, just to show we are still waiting. I'll prepare a pull request.

mysticaltech:
It's a pleasure to learn from you! Thanks...

locals.tf
mysticaltech: It redid it, so `partprobe /dev/sda`

ksnip_20220208-012740

So I just added the following, and it works, and gives me peace of mind to print the partitions :) 

ksnip_20220208-013108

ksnip_20220208-013446

redcapcloud (approved):
## Post-hoc Review

**Overall Risk:** 🟢 LOW
**Release Type:** PATCH (performance/reliability improvement; no API or state-affecting changes)

### Summary
Replaces fixed `sleep 60 && ping … && sleep 60` (≥120s) post-image-write wait with:
1. Out-of-band reboot via `ssh … '(sleep 2; reboot)&'; sleep 3`
2. Polling readiness loop: `until ssh … -o ConnectTimeout=2 … true; do sleep 1; done`
3. Removes in-band `shutdown -r +1; sleep 1; exit 0` from `locals.tf` (now redundant)
4. Reorders `partprobe`/`udevadm settle` around partition resize

### Security
- No hardcoded credentials, suspicious URLs, or obfuscation
- `StrictHostKeyChecking=no` is acceptable here — first-boot bootstrap, TOFU window unchanged from prior code
- `(sleep 2; reboot)&` is the idiomatic way to detach the SSH session before reboot kills it

### Backward Compatibility
No variable/output/naming/provider changes. Provisioner edits don't trigger recreation on existing servers.

### Recommendation
APPROVE. Clean, focused change. Polling-until-ready is strictly better than fixed sleeps; reboot orchestration is sound.

redcapcloud:
Apologies — my earlier review on this PR was posted in error. I was reviewing a different PR #41 in a different repository and selected the wrong target. The review verdict does not reflect any real evaluation of this 2022 contribution and should be disregarded. Sorry for the noise.

PR #42, opened 2022-02-07
Really disable servicelb

mnencia:
YAML syntax does not allow multiple fields with the same key; the last one overrides all the previous.


mysticaltech:
Oh my, indeed! Completely missed that... Thank you!! 🍾

PR #43, opened 2022-02-07
Taint control-plane nodes with node-role.kubernetes.io/master:NoSchedule

### LINKED ISSUE #39: Control-plane-0 should be tainted with 'node-role.kubernetes.io/master=true:NoSchedule'
Suggest to taint master node (control-plane-0) with `node-role.kubernetes.io/master=true:NoSchedule` as currently it is not tainted. 
I.e. adding 
```
  # Taint Control Plane as master node to avoid scheduling of workloads here
  provisioner "local-exec" {
    command = <<-EOT
      kubectl taint nodes "${self.name}" node-role.kubernetes.io/master=true:NoSchedule
    EOT
  }
```
to master.tf

Because of being not tainted, workload pods can get placed on `control-plane-0` and disrupt cluster 
(i.e. Cassandra or HDFS pod placed on `control-plane-0` with almost 100% guarantee disrupts cluster)

mnencia:
That is the expected status when you create a cluster with most of the deployment systems (i.e., kubeadm or kind)

Fix #39

### DIFF
--- a/templates/master_config.yaml.tpl
+++ b/templates/master_config.yaml.tpl
@@ -9,3 +9,5 @@ tls-san: ${node_ip}
 kubelet-arg: "cloud-provider=external"
 token: ${token}
 node-name: ${node_name}
+node-taint:
+- node-role.kubernetes.io/master:NoSchedule
--- a/templates/server_config.yaml.tpl
+++ b/templates/server_config.yaml.tpl
@@ -9,3 +9,5 @@ tls-san: ${node_ip}
 kubelet-arg: "cloud-provider=external"
 token: ${token}
 node-name: ${node_name}
+node-taint:
+- node-role.kubernetes.io/master:NoSchedule


phaer:
As @mysticaltech wrote in #39, it would be useful to have that configurable. That could of course be done in the template, but I am currently working on a PR to replace the templates with terraforms `yamlencode`, which would make stuff like this a bit easier to override IMO. Just needs testing and I just lost some time with #44. After lunch, probably

mysticaltech:
Thanks, @mnencia! Would just love to make it optional, because am using all nodes myself, even if apparently not the proper way to do it. Just having a variable for this, and tainted by default would be great, like `allow_scheduling_on_control_plane` default false, please! 🙏

mysticaltech:
I'll add the variable myself later on! Thanks for this.

PR #44, opened 2022-02-07
fix ssh identity in staging

phaer:
Newly added ssh commands were missing the flag -i to pass an
identity file. This means that those commands use different
settings then the provisioners and their connection blocks
around them.

While adding this parameter, I decided it would be cleanest
to add local.ssh_args.

### DIFF
--- a/agents.tf
+++ b/agents.tf
@@ -45,12 +45,12 @@ resource "hcloud_server" "agents" {
 
   # Issue a reboot command
   provisioner "local-exec" {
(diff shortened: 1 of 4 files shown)


mnencia (approved):

phaer:
Ah, just a sec - scp is broken atm, because `-l` means something different there ;)

mnencia:
You are right. I would just take `-l` out of the ssh_args string

mysticaltech:
Awesome!!

PR #45, opened 2022-02-07
fix kubectl get nodes missing --kubeconfig

phaer:
the current implementation works co-incidentally for most
setups, when terraform apply is run from the repos root,
but not when kube-hetzner is used as a terraform module


mnencia (approved):

mysticaltech:
Good catch @phaer!

PR #50, opened 2022-02-08
Replace k3s config templates with yamlencode

phaer:
Hi,

So this is a bit of a style question, but I'd personally prefer this approach for 3 small reasons:

* it avoids constructing YAML values via string templating (as with the node-taint)
* config values are closer to their use in source
* its a bit less code.

What do you think?


master.tf
```diff
@@ -59,11 +59,17 @@ resource "hcloud_server" "first_control_plane" {
...
-    content = templatefile("${path.module}/templates/master_config.yaml.tpl", {
-      node_ip                           = local.first_control_plane_network_ip
-      token                             = random_password.k3s_token.result
-      node_name                         = self.name
-      allow_scheduling_on_control_plane = var.allow_scheduling_on_control_plane
+    content = yamlencode({
+      node-name                = self.name
+      cluster-init             = true
+      disable-cloud-controller = true
+      disable                  = "servicelb, local-storage"
```
mnencia, line 66: I know this works, but I would prefer to use a list here.
```suggestion
      disable                  = ["servicelb", "local-storage"]
```

mysticaltech: Seems more in line with a YAML list indeed.

mysticaltech:
@phaer At first, I liked the YAML files, but seeing your code now, it's cleaner and more compact, to have the core logic centralized in their respective files (either master, server, or agent). I like it, please resolved @mnencia's suggestion, and if you've tried it, will merge. Thanks! 🙏

master.tf
mnencia: In the YAML configuration, it is currently:

``` yaml
disable:
- servicelb
- local-storage
```

However, both forms work.

phaer: Thanks, I'd missed that!

PR #51, opened 2022-02-09
Fix the kured reboot command

mnencia:


mysticaltech:
Oh my! :) Who would have guessed, well done!!

PR #52, opened 2022-02-09
Agents IP should not depend on servers_num

mnencia:
Otherwise, when you scale the number of the servers, all the agents will change IP, and the load balancer will become confused.

### DIFF
--- a/agents.tf
+++ b/agents.tf
@@ -76,13 +76,13 @@ resource "hcloud_server" "agents" {
     }
   }
 
-  # Generating k3s  agent config file
+  # Generating k3s agent config file
   provisioner "file" {
     content = yamlencode({
       node-name     = self.name
       kubelet-arg   = "cloud-provider=external"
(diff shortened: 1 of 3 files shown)


agents.tf
```diff
@@ -76,13 +76,13 @@ resource "hcloud_server" "agents" {
...
 
-  # Generating k3s  agent config file
+  # Generating k3s agent config file
   provisioner "file" {
     content = yamlencode({
       node-name     = self.name
       kubelet-arg   = "cloud-provider=external"
       flannel-iface = "eth1"
-      node-ip       = cidrhost(hcloud_network.k3s.ip_range, 2 + var.servers_num + count.index)
+      node-ip       = cidrhost(hcloud_network_subnet.k3s.ip_range, 257 + count.index)
```
mnencia, line 85: This will put all the agents in 10.0.1.x range

mysticaltech: Awesome!!

PR #62, opened 2022-02-10
fix: openSUSE url

shlyamster:
Fixed use of openSUSE for the master branch, not the staging


mysticaltech:
Thank you, realized it too at the same time. Appreciate the PR! 🙏

PR #46, opened 2022-02-07
WIP: Expose kubeconfig (staging)

phaer:
Same as https://github.com/kube-hetzner/kube-hetzner/pull/28, just for the staging branch

PR #66, opened 2022-02-11
Expose kubeconfig in outputs...

phaer:
This allows to use kubeconfig when including kube-hetzner into other terraform projects without having the path depend on the
modules source. It's local in this example but could just as well be a remote github repository.

Example terraform code:

```
module "kube-hetzner" {
  source = "./kube-hetzner"

  hcloud_token = "hahaha-not-really"
  public_key   = "/home/me/.ssh/my-first-key.pub"
  private_key  = null
  # For Hetzner locations see https://docs.hetzner.com/general/others/data-centers-and-connection/
  # For Hetzner server types see https://www.hetzner.com/cloud
  location                  = "fsn1"
  network_region            = "eu-central" # change to `us-east` if location is ash
  agent_server_type         = "cpx21"
  control_plane_server_type = "cpx11"
  lb_server_type            = "lb11"
  lb_disable_ipv6           = true
  servers_num               = 3
  agents_num                = 3
  allow_scheduling_on_control_plane = false

  # If you want to use a specific Hetzner CCM and CSI version, set them below, otherwise leave as is for the latest versions
  # hetzner_ccm_version = ""
  # hetzner_csi_version = ""

  # If you want to kustomize the Hetzner CCM and CSI containers with the "latest" tags and imagePullPolicy Always,
  # to have them automatically update when the node themselve get updated via the rancher system upgrade controller, the default is "false".
  # If you choose to keep the default of "false", you can always use ArgoCD to monitor the CSI and CCM manifest for new releases,
  # that is probably the more "vanilla" option to keep these components always updated.
  hetzner_ccm_containers_latest = true
  hetzner_csi_containers_latest = true
}

resource "local_file" "kubeconfig" {
  content         = module.kube-hetzner.kubeconfig_file
  filename        = "${path.module}/kubeconfig.yaml"
  file_permission = "0600"
}

output "controlplanes_public_ip" {
  value = module.kube-hetzner.controlplanes_public_ip
}

output "agents_public_ip" {
  value = module.kube-hetzner.agents_public_ip
}
```

* To do so, we need to ensure that the generated kubeconfig is part of
  terraforms dependency graph. This has the additional benefit of not
  depending on local files anymore which should enable multi-user
  setups.

* This also means that we can't deploy CCM, CSI & Traefik from our local
  host, because we don't have kubeconfig.yaml locally while provisioning
  the control plane, only afterwards.

* So we just run kubectl apply on the control plane itself, after k3s is
  ready.

* To do so, we need to deploy all manifests. I've merged the patches
  into a single kustomization.yaml file, because that makes the
  deployment of those files to the control-plane server easier.

* we could also put the traefik config into the same kustomization file,
  which would save us one of the file provisioner blocks. I didn't want
  this PR to get any bigger, and will consider merging this config later
  on. kustomization.yaml is small enough that we could yamlencode() for
  it and store the patches in separate files again, not as
  inline-strings which is kind of ugly.

This is the same as #28 and #46, just for the latest master branch.

What do you think @mysticaltech and @mnencia ?


mnencia:
I think that the idea is fine. I'll review and try it this evening.

I have some more ideas to evolve the project in further PRs:

1) Move the single host initial setup in a submodule: this would remove some code duplication
2) Separate the machine provision step from the Kubernetes node setup step: this allows to reduce the setup time further because you can provision all the nodes at the same time
3) Move the Kubernetes configuration in a submodule, so you can use the Kubernetes provider with an interpolated configuration to reconcile the CCM, CSI & Traefik services.

mysticaltech:
Looking good @phaer, will also try it tonight.

For the traefik config, please leave it separate because the file serves as a template to modify traefik's configuration manually later on, if ever needed.

mysticaltech:
Hey folks, out-of-context (we need a more direct comms channel) just so you know if my last commit I added changed these to default to true because of this issue from a user https://github.com/kube-hetzner/kube-hetzner/issues/59#issuecomment-1034753468. 

I should have done so through a PR, so we can discuss this better, will start doing more of those from now on.

ksnip_20220211-141259

mysticaltech:
First of all, just tried it! Works great, it's a great improvement, and makes everything clearer.

output.tf
```diff
@@ -7,3 +7,15 @@ output "agents_public_ip" {
   value       = hcloud_server.agents.*.ipv4_address
   description = "The public IP addresses of the agent server."
 }
+
+output "kubeconfig_file" {
```
mysticaltech, line 11: So if I understand correctly, this output is not meant to be read on the screen, but by another terraform program, correct? :) 

ksnip_20220211-221312

master.tf
```diff
@@ -98,59 +96,15 @@ resource "hcloud_server" "first_control_plane" {
           sleep 2
         done
       EOT
+      , <<-EOT
+         timeout 120 bash -c 'while [[ "$(curl -s -o /dev/null -w ''%%{http_code}'' curl -k https://localhost:6443/readyz)" != "200" ]]; do sleep 1; done'
+      EOT
+      , "kubectl -n kube-system create secret generic hcloud --from-literal=token=${var.hcloud_token} --from-literal=network=${hcloud_network.k3s.name}",
+      "kubectl -n kube-system create secret generic hcloud-csi --from-literal=token=${var.hcloud_token}",
+      "kubectl apply -k /tmp/",
```
mysticaltech, line 104: @phaer Just to make it clearer, can we apply `/tmp/post_install_kustomization` instead? I believe the file command will create missing folders, no?

master.tf
```diff
@@ -98,59 +96,15 @@ resource "hcloud_server" "first_control_plane" {
...
+      "kubectl apply -f /tmp/traefik.yaml",
+      "rm /tmp/traefik.yaml /tmp/kustomization.yaml"
```
mysticaltech, line 106: Then here, we can do a rm -rf

templates/kustomization.yaml.tpl
```diff
@@ -0,0 +1,128 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+- "https://github.com/hetznercloud/hcloud-cloud-controller-manager/releases/download/${ccm_version}/ccm-networks.yaml"
+- "https://raw.githubusercontent.com/hetznercloud/csi-driver/${csi_version}/deploy/kubernetes/hcloud-csi.yml"
+- "https://github.com/weaveworks/kured/releases/download/${kured_version}/kured-${kured_version}-dockerhub.yaml"
+
+patchesStrategicMerge:
```
mysticaltech, line 10: @phaer Shouldn't we make the order of patches follow the order that the resources are listed in? kured, ccm, csi, I would prefer, maybe not that useful, but feels safer.

master.tf
```diff
@@ -98,59 +96,15 @@ resource "hcloud_server" "first_control_plane" {
           sleep 2
         done
       EOT
+      , <<-EOT
+         timeout 120 bash -c 'while [[ "$(curl -s -o /dev/null -w ''%%{http_code}'' curl -k https://localhost:6443/readyz)" != "200" ]]; do sleep 1; done'
```
mysticaltech, line 100: This is great, and happy to discover that method! Just if we could provide some feedback, like what we do for the other sections it would be great! Something like "Waiting for the cluster to be ready...", because now there are a few minutes of silence, which could make some users panic and abort.

ksnip_20220211-220629
:

output.tf
phaer: Yes, see `resource "local_file" "kubeconfig"` in my example code above.

 `terraform output kubeconfig_file` should also work? (Just deleted my testing cluster before the weekend, so I can't verify atm)

master.tf
```diff
@@ -98,59 +96,15 @@ resource "hcloud_server" "first_control_plane" {
...
+      EOT
```
mysticaltech, line 101: Starting from here, there are some sensitive value, that hides the whole thing, so maybe we can dissociate this into another block? Just an idea, don't worry about it, I can work on it later.

And for the formatting, if we can have only one <<-EOT  ... EOT at the end, like other section, it would look nicer. 

Otherwise, great functionality.

master.tf
phaer: Ah, I thought it didn't, but haven't tested it tbh. Would absolutely be cleaner, going to check! (if nobody does before me ;))

master.tf
mysticaltech: @phaer Basically, don't worry too much about the formatting of the code, this is not a problem, but if we can just minimize

 this kind of opaque outputs while doing remote-exec, it would be great:

ksnip_20220211-222946

templates/kustomization.yaml.tpl
phaer: Order should be irrelevant as it matches on kind/namespace/name, afaik.

But I agree that its cleaner, but I don't really like the embedded patches anyway. Might replace `kustomization.yaml` with yamldecode + file for the patches. So we could keep them separate in the repo while still deploying only a single file.

output.tf
mysticaltech: Ah yes, it does! Beautiful...

master.tf
phaer: Not sure if this is the right place for this question - we might really need another communication channel - but, I don't really understand the loop before the code quoted here:

```sh
        until systemctl status k3s-server > /dev/null
        do
          systemctl start k3s-server
          echo "Initiating the cluster..."
          sleep 2
        done
```
Because `systemctl start k3s-server` should wait for k3s to be ready, right? And if it fails, the whole provisioner would fail? So is there any case where there's more than one iteration of this loop?

phaer: ...and you are totally right about the formatting. Will fix

templates/kustomization.yaml.tpl
mysticaltech: That would be awesome, and more maintainable.

master.tf
phaer: This sadly does not work. 
with

    destination = "/tmp/post_install/kustomization.yaml"
  
and 

    destination = "/tmp/post_install/traefik.yaml"
  
It runs fine, but you end up with "/tmp/post_install" being a file with the contents of traefik.yaml :/ But i am going to restructure the code so that we create that directory before

master.tf
mysticaltech: Yes, exactly, `systemctl start` does not return success or failure, only `systemctl status` does, so that basically just waits for k3s to be started before it moves on. If there is a real problem, the output will keep on saying "Initiating the cluster...", which is not bad.

mysticaltech: For more direct comms, shall we try elements (i.e the matrix protocol)? It seems super-evolved and it's like slack or discord, but probably better. I have created a private room here @phaer @mnencia https://matrix.to/#/!RowCRsakuwIKnDNrMe:matrix.org?via=matrix.org

Will delete this comment in 24h

phaer: Matrix would be great as I am already using it :) 
But I can't join using your link, could you invite @phaer:matrix.org directly?
image

mnencia: Same here  it says I'm not invited. I'm @mnencia:matrix.org

templates/kustomization.yaml.tpl
phaer: See new commits.

master.tf
```diff
@@ -88,69 +74,89 @@ resource "hcloud_server" "first_control_plane" {
       "hostnamectl set-hostname ${self.name}",
       # first we disable automatic reboot (after transactional updates), and configure the reboot method as kured
       "rebootmgrctl set-strategy off && echo 'REBOOT_METHOD=kured' > /etc/transactional-update.conf",
+      # prepare a directory for our post-installation kustomizations
+      "mkdir -p /tmp/post_install",
```
mysticaltech, line 78: Exactly! One line for the price of clarity :)

master.tf
```diff
@@ -88,69 +74,89 @@ resource "hcloud_server" "first_control_plane" {
...
+    inline = [
+      # This ugly hack is here, because terraform serializes the
+      # embedded yaml files with "- |2", when there is more than
+      # one yamldocument in the embedded file. Kustomize does not understand
+      # that syntax and tries to parse the blocks content as a file, resulting
+      # in weird errors. so gnu sed with funny escaping is used to
+      # replace lines like "- |3" by "- |" (yaml block syntax).
+      # due to indendation this should not changes the embedded
+      # manifests themselves
+      "sed -i 's/^- |[0-9]\\+$/- |/g' /tmp/post_install/kustomization.yaml",
```
mysticaltech, line 157: Perfect!

master.tf
```diff
@@ -88,69 +74,89 @@ resource "hcloud_server" "first_control_plane" {
...
       # first we disable automatic reboot (after transactional updates), and configure the reboot method as kured
       "rebootmgrctl set-strategy off && echo 'REBOOT_METHOD=kured' > /etc/transactional-update.conf",
+      # prepare a directory for our post-installation kustomizations
+      "mkdir -p /tmp/post_install",
       # then we initiate the cluster
       "systemctl enable k3s-server",
+      # start k3s
+      "systemctl start k3s-server",
+      # wait for k3s to get ready
       <<-EOT
```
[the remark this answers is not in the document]
phaer: Ah, this is actually a duplicate. I've re-added the original loop later on, after running into errors.

mysticaltech:
Thanks @phaer, amazing work! 🙏

PR #69, opened 2022-02-14
Expose load balancer ip

phaer:
Hi,

So this is not 100% percent solid, as data.hcloud_load_balancer depends on the load balancer being deployed quick enough after CCM deployment is started - but it worked for me in practice on three runs and I think it's really useful to have this external endpoint in terraforms output.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -156,5 +156,5 @@ resource "hcloud_placement_group" "k3s" {
 data "hcloud_load_balancer" "traefik" {
   name = "traefik"
 
-  depends_on = [ hcloud_server.first_control_plane ]
+  depends_on = [hcloud_server.first_control_plane]
 }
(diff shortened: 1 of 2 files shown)


mysticaltech:
Sounds great! Now, if we are insured to have at least one agent, we could make it depend on agent 0, and that would be enough time, no? What do you think? And if there is no agent 0, then the output does not show. Can this be done?

phaer:
Good suggestion, but that would only add more time, while having no semantic relation to agent-0, right? So I'd personally prefer to leave it depending on first-control-plane because that's semantically correct (we need to deploy CCM for the load-balancer to be created) and it's been working fine in my experience.
While testing it, terraform timed out twice on deleting the k3s network and one of the agents failed once to enter rescue mode. So the load-balancer-ip thing seems to be relatively stable and we can still work on a better solution if necessary ;)

mysticaltech:
You are right! Honestly, I did not even know we could fetch the IP in such a way without creating it ourselves, so well done! :)

PR #72, opened 2022-02-16
adds support for inline ssh keys

mathewmeconry:
Adds support for inline ssh public and private key.
With this change, we can generate the ssh key with https://registry.terraform.io/providers/hashicorp/tls/latest/docs/resources/private_key and pass it into the module.

This way the ssh keys don't have to be distributed between multiple engineers and the keys can be stored inside the state (Read the "Important Security Notice" in the `tls` documentation before storing your keys in the state)

### DIFF
--- a/agents.tf
+++ b/agents.tf
@@ -37,22 +37,27 @@ resource "hcloud_server" "agents" {
   }
 
   # Issue a reboot command
-  provisioner "local-exec" {
-    command = "ssh ${local.ssh_args} root@${self.ipv4_address} '(sleep 2; reboot)&'; sleep 3"
+  provisioner "remote-exec" {
+    inline = [
+      "sleep 2",
+      "reboot"
+    ]
+    # reboot doesn't return a proper exit code, so we have to trust that it works
+    on_failure = continue
   }
 
   # Wait for MicroOS to reboot and be ready
   provisioner "local-exec" {
     command = <<-EOT
-      until ssh ${local.ssh_args} -o ConnectTimeout=2 root@${self.ipv4_address} true 2> /dev/null
+      sleep 5
+      until ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o PubkeyAuthentication=no -o PasswordAuthentication=no -o KbdInteractiveAuthentication=no -o ChallengeResponseAuthentication=no -o ConnectTimeout=2 ${self.ipv4_address}  2>&1 | grep "Permission denied"
       do
         echo "Waiting for MicroOS to reboot and become available..."
         sleep 2
       done
     EOT
   }
-
-
+  
   # Generating and uploading the agent.conf file
   provisioner "file" {
     content = templatefile("${path.module}/templates/agent.conf.tpl", {
--- a/master.tf
+++ b/master.tf
@@ -35,14 +35,20 @@ resource "hcloud_server" "first_control_plane" {
   }
 
   # Issue a reboot command
-  provisioner "local-exec" {
-    command = "ssh ${local.ssh_args} root@${self.ipv4_address} '(sleep 2; reboot)&'; sleep 3"
+  provisioner "remote-exec" {
+    inline = [
+      "sleep 2",
+      "reboot"
+    ]
+    # reboot doesn't return a proper exit code, so we have to trust that it works
+    on_failure = continue
   }
 
   # Wait for MicroOS to reboot and be ready
   provisioner "local-exec" {
     command = <<-EOT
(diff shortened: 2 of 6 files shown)


agents.tf
```diff
@@ -37,22 +37,27 @@ resource "hcloud_server" "agents" {
   }
 
   # Issue a reboot command
-  provisioner "local-exec" {
-    command = "ssh ${local.ssh_args} root@${self.ipv4_address} '(sleep 2; reboot)&'; sleep 3"
+  provisioner "remote-exec" {
+    inline = [
```
mysticaltech, line 41: Actually, we do not need this, at line 40, we have already exited when it reboots, by using a background process. So that is not needed.

agents.tf
```diff
@@ -37,22 +37,27 @@ resource "hcloud_server" "agents" {
...
+    # reboot doesn't return a proper exit code, so we have to trust that it works
+    on_failure = continue
   }
 
   # Wait for MicroOS to reboot and be ready
   provisioner "local-exec" {
     command = <<-EOT
-      until ssh ${local.ssh_args} -o ConnectTimeout=2 root@${self.ipv4_address} true 2> /dev/null
+      sleep 5
+      until ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o PubkeyAuthentication=no -o PasswordAuthentication=no -o KbdInteractiveAuthentication=no -o ChallengeResponseAuthentication=no -o ConnectTimeout=2 ${self.ipv4_address}  2>&1 | grep "Permission denied"
```
mysticaltech, line 53: @phaer I believe that is your area of expertise :)

master.tf
```diff
@@ -35,14 +35,20 @@ resource "hcloud_server" "first_control_plane" {
   }
 
   # Issue a reboot command
-  provisioner "local-exec" {
-    command = "ssh ${local.ssh_args} root@${self.ipv4_address} '(sleep 2; reboot)&'; sleep 3"
+  provisioner "remote-exec" {
+    inline = [
```
mysticaltech, line 39: Same as before, not needed.

mysticaltech:
Thanks, @mathewmeconry! 

@phaer Please have a look at the SSH changes. I believe using the private key in PEM format is not a bad idea, but up to you.

mathewmeconry:
Hi @phaer @mysticaltech 

I agree with you but the use case I try to support now is to keep the code on our side as generic as possible and allow anybody in the team to launch a k8s on Hetzner without having the risk of committing the private key to our git repo. And having the state not in the git repo but rather on Google, AWS or any other remote state storage is from my point of view secure enough to store an ssh keypair. What do you think?

Good idea with the `ssh_args_no_auth` in locals. Gonna do that shortly. Unfortunately, the sleep command is necessary, otherwise it is possible that the loop can connect to the server before it started rebooting (had that issue a few times while testing)

Supporting multiple public keys and adding them to Hetzner Could would be from my point of view the best approach. 
What do you guys think when I rewrite the ssh part that it would support multiple ssh public keys? Then we can rely on the ssh-agent (which is already there) for most production environments with multiple operators

agents.tf
mnencia: Both the previous version and this one are tricks. It's a matter of taste which one you prefer. If you prefer this version you can remove the `sleep 2` command and just reboot. 

My preference would be for the old version, because it is able to detect eventual communication issues.

agents.tf
mnencia: I agree with @phaer to move the list of arguments in a local variable

terraform.tfvars.example
```diff
@@ -1,8 +1,13 @@
 # You need to replace these
 hcloud_token = "xxxxxxxxxxxxxxxxxxYYYYYYYYYYYYYYYYYYYzzzzzzzzzzzzzzzzzzzzz"
+
+# public_key or public_key_pem has to be set
 public_key   = "/home/username/.ssh/id_ed25519.pub"
-# Must be "private_key = null" when you want to use ssh-agent, for a Yubikey like device auth or an SSH key-pair with passphrase
+public_key_pem = "ssh-rsa ...."
```
mnencia, line 6: I do not think that "pem" suffix is right here. It is not in PEM format.

terraform.tfvars.example
```diff
@@ -1,8 +1,13 @@
...
+
+# Must be "private_key = null" and "private_key_pem = null" when you want to use ssh-agent, for a Yubikey like device auth or an SSH key-pair with passphrase
```
mnencia, line 8: I would split the long line

mathewmeconry:
I close this in favor of #75. It resolves the underlining problem I had in a better way.

PR #75, opened 2022-02-18
adds functionality to add more public ssh keys

mathewmeconry:
Replaces pull request #72

It doesn't support the same use case as #72 but accomplishes nearly the same feature in an easier way.
With this change we are able to add more than 1 ssh public key into the `authorized_hosts` file

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -2,7 +2,7 @@ locals {
   first_control_plane_network_ip = cidrhost(hcloud_network_subnet.k3s.ip_range, 2)
   hcloud_image_name              = "ubuntu-20.04"
 
-  ssh_public_key = trimspace(file(var.public_key))
+  ssh_public_key  = trimspace(file(var.public_key))
   ssh_public_keys = concat(var.additional_public_keys, [local.ssh_public_key])
   # ssh_private_key is either the contents of var.private_key or null to use a ssh agent.
   ssh_private_key = var.private_key == null ? null : trimspace(file(var.private_key))
--- a/locals.tf
+++ b/locals.tf
@@ -3,6 +3,7 @@ locals {
   hcloud_image_name              = "ubuntu-20.04"
 
   ssh_public_key = trimspace(file(var.public_key))
+  ssh_public_keys = concat(var.additional_public_keys, [local.ssh_public_key])
   # ssh_private_key is either the contents of var.private_key or null to use a ssh agent.
   ssh_private_key = var.private_key == null ? null : trimspace(file(var.private_key))
   # ssh_identity is not set if the private key is passed directly, but if ssh agent is used, the public key tells ssh agent which private key to use.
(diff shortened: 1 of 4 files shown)


phaer:
Thanks! I love the approach of just authorizing more public keys instead of storing private keys in terraform state, but I have two suggestions:

* I think introducing and using `local.all_server_ips` or so would be preferable.
* Why don't we just add additional public keys to `templates/config.ign.tpl`? If it's to much of a burden to properly handle a list there, replacing that template with `yamlencode()` might be a good idea - there are a few examples for that in this repo already.
That would save us a null resource, which would be nice IMO as they tend to be(come) kind of hacky quickly. A solution using a null resource might be worth it if it not only allowed additional ssh keys to be added to *and removed from* an existing cluster (without re-provisioning nodes). But as that can be easily achieved through an interactive shell(script), I think just adding additional keys to `config.ign` would be sufficient.

mathewmeconry:
hey
I had to split the `null_resource` up because the number of instances couldn't have been precomputed.
Personally, I would love to keep the `null_resource` in that we can exactly update the keys on the existing servers without redeploying them. And I would love to have it in the module that we don't have to fork the repo into our org and update the template.

phaer:
Hi,

Sorry, but even more null resources with `always_run` triggers does not seem like the best way forward for me. Those would make deployments unnecessarily more complicated and a bit slower even for people who just a single ssh key.
Additionally, it still does not support the removal ob existing keys, which it should if it already allows adding keys during runtime.

I personally believe that terraform is just the wrong tool if you'd like to add and remove ssh keys dynamically during the live-time of a node - and the existing null resource hacks seem to support that argument. Both a simple shell script or something like ansible could be easily used for such task.

Feel free to submit a PR which adds more ssh keys to config.ign during node provisioning! But I sadly don't think it's a good idea to merge this as-is - if you find a way to make it less hacky, remove key removal and so on, we might reconsider :)

mathewmeconry:
Hey @phaer 

thx for the reply. 
It only triggers if the ssh keys change and it does replace all other keys on the server. See lines 14, 38 and 64 where it replaces the existing contents of `/root/.ssh/authorized_keys`

the triggers on line 23, 47 and 73 only fire when there is a change in the `local.ssh_public_keys` public keys variable.

phaer:
> the triggers on line 23, 47 and 73 only fire when there is a change in the local.ssh_public_keys public keys variable.

...which always is the case during the first run, even if local.ssh_public_keys is just an empty list, right? 

But regardless, the other points of my comment still stand :)

mathewmeconry:
> > the triggers on line 23, 47 and 73 only fire when there is a change in the local.ssh_public_keys public keys variable.
> 
> ...which always is the case during the first run, even if local.ssh_public_keys is just an empty list, right?
> 
> But regardless, the other points of my comment still stand :)

right. for the first run we need to add the additional keys. 

Ok your decision :) ...

phaer:
...for users which use that feature, which will be a small minority for the foreseeable future. For all others it's just complicating and slowing down things. I've already sketched out a solution which would provision keys during installation without adding much overhead and maintenance cost for existing users *and* still allows you to change keys as you like through ssh.

Your approach does not even remove keys from servers after they are removed from "local.ssh_public_keys", which makes it pratically non-declarative. You are more than welcome to propose a solution which fixes those issues! But to get the current approach merged, you'd need to fork this repo and maintain that yourself, I am afraid ;)

PR #78, opened 2022-02-19
Alllow additional SSH public keys, use jsonencode for config.ign

phaer:
Hello!

This is my approach to support a subset of the use-cases in #72 and #75 with a simplified model. A drawback is that new keys can't be added via terraform after nodes have been provisioned, benefit is that the code is a bit simpler. And modifying the authorized_keys files for existing nodes is still possible via conventional means such as interactive shell access, ansible playbooks, etc.


mysticaltech:
Thanks, @phaer, awesome!

mathewmeconry:
thanks @phaer

PR #85, opened 2022-02-20
Added documentation to use TLS with traefik

Olivierwenger:
Here's  the tested documentation to quickly spin up an nginx pod with TLS, as asked here https://github.com/kube-hetzner/kube-hetzner/pull/37#issuecomment-1037554173

I'm just not sure about the subdomains part, I usually use one certificate per subdomain,

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,41 @@ _To turn off k3s upgrades, you can either set the `k3s_upgrade=true` label in th
 kubectl -n system-upgrade label node <node-name> k3s_upgrade-
 ```
 
+## Example Ingress with TLS
+
+Here is an example of an ingress to run an application with TLS, change the host to fit your need in `examples/tls/ingress.yaml` and then deploy the example
+```sh
+kubectl apply -f examples/tls/.
+```
+```yml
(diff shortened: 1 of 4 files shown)


mysticaltech:
I love the idea of an examples directory. Thanks for that @Olivierwenger!

@phaer and @mnencia, I propose we merge this without further do, what do you say?

phaer:
Looks good to me, I think there's more to be done regarding more structured examples in the future, but it's a good start
One suggestion: I would prefer `examples` instead of `.examples` in this case. Using hidden directories might make sense for the images, but examples should be discoverable so there's no benefit in hiding them on *nix systems in my opionion.

Thanks, @Olivierwenger!

TimHeckel:
@Olivierwenger  - thank you for documenting this!

mysticaltech:
Thanks folks! @Olivierwenger if you can change from . examples to examples it would be great, then we'll merge! 🙏

phaer:
@Olivierwenger Thanks!

Olivierwenger:
@phaer thanks to you for being so quick :)

PR #86, opened 2022-02-21
Expose load balancer ip

phaer:
This is the same as #69, but this time we explicitly wait until the `traefik` service exists and has an ingress ip set (which means that the load-balancer has been deployed by hetzner CCM).

The extra `null_resource` seems to be necessary because traefik does not start until the cluster is fully operational, which it isn't with just a single node. It might be sufficient to depend only on the control planes, but I did not test that explictly.

We also run `hcloud load-balancer delete traefik` locally on destroy, which seems to be sufficient to make the destruction of hcloud_network.k3s work as expected.

Please tell me if I've missed yet another edge case, @mnencia and @mysticaltech ;)

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -156,19 +156,5 @@ resource "hcloud_placement_group" "k3s" {
 data "hcloud_load_balancer" "traefik" {
   name = "traefik"
 
-  depends_on = [null_resource.cluster_provisioning]
-}
-
-
-resource "null_resource" "cluster_provisioning" {
-
-  provisioner "local-exec" {
-    when    = destroy
-    command = <<-EOT
-      hcloud load-balancer delete traefik
-      hcloud network delete k3s
-    EOT
-  }
-
   depends_on = [null_resource.first_control_plane]
 }
--- a/main.tf
+++ b/main.tf
@@ -156,5 +156,40 @@ resource "hcloud_placement_group" "k3s" {
 data "hcloud_load_balancer" "traefik" {
   name = "traefik"
 
-  depends_on = [module.first_control_plane]
+  depends_on = [null_resource.cluster_provisioning]
+}
+
+
+resource "null_resource" "cluster_provisioning" {
+
+  triggers = {
+    agent_ids         = "${join(",", module.agents.*.id)}"
+    control_plane_ids = "${join(",", concat([module.first_control_plane.id], module.control_planes.*.id))}"
+  }
+
+  depends_on = [ null_resource.first_control_plane, null_resource.control_planes, null_resource.agents ]
+
+  provisioner "remote-exec" {
+    connection {
+      user           = "root"
+      private_key    = local.ssh_private_key
+      agent_identity = local.ssh_identity
+      host           = module.first_control_plane.ipv4_address
+    }
+
+    inline = [
+      <<-EOT
+      timeout 120 bash <<EOF
+      until [ -n "\$(kubectl get -n kube-system service/traefik --output=jsonpath='{.status.loadBalancer.ingress[0].ip}')" ]; do
+          echo "Waiting for load-balancer to get an IP..."
+      done
+      EOF
+      EOT
+    ]
+  }
+
+  provisioner "local-exec" {
+    when = destroy
+    command = "hcloud load-balancer delete traefik"
+  }
 }
--- a/main.tf
+++ b/main.tf
@@ -162,34 +162,13 @@ data "hcloud_load_balancer" "traefik" {
 
 resource "null_resource" "cluster_provisioning" {
 
-  triggers = {
-    agent_ids         = "${join(",", module.agents.*.id)}"
-    control_plane_ids = "${join(",", concat([module.first_control_plane.id], module.control_planes.*.id))}"
-  }
-
-  depends_on = [null_resource.first_control_plane, null_resource.control_planes, null_resource.agents]
-
-  provisioner "remote-exec" {
-    connection {
-      user           = "root"
-      private_key    = local.ssh_private_key
-      agent_identity = local.ssh_identity
-      host           = module.first_control_plane.ipv4_address
-    }
-
-    inline = [
-      <<-EOT
-      timeout 120 bash <<EOF
-      until [ -n "\$(kubectl get -n kube-system service/traefik --output=jsonpath='{.status.loadBalancer.ingress[0].ip}')" ]; do
-          echo "Waiting for load-balancer to get an IP..."
-      done
-      EOF
-      EOT
-    ]
-  }
-
   provisioner "local-exec" {
     when    = destroy
-    command = "hcloud load-balancer delete traefik"
+    command = <<-EOT
+      hcloud load-balancer delete traefik
+      hcloud network delete k3s
+    EOT
   }
+
+  depends_on = [null_resource.first_control_plane]
 }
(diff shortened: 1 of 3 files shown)


mysticaltech:
It just works out of the box! :-) This null_resource business is magic haha. I say we can merge this now.

mysticaltech:
Well done @phaer, beautiful and elegant solution!

mysticaltech:
Hmm, it's the first time I see this, and do not think it is related to this PR, maybe it's something happening randomly since the new setup? Could be related to https://github.com/kube-hetzner/kube-hetzner/issues/89? No idea. 

All nodes create fine, but then, one becomes NotReady and can't be reached by SSH.

Some outputs here, and will destroy and retry.

ksnip_20220221-223512
ksnip_20220221-223527
ksnip_20220221-223603
ksnip_20220221-223623

mysticaltech:
On the second try, completely different behavior, k3s installs correctly but fails to start on first-control-plane, because of a missing service. I have honestly no idea what is happening folks, like nada! It does not seem related to these changes, but am only seeing this today. Will retry tomorrow. Please @phaer and @mnencia, try deploying on your side.

ksnip_20220221-224830
ksnip_20220221-224859

mysticaltech:
`nm` is for network manager no? That would explain why even SSH was not working on the NotReady node in the previous attempt, somehow the network goes down randomly?!

mysticaltech:
On master, everything seems smooth, though k3s-control-plane-1 became notReady for a sec and shows these events, however it became ready again very fast.

ksnip_20220221-230144

mysticaltech:
For other nodes, no reboot. And all fine for all nodes in the end.
ksnip_20220221-230218

mysticaltech:
What could be happening here, folks? It seems the null-resources may be tricky, no? Or is it Hetzner that is unstable and random, because of the different placement groups?

mysticaltech:
The `image capacity 0` warning seems normal because it does not have time to collect metrics at the very beginning of operations. See https://github.com/k3s-io/k3s/issues/1857#issuecomment-993774346

mnencia:
I did a couple of rounds of testing. One, it worked like a charm; the 2nd time, one of the hosts went up without the second interface. The web console shows the network as attached, but the host does not see it, even after a restart. I've had to destroy and recreate the host to make it work.

Given the variety of issues we experienced this evening, I'm inclined to think it's Hetzner having problems.

mysticaltech:
Re-deployed this branch, and also get an inconsistent state. It shows at circa minute 7.

ksnip_20220221-232853
ksnip_20220221-232915

mysticaltech:
That network describe seems weird, we have 2 agents but 3 IP with 10.0.1.x !

ksnip_20220221-233310

main.tf
```diff
@@ -153,10 +153,43 @@ resource "hcloud_placement_group" "k3s" {
...
+  depends_on = [null_resource.cluster_provisioning]
+}
+
+
+resource "null_resource" "cluster_provisioning" {
+
+  triggers = {
+    agent_ids         = "${join(",", module.agents.*.id)}"
+    control_plane_ids = "${join(",", concat([module.first_control_plane.id], module.control_planes.*.id))}"
+  }
```
mnencia, line 168: Do we need the triggers here? We do not need to retrigger this if a host changes.

mysticaltech:
Am out of ideas for today, please test more folks, both master and this, and let it sit for a few minutes.

mysticaltech:
Just saw your comment @mnencia, yeah me too, think the same, something is up with Hetzner tonight.

mysticaltech:
Nodes are back to Ready, confirming the theory above.

ksnip_20220221-234310

ksnip_20220221-234300

mysticaltech:
@phaer On second view, are you sure the null_ressource is really necessary here? Maybe just wait on first_control_plane, because it is the one that configures traefik, and anyway, even if Traefik will not get an IP before other nodes, we are waiting for it with kubectl, and the other nodes get deployed in parallel, so eventually it should succeed, without the null resource, no?

mysticaltech:
The null ressource that you use is actually very smart @phaer, it represents the cluster as a whole! :) But based on @mnencia comment above, it got me thinking. I removed the triggers and moved the waiting to the end of master. Because anyways, other nodes get deployed in parallel.

Worked like a charm!

ksnip_20220222-001225
ksnip_20220222-001515

main.tf
phaer: You are right, that was a relic from an earlier approach

phaer:
> But based on @mnencia comment above, it got me thinking. I removed the triggers and moved the waiting to the end of master. Because anyways, other nodes get deployed in parallel.

Awesome! Yes, you both are right about the triggers - I've removed them. Feel free to add your additional changes to the branch @mysticaltech and/or to merge as-is.

mysticaltech:
Awesome, pushing the proposed changes now.

phaer:
I wouldn't delete the network here, because thats its own terraform resource and because this changes run after terraform has finished planning, terraform might fail to delete it when it expects the network still to exist, no? 

Additionally, we need to allow it to run without failure, even if the load-balancer got deleted beforehand. Fixing.

mysticaltech:
So for the null resource, I chose to depend only on first_control_plane, as it is the only node responsible for the cluster provisioning, also where Traefik is deployed and waited for. 

Also did this, before some users for instance in issue #89, deploy with only 1 control plane and nothing else. Maybe it would have worked too, but just in case.

As for the destroy, it got stuck sometimes, and from experience, just added delete network, and now it's faster than ever before.

If you are good @phaer and @mnencia, we merge! Deployed a few times and everything seems stable.

mysticaltech:
@mnencia Thinking about your wish to make first_control_plan more independent... Maybe we could embed all the logic of master, in servers, but have it run only for the when index == 0, what do you say, is that possible?

phaer:
I think stylistically, the cluster_provisioning resource is a bit empty with just a destroy.
It might be a good idea to decide to *either* keep the first control plane special and move the destroy provisioner there as well *or* to try to keep "cluster provisoning" as separate as possible from "node provisoning". I personally would prefer the latter approach because I think it's semantically clearer and might make it easier to allow things such as re-provisioning the first control plane later on.

Another "edge-case" I ran into is the fact that the destroy provisioner also runs whenever the null_resource gets re-provisioned (because it changed), which deletes the load-balancer, which in turn makes the cluster pretty much unusable. (I am not sure how to bring CCM to re-create the load-balancer). we should figure out how to handle this before a merge ;)

mysticaltech:
@phaer To get the load balancer back, all you need to do is reapply the traefik_config, provided something has changed I would guess. It is has changed, it recreates it, I have tried yesterday, and found this to be the only obvious way.

mysticaltech:
Great ideas @phaer. I love the direction you are taking with node provisioning. I feel that first_control_plane has done its time, haha, if we could kill it in this PR, it would be great! 

@mnencia Also curious to see what you have in store for us about the subject.

mysticaltech:
Please feel free to overwrite my changes folks, we're just all seeking the best path here.

mysticaltech:
Removed the null resource, as indeed it's not necessary to get the Traefik IP at all. The destroy anywhere on master.tf, or even the module did not work, it gave terraform errors. So removed it completely, for now. 

And I sense that another null resource may be needed when we kill first_control_plane for good, and we could reintroduce it there.

mysticaltech:
This works for now folks and is simple. So merging, for now, we can always change or make it better later.

PR #90, opened 2022-02-22
Update agents.tf

TimHeckel:
Ensuring the right agent server type is attaching - I tore down and rebuilt my cluster and was noticing that only the control plane server type was being honored from the tfvars.


mysticaltech:
Good catch! Thank you :)

mysticaltech:
Oh wow, did not know that we can upgrade server types without destroying the nodes! Beautiful. 

FYI @phaer and @mnencia.

ksnip_20220222-030509

mnencia:
Yep! IIRC it works unless you are trying to reduce the disk size.

PR #92, opened 2022-02-22
First control plane node is not special anymore

mnencia:
The first control plane node is now identical to any other server
node. The cluster initialization happens once in two steps: first, make
sure that the k3s cluster is initialized and then apply our
configurations while the other nodes join. This change makes the
initialization more resilient and even faster than before.

### DIFF
--- a/init.tf
+++ b/init.tf
@@ -120,8 +120,8 @@ resource "null_resource" "kustomization" {
   provisioner "remote-exec" {
     inline = [
       "set -ex",
-      "kubectl -n kube-system create secret generic hcloud --from-literal=token=${var.hcloud_token} --from-literal=network=${hcloud_network.k3s.name}",
-      "kubectl -n kube-system create secret generic hcloud-csi --from-literal=token=${var.hcloud_token}",
+      "kubectl -n kube-system create secret generic hcloud --from-literal=token=${var.hcloud_token} --from-literal=network=${hcloud_network.k3s.name} --dry-run=client -o yaml | kubectl apply -f -",
+      "kubectl -n kube-system create secret generic hcloud-csi --from-literal=token=${var.hcloud_token} --dry-run=client -o yaml | kubectl apply -f -",
     ]
   }
 
--- a/init.tf
+++ b/init.tf
@@ -1,54 +1,24 @@
-module "first_control_plane" {
-  source = "./modules/host"
-
-  name = "k3s-control-plane-0"
-
-  ssh_keys               = [hcloud_ssh_key.k3s.id]
-  public_key             = var.public_key
-  private_key            = var.private_key
-  additional_public_keys = var.additional_public_keys
-  firewall_ids           = [hcloud_firewall.k3s.id]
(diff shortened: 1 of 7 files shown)


locals.tf
```diff
@@ -22,7 +22,9 @@ locals {
     # prepare the k3s config directory
     "mkdir -p /etc/rancher/k3s",
     # move the config file into place
-    "mv /tmp/config.yaml /etc/rancher/k3s/config.yaml"
+    "mv /tmp/config.yaml /etc/rancher/k3s/config.yaml",
+    # if the server has already been initialized just stop here
+    "[ -e /etc/rancher/k3s/k3s.yaml ] && exit 0",
```
mysticaltech, line 27: Yes Sir!

mnencia: It always happens when you move something just before pushing :P

servers.tf
```diff
@@ -41,16 +41,16 @@ resource "null_resource" "control_planes" {
   provisioner "file" {
     content = yamlencode({
       node-name                = module.control_planes[count.index].name
-      server                   = "https://${local.first_control_plane_network_ip}:6443"
+      server                   = "https://${element(module.control_planes.*.private_ipv4_address, count.index > 0 ? 0 : 1)}:6443"
       token                    = random_password.k3s_token.result
       cluster-init             = true
```
mysticaltech, line 46: @mnencia Why does cluster-init does not have a condition on it, it should not be true unless for the first control plane, unless I'm missing something?

mnencia: I didn't change that part. It was always `true` even on servers, so I left it that way. I suppose we can remove it in server.tf, and leave it only in init.tf.

mysticaltech: Or it should be removed completely here, no?

mysticaltech: I saw it's on master too, I have no idea how it got there 🤦 I will remove it ASAP, so you can remove it here too.

mysticaltech:
Wonderful - works like a charm! Merging now.

init.tf
```diff
@@ -136,8 +120,8 @@ resource "null_resource" "first_control_plane" {
   provisioner "remote-exec" {
     inline = [
       "set -ex",
-      "kubectl -n kube-system create secret generic hcloud --from-literal=token=${var.hcloud_token} --from-literal=network=${hcloud_network.k3s.name}",
-      "kubectl -n kube-system create secret generic hcloud-csi --from-literal=token=${var.hcloud_token}",
+      "kubectl -n kube-system create secret generic hcloud --from-literal=token=${var.hcloud_token} --from-literal=network=${hcloud_network.k3s.name} --dry-run=client -o yaml | kubectl apply -f -",
+      "kubectl -n kube-system create secret generic hcloud-csi --from-literal=token=${var.hcloud_token} --dry-run=client -o yaml | kubectl apply -f -",
```
phaer, line 124: I was struggling to understand that trick, but I guess this one is idempotent, which the former wasn't. Correct, @mnencia?

Great PR! :tada:

PR #96, opened 2022-02-23
[WIP] nodepools feature

jodhi:
nodepools feature: for https://github.com/kube-hetzner/kube-hetzner/issues/65

So we use the node name such as "worker-1" for indexing the module.agents. 
This PR can be expanded to cover different location as well

**What missing?**
If we create another nodepool, the IP will collide (since we use index+513 😅). Still thinking how to make the IP unique and not random

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -30,4 +30,14 @@ locals {
   install_k3s_server = concat(local.common_commands_install_k3s, ["curl -sfL https://get.k3s.io | INSTALL_K3S_SKIP_SELINUX_RPM=true INSTALL_K3S_SKIP_START=true INSTALL_K3S_EXEC=server sh -"])
 
   install_k3s_agent = concat(local.common_commands_install_k3s, ["curl -sfL https://get.k3s.io | INSTALL_K3S_SKIP_SELINUX_RPM=true INSTALL_K3S_SKIP_START=true INSTALL_K3S_EXEC=agent sh -"])
+
+  agent_nodepools = merge([
+    for nodepool_name, nodepool_obj in var.agent_nodepools : {
+      for index in range(nodepool_obj.count) :
+      format("%s-%s", nodepool_name, index) => {
+        index : index, # just for the compatibility with previous structure
+        server_type : nodepool_obj.server_type
+      }
+    }
+  ]...)
 }
--- a/variables.tf
+++ b/variables.tf
@@ -35,10 +35,6 @@ variable "control_plane_server_type" {
   type        = string
 }
 
-variable "agent_server_type" {
-  description = "Default agent server type"
-  type        = string
-}
 
 variable "lb_server_type" {
   description = "Default load balancer server type"
@@ -56,9 +52,12 @@ variable "servers_num" {
   type        = number
 }
 
-variable "agents_num" {
+variable "agent_nodepools" {
   description = "Number of agent nodes."
-  type        = number
+  type = map(object({
+    server_type = string
+    count       = number
+  }))
 }
 
 variable "hetzner_ccm_version" {
--- a/agents.tf
+++ b/agents.tf
@@ -1,9 +1,9 @@
 module "agents" {
   source = "./modules/host"
 
-  count = var.agents_num
-  name  = "k3s-agent-${count.index}"
+  for_each = local.agent_nodepools
 
+  name                   = each.key
   ssh_keys               = [hcloud_ssh_key.k3s.id]
   public_key             = var.public_key
   private_key            = var.private_key
@@ -12,8 +12,8 @@ module "agents" {
   placement_group_id     = hcloud_placement_group.k3s.id
   location               = var.location
   network_id             = hcloud_network.k3s.id
-  ip                     = cidrhost(hcloud_network_subnet.k3s.ip_range, 513 + count.index)
-  server_type            = var.agent_server_type
+  ip                     = cidrhost(hcloud_network_subnet.k3s.ip_range, 513 + each.value.index)
+  server_type            = each.value.server_type
 
   labels = {
     "provisioner" = "terraform",
@@ -24,28 +24,28 @@ module "agents" {
 }
 
 resource "null_resource" "agents" {
-  count = var.agents_num
+  for_each = local.agent_nodepools
 
   triggers = {
-    agent_id = module.agents[count.index].id
+    agent_id = module.agents[each.key].id
   }
 
   connection {
     user           = "root"
     private_key    = local.ssh_private_key
     agent_identity = local.ssh_identity
-    host           = module.agents[count.index].ipv4_address
+    host           = module.agents[each.key].ipv4_address
   }
 
   # Generating k3s agent config file
   provisioner "file" {
     content = yamlencode({
-      node-name     = module.agents[count.index].name
+      node-name     = module.agents[each.key].name
       server        = "https://${local.first_control_plane_network_ip}:6443"
       token         = random_password.k3s_token.result
       kubelet-arg   = "cloud-provider=external"
       flannel-iface = "eth1"
-      node-ip       = cidrhost(hcloud_network_subnet.k3s.ip_range, 513 + count.index)
+      node-ip       = cidrhost(hcloud_network_subnet.k3s.ip_range, 513 + each.value.index)
       node-label    = var.automatically_upgrade_k3s ? ["k3s_upgrade=true"] : []
     })
     destination = "/tmp/config.yaml"
--- a/output.tf
+++ b/output.tf
@@ -4,7 +4,9 @@ output "controlplanes_public_ip" {
 }
 
 output "agents_public_ip" {
-  value       = module.agents.*.ipv4_address
+  value = [
+    for obj in module.agents : obj.ipv4_address
+  ]
   description = "The public IP addresses of the agent server."
 }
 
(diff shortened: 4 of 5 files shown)


variables.tf
```diff
@@ -56,9 +52,12 @@ variable "servers_num" {
   type        = number
 }
 
-variable "agents_num" {
+variable "agent_nodepools" {
```
mysticaltech, line 55: If we could keep the agent_num, and use it as default values for a default agent node pool, it would be great!

locals.tf
```diff
@@ -30,4 +30,14 @@ locals {
   install_k3s_server = concat(local.common_commands_install_k3s, ["curl -sfL https://get.k3s.io | INSTALL_K3S_SKIP_SELINUX_RPM=true INSTALL_K3S_SKIP_START=true INSTALL_K3S_EXEC=server sh -"])
 
   install_k3s_agent = concat(local.common_commands_install_k3s, ["curl -sfL https://get.k3s.io | INSTALL_K3S_SKIP_SELINUX_RPM=true INSTALL_K3S_SKIP_START=true INSTALL_K3S_EXEC=agent sh -"])
+
+  agent_nodepools = merge([
+    for nodepool_name, nodepool_obj in var.agent_nodepools : {
+      for index in range(nodepool_obj.count) :
+      format("%s-%s", nodepool_name, index) => {
+        index : index, # just for the compatibility with previous structure
```
mysticaltech, line 38: Breaking changes are ok at this stage, we still do not support semver. Please remove all compatibility with previous versions, it's not needed. We need the best way only! :)

mysticaltech:
> If we create another nodepool, the IP will collide (since we use index+513 ). Still thinking how to make the IP unique and not random

About this, can't we just maintain a list (array) of IPs of a specific nodepool, and join one big list too with all IPs used. 

Then when a new IP of any kind is requested, we choose one randomly, test if present already, if present, we try again until we find one that is not present in the array used IPs. Is this possible?

mysticaltech:
Great PR @jodhi, can't wait for other members of the team to see it. @phaer, @mnencia please have a look when you can 🙏

mysticaltech:
> > **What missing?**
> > If we create another nodepool, the IP will collide (since we use index+513 sweat_smile). Still thinking how to make the IP unique and not random
> 
> What about this schema:
> 
> * 10.0.0.0/24 is kept for loadbalancers
> * 10.0.1.0/24 is for control-plane
> * 10.0.X.0/24 for the N-th node pool X=N+1 (2, 3, 4, ...)
> 
> This way node has a well-defined IP that you can calculate statically. The only thing that you should avoid is reordering the node pools smile

That is a lot cleaner indeed!

jodhi:
Its possible to have non-conflicted nodepool if we just let the hetzner handle the IP (simple changes + no manual ip calculation), for this changes, I already have the working code in my local
```
NAME                  STATUS   ROLES                       AGE     VERSION
big-0                 Ready    <none>                      3m29s   v1.22.6+k3s1
k3s-control-plane-0   Ready    control-plane,etcd,master   4m5s    v1.22.6+k3s1
k3s-control-plane-1   Ready    control-plane,etcd,master   3m19s   v1.22.6+k3s1
k3s-control-plane-2   Ready    control-plane,etcd,master   2m48s   v1.22.6+k3s1
small-0               Ready    <none>                      3m27s   v1.22.6+k3s1
small-1               Ready    <none>                      3m28s   v1.22.6+k3s1
```

> What about this schema:
> 10.0.0.0/24 is kept for loadbalancers
> 10.0.1.0/24 is for control-plane
> 10.0.X.0/24 for the N-th node pool X=N+1 (2, 3, 4, ...)
> This way node has a well-defined IP that you can calculate statically. The only thing that you should avoid is reordering the node pools 😄
> 

That schema also possible but not sure good idea or not (cmiiw)
pros:
- looks clean 
- the ip is predictable 
cons:
- limiting nodepool counts + node inside nodepool also limited

For the approach to implement that schema, what i have in my mind (to avoid deleting one nodepool impacting other nodepool):
run shell script locally ([using this](https://github.com/matti/terraform-shell-resource))  using input from the previous output.
Then that script will generate output of jsonencoded : `{1:"small-nodepool":,2:"big-nodepool"}`
Then if we remove small-nodepool, then add another nodepool, lets say medium-nodepool, the result will be `{1:"medium-nodepool":,2:"big-nodepool"}`

what do you guys think?

mysticaltech:
Sounds very interesting @jodhi, happy to see you are finding innovative solutions for this problem, please update the PR with your new branch, will take a good look tonight! 🙏

jodhi:
updated. feel free to check.
changes:
- No hardcoded internal IP for servers and agents
- support multiple nodepools

mysticaltech:
I like the elegance of it! So @jodhi, no more use of bash then, correct? One more question, the load balancer takes a random IP usually in the 10.0.0.x, we have set cp and agent nodes on 10.0.1.x, and 10.0.2.x respectively to avoid collisions with the LB IP, what can we do here to avoid collision too? Am not sure I understand fully how the IPs are determined, if Hetzner gives them, then it should be automatically good indeed.

Could you also, fix the conflict with master, very simple. And also, remove agent_num, finally, it's not needed, we will just have one default node pool declared and that's it, please! 🙏

phaer:
If I read this right, the IPs are allocated by hetzner via [`hcloud_server_network`](https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs/resources/server_network), which is a nice addition! I believe there could be a small risk of collisions with control plane IPs, depending on how clever Hetzner handles that.

But I think that could easily avoided by using the same `hcloud_server_network` mechanism for control planes as well, which would also make the code a bit more consistent and easier to understand.

Using different subnets per IP role still makes sense to me, because it eases custom firewall setups and so on. That should easily be do-able by using `cloud_network_subnet` and the `subnet_id` parameter for `hcloud_server_network`. 

All in all a great PR, thanks @jodhi! :tada: :100: Would you change control planes as well and add subnets?

modules/host/main.tf
```diff
@@ -12,11 +12,6 @@ resource "hcloud_server" "server" {
 
   labels = var.labels
 
-  network {
-    network_id = var.network_id
```
mnencia, line 16: Am I missing something or we should put the subnet id here? Otherwise how they know which subnet to attach?

phaer: Exactly, that's what I meant above. I believe they aren't attached to a specific subnet atm.

mysticaltech: They actually are folks, one subnet is mandatory on for Hetzner private networks. You can replace the above with the subnet id, it doesn't matter, this is what is done under the hood.

ksnip_20220225-002653

mysticaltech: They actually are folks, one subnet is mandatory on for Hetzner private networks. You can replace the above with the subnet id, it doesn't matter.

mysticaltech: This was actually the main point that made this project cool, I enforced since the beginning that all intra cluster communications happen on the private network, so the main subnet. We were even using Cilium at some point.

mysticaltech: Now the question I have is, will multiple subnets work? If they do, it's great.

mysticaltech: Here's the source of the above https://docs.hetzner.com/cloud/networks/faq/#what-are-subnets. Based on reading this again, I guess that manually creating subnets with https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs/resources/network_subnet should work!

mysticaltech: In recap, @jodhi, if you have the courage to undertake this, I propose that:
- The server blocks in nodes definitions would need to drop their current network {} config, this is optional anyway.
- "hcloud_network" "k3s" would stay defined.
- "hcloud_network_subnet" "k3s" would go, and be replaced by the many subnets aforementioned.
- hcloud_server_network would be created for each node, and be attached to the `subnet_id`, with network_id not specified.

jodhi: > One hcloud_server_network would be created for each node, and be attached to the correct subnet_id, with network_id not specified.

Ah right, just realize that we can attach machine to specific subnet too 😎 . Multiple subnet also very make sense (one for LB, one for control plane, and multiple for nodepool [similar with GKE]). Definitely will add this. Nice pointer with `hcloud network describe k3s`

mysticaltech: Yes, exactly, just that for the LB, I believe we can't choose one as its IP gets assigned automatically (plus no [available annotations](https://pkg.go.dev/github.com/hetznercloud/hcloud-cloud-controller-manager/internal/annotation#Name) for it) - or the CCM will probably put it in the last created subnet, or the first. So don't worry about the LB too much now, we can always adjust later on by maybe creating a default empty subnet and ensuring the order of creation!

jodhi: updatedd, so now each nodepools can be on different subnet (the limit is the subnet size). feel free to check.thanks

Why the ip calculated manually? turns out the [hetzner terraform has weird issue](https://github.com/hetznercloud/terraform-provider-hcloud/issues/393) when you assign machine to specific subnet, the ip can be different 😞 , this make things trickier

mysticaltech: Thanks for pushing the envelope @jodhi, looking great, will try tonight! 🙏

mysticaltech:
@jodhi I have just nothing to add, it's beautiful and works great for me! Well done :) Just waiting for my colleagues to review and approve.

@mnencia @phaer FYI, all green and good to go on my end, please try it.

phaer (approved):
Great code! Works fine for me with a single node-pool! :tada:

mnencia (approved):

mnencia:
I tested it with several configurations. The code works like a charm. We could improve the documentation, but we can do it in another commit.

jodhi:
Thanks for the invitation @mysticaltech , i feel honored invited to this great project. But i won't be able to accept the invitation for now. Keep up the good work! 💪

mysticaltech:
No problem @jodhi, I understand fully, time is precious ressource ! :) However, never hesitate to submit PRs, and be well 🙏

PR #97, opened 2022-02-24
Auto destroy LB on terraform destroy

mysticaltech:
Hey folks, this build-up on the good work of @phaer in previous commits.

It was achieved with a simple null resource that triggers on the change of the token, and my testing indicated that it's also necessary to delete the network.

The end result is that it works well and the destroy is smooth. Even in the case of an eventual error, it will continue, if the LB was not there for instance, or the hcloud command was not installed.

ksnip_20220224-014430

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -159,19 +159,29 @@ data "hcloud_load_balancer" "traefik" {
   depends_on = [null_resource.kustomization]
 }
 
-resource "null_resource" "destroy_lb" {
+
+resource "null_resource" "destroy_traefik_loadbalancer" {
+  # this only gets triggered before total destruction of the cluster, but when the necessary elements to run the commands are still available
   triggers = {
-    token = random_password.k3s_token.result
+    kustomization_id = null_resource.kustomization.id
   }
 
   # Important when issuing terraform destroy, otherwise the LB will not let the network get deleted
   provisioner "local-exec" {
-    when    = destroy
-    command = <<-EOT
-      hcloud load-balancer delete traefik
-      hcloud network delete k3s
+    when       = destroy
+    command    = <<-EOT
+      kubectl -n kube-system delete service traefik --kubeconfig ${path.module}/kubeconfig.yaml
     EOT
-
     on_failure = continue
   }
+
+  depends_on = [
+    local_file.kubeconfig,
+    null_resource.control_planes[0],
+    hcloud_network_subnet.k3s,
+    hcloud_network.k3s,
+    hcloud_firewall.k3s,
+    hcloud_placement_group.k3s,
+    hcloud_ssh_key.k3s
+  ]
 }
--- a/main.tf
+++ b/main.tf
@@ -158,3 +158,20 @@ data "hcloud_load_balancer" "traefik" {
 
   depends_on = [null_resource.kustomization]
 }
+
+resource "null_resource" "destroy_lb" {
+  triggers = {
+    token = random_password.k3s_token.result
+  }
+
+  # Important when issuing terraform destroy, otherwise the LB will not let the network get deleted
+  provisioner "local-exec" {
+    when    = destroy
+    command = <<-EOT
+      hcloud load-balancer delete traefik
+      hcloud network delete k3s
+    EOT
+
+    on_failure = continue
+  }
+}
--- a/modules/host/locals.tf
+++ b/modules/host/locals.tf
@@ -65,8 +65,7 @@ locals {
 rpm --import https://rpm.rancher.io/public.key
 zypper refresh
 zypper --gpg-auto-import-keys install -y https://rpm.rancher.io/k3s/stable/common/microos/noarch/k3s-selinux-0.4-1.sle.noarch.rpm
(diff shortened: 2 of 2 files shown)


mnencia:
This trick also doesn't work if the current `hcloud` context is incorrect. The best thing to do would be to delete the traefik service in the kube-system namespace and let the ccm to delete the real load-balancer.

main.tf
```diff
@@ -158,3 +158,20 @@ data "hcloud_load_balancer" "traefik" {
 
   depends_on = [null_resource.kustomization]
 }
+
+resource "null_resource" "destroy_lb" {
+  triggers = {
+    token = random_password.k3s_token.result
```
phaer, line 164: While I think it's clever to use the token as trigger to avoid deleting the load-balancer on node re-provisioning, I think a comment explaining why might sense here. That whole thing might not be intuitive for new users.

main.tf
```diff
@@ -158,3 +158,20 @@ data "hcloud_load_balancer" "traefik" {
...
+  }
+
+  # Important when issuing terraform destroy, otherwise the LB will not let the network get deleted
+  provisioner "local-exec" {
+    when    = destroy
+    command = <<-EOT
+      hcloud load-balancer delete traefik
+      hcloud network delete k3s
```
phaer, line 172: I assume you tested that? If so, I am positively surprised that it works that way, because I would expect the network deletion to clash with or at least confuse terraform?

phaer (approved):
Looks okay to me! Quite hacky to be honest, but if it works it's a step in the right direction.

That said, the approach @mnencia mentions sounds much better if easy to implement. If one of you - or anyone else - finds time to implement & test that, please do!

main.tf
phaer: Nevermind, you even pasted the tests output above ;) The "hcloud_network.k3s Destroying..." line does not seem to care whether the network actually still exists, even though it did still exist during terraforms planning phase.

mysticaltech:
Thanks, folks, will definitely test with `kubectl -n kube-system delete svc traefik`, my fear is that the CCM would take a long time to abide, but maybe not.

mnencia:
It has a finalizer, it should be synchronous

mysticaltech:
Did some changes folks, it works in deleting Traefik in a deterministic fashion. But the networks it's something else, they get deleted 5 minutes later, so either they do this because it's a fresh cluster, or it's just the IP routing table that takes time to refresh and clear.

ksnip_20220225-002653
ksnip_20220225-002637

mysticaltech:
Tried deleting the ccm with kubectl delete -f https://github.com/hetznercloud/hcloud-cloud-controller-manager/releases/latest/download/ccm-networks.yaml, but changed nothing. Am out of options, for now, it the network is slow to delete, folks will understand that they need to delete it manually.

I propose we merge, what do you think?

modules/host/locals.tf
```diff
@@ -66,6 +66,7 @@ rpm --import https://rpm.rancher.io/public.key
 zypper refresh
 zypper --gpg-auto-import-keys install -y https://rpm.rancher.io/k3s/stable/common/microos/noarch/k3s-selinux-0.4-1.sle.noarch.rpm
 udevadm settle
+exit 0
```
phaer, line 69: Is this related to the network destruction as well or is it in this PR by mistake? 
The shell script should exit with 0 by default anyway, if we - for some reason - need to ignore the return code of `udevadm settle`, `udevadm settle || true` might be more explicit.

main.tf
```diff
@@ -158,3 +158,30 @@ data "hcloud_load_balancer" "traefik" {
...
+  }
+
+  depends_on = [
+    local_file.kubeconfig,
+    null_resource.control_planes[0],
+    hcloud_network_subnet.k3s,
+    hcloud_network.k3s,
+    hcloud_firewall.k3s,
+    hcloud_placement_group.k3s,
+    hcloud_ssh_key.k3s
```
phaer, line 185: Shouldn't hcloud_network.k3s & local_file.kubeconfig be enough? I'd prefer to keep that as small as possible

phaer (approved):

modules/host/locals.tf
mysticaltech: Ah, during the testing the failed to unmount error at the end of ignition showed up again once. So added the exit 0. But definitely will use `udevadm settle || true` thanks!

main.tf
mysticaltech: We need to communicate from our localhost to the node, so the node and all the network infra better be intact before that happens, that's why I depend on all of this.

mysticaltech:
Ah, turns out if we let the cluster "settle" for at least 10 minutes, and then delete, the network goes away without resistance! 🥳 Otherwise, in testing scenarios, quickly create and delete,  it takes under 5 minutes to go, in which case, the tester can use his hcloud command rightfully, to speed things up a tiny bit 😂

ksnip_20220225-015024

mysticaltech:
That should be good enough folks, merging.

PR #98, opened 2022-02-24
added cloudflare dns provider support

michaelsstuff:
Hi @mysticaltech ,

not sure if you still want it, but here is an added example integration of cloudflare. 

Cheers
Michael

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ kubeconfig.yaml-e
 terraform.tfvars
 plans.yaml
 traefik_config.yaml
+/cloudflare.tf
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ _Please note that we are not affiliated to Hetzner, this is just an open source 
 - Ability to add or remove as many nodes as you want while the cluster stays running.
 - Automatic Traefik ingress controller attached to a Hetzner load balancer with proxy protocol turned on.
 - (Optional) Out of the box config of Traefik with SSL certficate auto-generation.
+- (Optional) Config for Cloudflare DNS configuration.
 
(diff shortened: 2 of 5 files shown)


mysticaltech:
It's great, thanks @michaelsstuff, appreciate it very much! About the Cloudflare record for the LB, why do you need it on top of external DNS (that configures ingresses automatically), just curious?

mysticaltech:
The reason I have hesitant about that, is that sometimes the IP changes if we update the number of nodes, for instance, it gets "refreshed" by the CCM and sometimes assigned a new IP.

michaelsstuff:
> 

Hmm sorry I cant quite follow. Assume for a moment I know nothing 🙃
The ingress does not have an inherit capability to change public dns records, rigth? I could run [external-dns](https://github.com/kubernetes-sigs/external-dns) which might be more flexible for deployments.

If so then the Cloudflare record for the LB would only be usefull to address your kube with a name... so not too usefull.

mysticaltech:
Basically, what I do, is connect external-dns to cloudflare, then on each ingress definition I create, I add these annotations:
```
    external-dns.alpha.kubernetes.io/hostname: "nameiwant.myrootdomain.dev"
    external-dns.alpha.kubernetes.io/cloudflare-proxied: "true"
```

So basically, and that in turn will point to the LB automatically. I never have to create a direct entry for the LB manually. You do not do this?

michaelsstuff:
Jup that makes a lot more sense. Just tried it, works like a charm.
Then lets throw this pr away :-D

Thank you very much for the hints and explanation!
Cheers
Michael

PR #101, opened 2022-02-27
Switch to cloud-init for host initialization

mnencia:


mysticaltech:
It's clean and beautiful, love it!

mnencia:
Before merging I want to add some comments to explain why it is done this way. I'll do it this evening

phaer:
Great, thanks!

Removing the ignition config also removes
`sshAuthorizedKeys = concat([local.ssh_public_key], var.additional_public_keys)` in there. But I see no equivalent inside the new cloud init files  - am I overlooking it or is it missing in this PR?

mnencia:
Good catch. I need to add it to cloud-init config.Il 27 Feb 2022 19:34, Paul Haerle ***@***.***> ha scritto:
Great, thanks!
Removing the ignition config also removes
sshAuthorizedKeys = concat([local.ssh_public_key], var.additional_public_keys) in there. But I see no equivalent inside the new cloud init files  - am I overlooking it or is it missing in this PR?

mysticaltech:
So @mnencia @phaer, just to clarify, `qemu-img` writes the image to the whole disk and even expands the partitions? No need for `btrfs filesystem resize max`?

mnencia:
> So @mnencia @phaer, just to clarify, `qemu-img` writes the image to the whole disk and even expands the partitions? No need for `btrfs filesystem resize max`?

It is done automatically by cloud-init at the first start. The same is true also with combustion: if you search the startup logs you see it tries to expand `/var` partition to maximum size.

mysticaltech:
Oh wow, ok, thanks for the explanation!

modules/host/main.tf
```diff
@@ -68,3 +65,33 @@ resource "hcloud_server_network" "server" {
...
+        sshAuthorizedKeys = concat([local.ssh_public_key], var.additional_public_keys)
+      }
+    )
+  }
+
+  # Initialization script (runs at every reboot)
+  part {
+    content_type = "text/cloud-boothook"
+    filename     = "boothook.sh"
+    content = templatefile(
```
phaer, line 90: I recommend inlining that in `main.tf` as that requires the same amount of lines there, but saves us yet another file.

Personally I'd also prefer to use yamlencode for `userdata.yaml` to protect us from the edge-cases of yaml + string substition, but as you like :)

phaer (approved):
Looks good to me! :+1:

PR #102, opened 2022-02-27
Extra firewall rules

mysticaltech:
Hey folks, this is to close https://github.com/kube-hetzner/kube-hetzner/issues/64. In the example, we open a port for Postgres :)

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -17,39 +17,35 @@ locals {
   csi_version   = var.hetzner_csi_version != null ? var.hetzner_csi_version : data.github_release.hetzner_csi.release_tag
   kured_version = data.github_release.kured.release_tag
 
+  # The following IPs are important to be whitelisted because they communicate with Hetzner services and enable the CCM and CSI to work properly.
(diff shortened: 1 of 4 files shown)


locals.tf
```diff
@@ -17,6 +17,127 @@ locals {
...
+    # Allowing internal cluster traffic and Hetzner metadata service and cloud API IPs
+    {
+      direction = "in"
+      protocol  = "tcp"
+      port      = "any"
+      source_ips = [
+        var.network_ipv4_range,
+        "127.0.0.1/32",
+        "169.254.169.254/32",
+        "213.239.246.1/32"
```
phaer, line 30: Those IPs are hetzners metadata & cloud api? maybe an inline comment would be a good idea. Will we ever need to update them? If so a source link as well :)

phaer (approved):
Looks good to me, thanks! :rocket: 

Hard-coded Hetzner IPs could use a comment

locals.tf
mysticaltech: You are right! Just make it a tiny bit clearer, thanks!

PR #103, opened 2022-02-28
Switch to cloud-init for host initialization

mnencia:


joskuijpers:
Did you revert only to use a PR instead? I am using this repo as a base for my own setup using Kubic and k8s+calico so wonder why it was done.

mnencia:
The installation of k3s-selinux is not working in the cloud-init setup. Will investigate this evening why.

mysticaltech:
Folks, did some testing, and found out a few things:

- The kured reboot method, does not work, is not taken into consideration, with our without snapshots, it has nothing to do with it. But it's not needed, we can use the just rely on the reboot-needed flag file and it works as well.

- On MicroOS, whatever you do, if you install k3s-linux with the k3s installer, the selinux never gets applied, this command never gets executed successfully chcon -u system_u -r object_r -t container_runtime_exec_t ${BIN_DIR}/k3s, see https://get.k3s.io/.

- Turns out, k3s runs perfectly fine without SELinux, this is just an added security measure, that for us, never got used, because even when we installed k3s-selinux via combustion, we never manually applied the above command.

- For the above command to work however, the package has to be installed, and we need to have rebooted into the new snapshot. So if we want SELinux, rebooting is not an option. Turns, out even with reboot, it's under 5 min install time, so all good there.

- So we install k3s-selinux, reboot, then install k3s with skip rpm, and then apply the above command. And that command executes succesfully, meaning SELinux has been applied.

mysticaltech:
Turns out it's not about k3s-selinux being installed, it should be installed, after k3s, because the postinstall script in the RPM applies SELinux policies to all the k3s files and directories! So the order counts. 

And according, to the folks over at k3s, see the issue [mentioned above](https://github.com/k3s-io/k3s/issues/5191), it doesn't matter if it reboots in between to apply the policy (pending final confirmation).

Basically, the install.sh gets the order right! So that's the only way to go, with an immediate reboot after starting k3s.

ksnip_20220302-005740

mysticaltech:
Just pushed the final tweaks, now the order is correct and we reboot to apply the snapshot.

mysticaltech:
As for the reboot method, it just works, no need to worry about that too much. Probably the "kured" method was added before the kured reboot-sentinel config flag came about, that's a pretty recent addition.

But definitely good for MicroOS if we open a ticket and let them know, will let you do this @mnencia, as you dove deep down that issue.

modules/host/main.tf
```diff
@@ -8,63 +8,83 @@ resource "hcloud_server" "server" {
...
 
+  # Prevent destroying the whole cluster if the user changes
+  # any of the attributes that force to recreate the servers.
+  lifecycle {
+    ignore_changes = [
+      location,
+      ssh_keys,
+      user_data,
+    ]
+  }
```
phaer, line 23: Is this what we want? Then I'd suggest to mention that behavior in our README, as I believe it's somewhat unexpected for terraform users. All of those attributes should not change too often during normal operation, but if they do I'd expect my nodes to be re-provisioned.

An interesting topic to explore would be whether terraform enables us to drain & re-provision nodes one-by-one in a single terraform apply run, so that we could change e.g. "location" but wait for each node to be operational before killing the next one. Could be difficult and would make the run much longer but it would be awesome to apply updates to this repo to an existing cluster without downtime

phaer (approved):
Looks good to me! I've tested it by first applying it to an existing cluster (current master) and then on a new cluster. Haven't checked the new upgrade logic yet, but everything else seems to run fine. 

Maybe we should mention that updating kube-hetzner after merging this and then running `terraform apply` will be a noop (at least it was for me), due to the lifecycle arguments (commented inline). So one would think everything is up to date, while still running nodes with the old ignition/combustion logic. This is fine for me, we should just keep it in mind while debugging upcoming issues in the near future and maybe announce it.

phaer:
Added a small fix to make filesystem resizing work (with /var/, not /)

modules/host/main.tf
mysticaltech: Makes sense! It looks good to me personally, but mentioning it in the readme seems ok if you feel the need for it. But this is @mnencia's addition, so will leave him answer that better.

phaer: Yes I got that, tried to address that in the second paragraph. I still think documenting it for users is a good short-term solution

phaer: Added a note regarding the fact that those values are ignored if nodes are already provisioned. 

But doesn't that behavior of effectively deleting everything deployed to the cluster also apply to changing things like network-location or ip subnets? Should we ignore them too or just leave the lifecycle thing and put a general warning in the README? Something like "Be extra careful when approving changes where terraform (destroys and) re-creates all nodes at once as that will result in data loss on the cluster. Cordon, drain and re-provision each node one-by-one if needed"?

mnencia: The other fields doesn't require a redeploy.

mysticaltech:
Ok, folks, so it the last commits:

- The hostname was fixed for good, it was indeed DHCP @mnencia, all that was needed was to set a flag. 
- DNS was set to use Google DNS servers primarily, and Hetzner comes up third as a backup.
- Snapshot numbers are reduced.
- Finally, `k3s-selinux` gets installed with a compatible version, and gets applied post k3s installation successfully, before starting k3s.

mysticaltech:
Finally, there is no need even to define any reboot method. For one, it does not seem to be read, for two, relying on the default works great. Here's why: 

- The default reboot method is "auto", and it tests for the presence of the `rebootmgrctl` binary. In our case it is present, so `rebootctlmgr` is selected automatically. By default. And we set its strategy to off explicitly.

- The default path to the reboot flag file is `/var/run/reboot-needed`.

Those are deterministic conditions. It just works and with fewer lines of code for us!

ksnip_20220303-031054

mysticaltech:
All seems clear on my end. Now I just need to give it 24h to 48h to confirm automatic reboot upon upgrade. I want to see the real deal, not just the simulation.

kustomize/kured.yaml
```diff
@@ -18,3 +18,4 @@ spec:
           command:
             - /usr/bin/kured
             - --reboot-command=/usr/bin/systemctl reboot
+            - --reboot-sentinel=/var/run/reboot-needed
```
mnencia, line 21: I would prefer to not do this. We are short-circuiting one of the steps in microOS reboot management. The transactional-update.service unit already execute 

```
/usr/sbin/transactional-update cleanup ${UPDATE_METHOD} reboot
```

So the configuration on the main (REBOOT_METHOD=kured and the standard file) works like a charm.

If the user manually installs things using transactional-update, is their duty to also invoke `transactional-update reboot` to trigger a reboot.

.gitignore
```diff
@@ -6,3 +6,4 @@ kubeconfig.yaml-e
 terraform.tfvars
 plans.yaml
 traefik_config.yaml
+kured.yaml
```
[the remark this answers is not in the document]
mysticaltech: Ah, that's just for my local file, it does not ignore `templates/kured.yaml`, does it?

kustomize/kured.yaml
mysticaltech: Oh wow, I did not remember this in previous images. If you are sure of the above, will change.

mysticaltech: It definitely does make sense! Thanks for getting to the bottom of this 🙏 Will give it a shot and merge.

.gitignore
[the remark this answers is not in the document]
mnencia: It ignores any file in the repo called `kured.yaml`, no matter the directory. If you want to ignore something in your repo only you can add it to `.git/info/exclude`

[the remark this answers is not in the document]
mysticaltech: Ok, thanks for clarifying!

kustomize/kured.yaml
mysticaltech: Yes Sir!! 😎

ksnip_20220304-143948

mysticaltech:
Thanks, @mnencia and @phaer, great collab as always!

PR #108, opened 2022-03-03
Improve support for single-node clusters

phaer:
Hi,

As suggested by @mysticaltech we use k3s klipper/"service load balancer" instead of traefik here and adapt a few variables, including firewall rules.

This is based on the `cloud-init2`-branch so should be merged after that one.

Not yet sure if auto-upgrades (with downtime) work correctly with this, could you verify @mysticaltech :pray: 

EDIT: force-pushed for rebase


mysticaltech:
Looking good @phaer, what issue did you encounter?

phaer:
Nothing much to speak of, needed a bit to figure out some terraform quirks, got confused with traffik for the moment but everything went well

mysticaltech:
Ah ok, I must have misunderstood you on matrix. Awesome then, will do one quick round of testing then merge!

mysticaltech:
Hey @phaer, works like a charm unless you try to scale up? 

ksnip_20220304-095114

So maybe we need to decouple, the use of Hetzner LB or Klipper LB, from the node number, and have a flag for it. Please think about that more. Because it would be ideal for something using Kube-Hetzner in single mode, to be able to scale up in a flexible fashion, though it adds some complexity.

mysticaltech:
So maybe we need to decouple, the use of Hetzner LB or Klipper LB, from the node number, and have a flag for it. 

Please think about that more. Because it would be ideal for something using Kube-Hetzner in single mode, to be able to scale up in a flexible fashion.

phaer:
Ah yes, scaling up isn't supported at  the moment - one would have to re-provision for that. 

Can look into it, but not sure how much sense that makes. A single-node "cluster" is only good for development and testing IMO. Besides traefik/klipper, the allow_scheduling_on_control_planes var must logically be true on single-node clusters, but isn't best practice as soon as you have enough nodes to avoid it.

mysticaltech:
Ok... I'm perfectly fine with imposing a re-provisioning for that. But can't we show that in nicely, instead of showing an error. Like just lock it, or output a controlled error?

Also, wouldn't it be nice to support explicitly choosing Hetzner LB or Klipper, whatever the number of nodes, we can leave this for another PR, but curious on your opinion.

phaer:
Just thought about it a bit more and I think gracefully scaling up is pretty much impossible unless we solve a similar problem to the one we ran into with changes to location, ssh-keys, etc in the cloud-init PR (lifecycle argument). That's because we would need to re-provision the first (and only) control plane without servicelb and with our traefik kustomization **after** we deployed the other nodes. 

Going to look into how to provide an error message in this case, though

mysticaltech:
You are right, will just merge then we improve in time.

PR #111, opened 2022-03-04
added traefik option for 'insecureSkipVerify'

michaelsstuff:
Hi,

small addition to add [insecureskipverify ](https://doc.traefik.io/traefik/routing/overview/#insecureskipverify) to traefik. 
This seems to be necessary if you configure an ingres, that connects to a service which itself also has a TLS service running with private certificates. 

It seems not many people do this, but with it you would come a lot closer to true end-to-end encryption. 

This is not the perfect solution, as traefik will now accept any certificate. I am looking into better ways, but this is nice workaround. 

Cheers & best regards
Michael


phaer:
Hi,

I am open to a PR which allows adding arbitrary kustomization resources, where you could add any traefik options you like.

Adding just `insecureSkipVerify` isn't really worth it IMO, as it's easy to add trusted certificates via LetsEncrypt and others for pretty much any public service nowadays. I'd recommend making traefik trust your self-signed cert/ca instead, as proxying to TLS services without certificate verification is almost as bad as not using TLS at all.

phaer:
Hi Michael,

I forgot to mention that you should also be able to just re-deploy your customized traefik config after provisioning the cluster. That's an extra command, but saves you from needing a fork until someone implements the more generic approach above. Thanks for your contribution!

mysticaltech:
@michaelsstuff What I do personally, it copy templates/traefik_config-yaml.tpl to my root folder, you can name it `traefik-custom.yaml` (it's already added to the gitignore) and customize it the way you want! That's what I do at least.

mysticaltech:
Oh yeah!
ksnip_20220305-002546

michaelsstuff:
Hi,

remove the single option and tried to make it work with any generic traefik cli option. I tested it with an empty list an with two arguments and it worked fine for me, but please test it. 

Thanks for all the great feedback here!

phaer:
@michaelsstuff Thanks! Please don't mix naming-conventions in a single variable name, so `traefik_additional_options` instead of `traefik_additionalOptions`. 
Otherwise looks good and we could merge it!

mysticaltech:
Hey @michaelsstuff, thank for your this, looking good and I want to merge but the terraform check fails because of the format, just run `terraform fmt`  and re-submit. Thanks!

michaelsstuff:
Hi @mysticaltech , thank you, did run the format and pushed it. 
There I go yet learning terraform :D

mysticaltech:
Thanks @michaelsstuff ! 🙏

Issue #109, opened 2022-03-03
Secret in TLS example


MartiniMoe:
The TLS example configures a secret while at the same time providing annotations to let traefik request a certificate from lets encrypt:
https://github.com/kube-hetzner/kube-hetzner/blob/6f6de884ec1baace14b894e5ee1917ffa947e1ca/examples/tls/ingress.yaml#L12

If I get this right the line referencing the secret is wrong here? If thats the case I could do a pull request to fix the documentation and example.

mysticaltech:
@MartiniMoe Yes, basically you also need to set this is your terraform.tfvars:

```
traefik_acme_tls = true
traefik_acme_email = "mail@example.com"
```

Or your cluster is already deployed, just copy this file https://github.com/kube-hetzner/kube-hetzner/blob/master/templates/traefik_config.yaml.tpl, and change it directly and re-apply.

As for the secrets, le will just use the name mentioned to create one for you. Am I missing something?

MartiniMoe:
Hmm, when I tried it like this I got errors from Traefik saying the secret does not exist.
When I deleted the line containing the secret it works.

MartiniMoe:
Okay, I created a pull request :) See #112

PR #112, opened 2022-03-04
Remove secret from TLS example

MartiniMoe:
This removes the secret from the TLS example. Otherwise you'll get errors that the secret does not exist, see #109.


phaer:
Thanks for your contribution @MartiniMoe!

PR #113, opened 2022-03-04
add random pet names for cluster & nodes

phaer:
Hi,

This is a quick implementation of a unique suffix for all resources in a cluster and then another one for each node. 

Needs more testing, but a first cluster deployed successfully and seems operational :)


mysticaltech:
Great work man! Testing now.

mysticaltech:
At first, it was a bit long, so removed the cluster name from the node name as this is not needed. Remove the index from the node name for agents, and finally set the length of the generated pet name to 1.

I love it, man, it adds uniqueness, while still being friendly and ensures that the sequentially is broken! The cluster name and node name also hold when we add or remove nodes, so IMHO, this is good to go! Well done :)

ksnip_20220304-235414

mysticaltech (approved):
Awesome work @phaer! Thanks.

phaer:
> At first, it was a bit long, so removed the cluster name from the node name as this is not needed. Remove the index from the node name for agents, and finally set the length of the generated pet name to 1.
Thanks, I was just about removing the index and shortening the length to one. But i'd still add the cluster-name, because I think with potentially multiple clusters in a hcloud project it's a nice feature to be able to associate nodes to their cluster just by name. I think one extra word in the name is worth it, agree?

mysticaltech:
@phaer Can we add it to Hetzner and not to the cluster? That should be doable.

mysticaltech:
Basically, to the k3s config node name, and to the hostname is user-data, we do not path the cluster name. But to the Hetzner config, yes! That would be ideal.

mysticaltech:
@phaer For the Hetzner server name, on top of the cluster name, we could even add a prefix like "k3s-" to help identify it better. Especially if you have many servers in use, which would be the case I imagine. It's also good that it will keep consistent with the other created ressources.

phaer:
But then host & kubernetes node names would not match the names in hetzners webinterface anymore, which could be confusing? Its just a single word now, so adding it doesn't increase name length by too much

mysticaltech:
They would match!

In Hetzner you will see:
`k3s-superb-bear-control-plane-elephant
`
In the cluster, it will be shorter:
`control-plane-elephant
`

phaer:
Thats not a match IMO, because `control-plane-elephant ` does not tell you which cluster it belongs to. And it makes scripting somewhat more cumbersome.

mysticaltech:
But when you `kubectl` into a cluster, you already know which one you are hitting. No?

mysticaltech:
So in recap, the way I see it now:

- Hetzner name: k3s-superb-bear-control-plane-elephant
- Hostname: k3s-superb-bear-control-plane-elephant
- Kube node name: control-plane-elephant

For those that want the full hostname in Kube, we can use a flag.

mysticaltech:
Hey @phaer, did some testing, and finally had to go with a flag, as the 3 names must be the same otherwise the CCM fails to work properly. 

So now, you can set `use_cluster_name_in_node_name = true`, and this will make it work the way you need with the cluster name included in the hostname. 

_Now if you want the exact same format as before, then we can replace the above flag with a `node_name_config` flag, that can take three possible values, and we test for those and give the right format._

I believe there is absolutely no harm in this. Flexibility in such a project is way better than rigidity. It actually makes it more valuable because it answers different needs and tastes!

ksnip_20220305-040129

mysticaltech:
Apart from the name, when I tried to scale, I got this error. It seems that `random_pets.agents` and `local.agent_pools` are not in sync.

ksnip_20220305-043850

ksnip_20220305-043440

mysticaltech:
I think it got introduced when I removed the index from the name, but not 100% sure. Ideally, the key of `local.agent_nodepools` should match the names of the nodes, but I do not know how to do that!

mysticaltech:
Small tweaks, and fixed an issue in master with the LB (still) taking other subnet IPs sometimes, it was missing a subnet, poor thing! :)

mysticaltech:
And getting used to the full names, I like getting to know my cluster name after all, but happy that it's optional!

ksnip_20220305-055701

mysticaltech:
I would love to merge this now @phaer, I have done the tests upscale, downscale, change in number. All passed with flying colors! The create_before_destroy is another topic, ideal for another PR. I propose we merge as-is, it's pretty stable on my end!

phaer:
@mysticaltech Feel free to merge! :)

PR #118, opened 2022-03-09
Automatic subnets creation

mysticaltech:
Hey folks, the subnets did not need manual input. Potentially this would have been error-prone and could be simplified. I used [cirdsubnet](https://www.terraform.io/language/functions/cidrsubnet) to generate them automatically. It creates the same kind of subnets we previously defined, while reducing the number of lines in the code, and the work required on the part of the user.

ksnip_20220309-012120

### DIFF
--- a/terraform.tfvars.example
+++ b/terraform.tfvars.example
@@ -28,7 +28,7 @@ control_plane_server_type = "cpx11"
 
 # As for the agent nodepools, below is just an example, if you do not want nodepools, just use one,
 # and change the name to what you want, it need not be "agent-big" or "agent-small", also give them the subnet prefer.
-# For single node clusters set this equal to [] or just set the counts to 0
+# For single node clusters set this equal to [] or just set the counts to 0.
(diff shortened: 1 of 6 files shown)


terraform.tfvars.example
```diff
@@ -37,28 +28,27 @@ control_plane_server_type = "cpx11"
...
-  agent-big = {
-    server_type = "cpx21",
-    count       = 1,
-    subnet      = "agent_big",
-  }
-  agent-small = {
+# For single node clusters set this equal to [] or just set the counts to 0.
+# IMPORTANT: Once the cluster is created, you can change the count, and even set it to 0, but do not remove a nodepool from the list. 
+# You can add others at the end of the list if you want.
+agent_nodepools = [
```
mysticaltech, line 34: I have changed this to a list instead of a map because now the order is fixed. As before with the map, the order is random, and it can create problems when changing counts or adding new node pools. Now it's completely predictable.

mnencia:
Tested it, works like a charm. Merging.

PR #119, opened 2022-03-09
Reducing the keepers list

mysticaltech:
Hey folks, here's why am doing so, please let me know if I am wrong. https://www.loom.com/share/ed1de222a9364f44af0723b150f29577

### DIFF
--- a/modules/host/main.tf
+++ b/modules/host/main.tf
@@ -2,21 +2,12 @@ resource "random_pet" "server" {
   length = 1
   keepers = {
     # We re-create the id (and server) whenever one of those attributes
-    # changes. This should include all input variables to this module,
-    # but NO SENSITIVE values as they might be logged here.
+    # changes.
     name                   = var.name
     public_key             = var.public_key
     additional_public_keys = join(",", var.additional_public_keys)
-    firewall_ids           = join(",", var.firewall_ids)
     placement_group_id     = var.placement_group_id
-    labels                 = join(",", [for k, v in var.labels : "${k}=${v}"])
-    ipv4_subnet_id         = var.ipv4_subnet_id
     private_ipv4           = var.private_ipv4
-    server_type            = var.server_type
-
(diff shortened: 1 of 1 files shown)


mysticaltech:
Hey @phaer, when you can please have a look at my explanations above, the video is only 3 minutes. And we can always bring them back later if I missed something.

modules/host/main.tf
```diff
@@ -7,21 +7,12 @@ resource "random_string" "server" {
...
   keepers = {
     # We re-create the id (and server) whenever one of those attributes
-    # changes. This should include all input variables to this module,
-    # but NO SENSITIVE values as they might be logged here.
+    # changes.
     name                   = var.name
     public_key             = var.public_key
     additional_public_keys = join(",", var.additional_public_keys)
-    firewall_ids           = join(",", var.firewall_ids)
     placement_group_id     = var.placement_group_id
```
mnencia, line 14: You can change a node placement group without recreating the node, and it will not affect the node as it is already provisioned. Still, it will affect any new node, taking into account the presence of the preexisting nodes in the placement group.

modules/host/main.tf
```diff
@@ -7,21 +7,12 @@ resource "random_string" "server" {
 
   keepers = {
     # We re-create the id (and server) whenever one of those attributes
-    # changes. This should include all input variables to this module,
-    # but NO SENSITIVE values as they might be logged here.
+    # changes.
     name                   = var.name
     public_key             = var.public_key
     additional_public_keys = join(",", var.additional_public_keys)
```
mnencia, line 13: I do not see great value in reprovisioning all the nodes to add/change an ssh key. It is something that you can change at any moment by pushing the list of keys using a remote file resource. init-cloud only pushed the keys during the first boot and will not interfere.

mysticaltech:
Thanks for your feedback @mnencia! It does make sense, so only kept the name and private_ipv4. It is true that if those change, regeneration is needed.

PR #120, opened 2022-03-09
Changed the cluster name to manual

mysticaltech:
The automatic pet name for both the cluster and nodes felt like overkill and I did not feel comfortable with them. Now setting the cluster name manually makes more sense, as the cluster is here to stay, its name can be meaningful. As the random pets are really random and meaningless, that suits more the ephemeral nature of nodes.

It defaults to "k3s" but can be anything like "k3s-cluster-xyz" as long as it is hostname compatible. It has the right checks for that.

### DIFF
--- a/agents.tf
+++ b/agents.tf
@@ -3,7 +3,7 @@ module "agents" {
 
   for_each = local.agent_nodepools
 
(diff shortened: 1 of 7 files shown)


mnencia (approved):

mnencia:
I agree with you. LGTM

PR #121, opened 2022-03-09
Random string node

mysticaltech:
This build upon https://github.com/kube-hetzner/kube-hetzner/pull/120.

After I saw this:

ksnip_20220309-054202

I decided that:

ksnip_20220309-055108

With a length 3, and limited to alphabetical lower case. It gives us 17576 possibilities while staying super easy to remember and to type.


mnencia (approved):

mnencia:
LGTM

Issue #122, opened 2022-03-09
Disable Traefik ingress controller & lb


oujonny:
Hi guys
Thanks to all contributors to this amazing project!

Is it possible to disable the treafik ingress controller at all? I would prefer using nginx-ingress or istio-gateway as ingress solution. For that, I didn't need the Loadbalancer and the treafik installation.

I'm also open to contributing such a toggle function but would need some inputs on how to implement it the best way. :)

cheers,
Johann Schley

mysticaltech:
@mnencia I believe that one is of interest to you too!

mnencia:
I think we should expose flags to disable traefik and also metrics server (you may want to use prometheus-adapter instead)

mysticaltech:
Definitely, that's a quick add too!

paulwollgast:
Hello guys, also from my side big thanks for that project and +1 to @oujonnys request, I'm also planning to use nginx and/or istio ingress so that would be nice!

mysticaltech:
@paulwollgast Thanks for sharing! Definitely, that's a low-hanging fruit, and will be done shortly.

oujonny:
Hi Guys
Thanks for all your anwnsers! 
I will open an PR now to introduce the possibility to disable traefik & metrics-server. By default traefik will be **disabled** and metrics-server enabled.

mysticaltech:
Closed in #127

PR #127, opened 2022-03-12
FEAT: Disable traefik & metrics-server

oujonny:
As disscused in #122 this PR introduces the possibility to disable traefik & metrics-server of k3s.
**Important: Traefik is now disabled by default!**

You can enable/disable these two features with those variables:

```
# If you want to use traefik ingress controller with a loadbalancer
# traefik_enabled = true
...
# If you want to enable k8s metric server or not
# metric_server_enabled = false
```

Additional it fixes an deprecated waring by switchting to local_sensitive_file resrouce for kubeconfig.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ _Please note that we are not affiliated to Hetzner, this is just an open source 
 - Proper use of the underlying Hetzner private network to remove the need for encryption and minimize latency.
 - Automatic HA with the default setting of three control-plane and two agents nodes.
 - Ability to add or remove as many nodes as you want while the cluster stays running.
-- Automatic Traefik ingress controller attached to a Hetzner load balancer with proxy protocol turned on.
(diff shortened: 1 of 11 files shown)


mysticaltech:
Well done @oujonny ! That clean code right there. Will just try it once ASAP and merge.

Will also change Traefik to on by default, as it's actually something that most users want and appreciate working out of the box. But definitely having the ability to disable if needed is fantastic! 👍

And thanks also for the metric server disable option. Two birds with one stone, that's awesome! 🙏

mnencia:
The patch needs a `terraform fmt` pass. Good to merge after this change and after setting traefik enabled by default.

mnencia:
The is another change to do: if traefik is disabled, we should return load_balancer_public_ipv4=null

mnencia (approved):

mysticaltech:
Thanks @oujonny, merging!

PR #131, opened 2022-03-15
Add testing infrastructure & single-node test

phaer:
Hello!

So this uses [terratest](https://terratest.gruntwork.io/) to setup a test suite for hetzner-kube based on golang. It can be run locally with "cd test && go test -v" if golang is installed, an SSH key exists in `~/.ssh/kube-hetzner{,.pub}` and `TF_VAR_hcloud_token` is set.

The test suite itself currently provisions a single-node setup to hetzner cloud and checks whether HTTP works. Additional test cases for multi-node clusters - and ideally updates from the last release - as well as a check if kubectl works and whether all pods are running/complete should be added in later PRs.

It also configures a github workflow which will run the test suite for tagged commits (but won't block the release for now). One reason is that this whole thing needs feedback & testing.

The other one is that the github workflow does **not pass yet** in [my personal fork](https://github.com/phaer/kube-hetzner/runs/5555207898) because it failed to connect to the target server via SSH twice.
<del>Interestingly enough it fails with another error if I run the github workflow in [act](https://github.com/nektos/act):
locally. `invalid character 'c' looking for beginning of value ` when parsing `terraform output` or something, might be a golang compatibility issue, I'll investigate. But as `go test` works, I thought I'll share it right away.</del>
EDIT: Works with act now, see new commits

To run the workflow locally with act
```
act --job "build" \
  --secret HCLOUD_TOKEN="..." \
  --secret SSH_PRIVATE_KEY="$(cat ~/.ssh/kube-hetzner)" \
  --secret SSH_PUBLIC_KEY="$(cat ~/.ssh/kube-hetzner.pub)"
```

# secrets

The workflow requires 3 credentials set in github secrets:
* `HCLOUD_TOKEN`
* `SSH_PRIVATE_KEY`
* `SSH_PUBLIC_KEY`
Those should of course be generated for this specific use-case if you want to test your personal forks. For `kube-hetzner` itself, we need to discuss how to finance testing hosts. I think only @mysticaltech has the permission to add a secret here, so maybe a project in your account could be used?

golang dependencies should now also be updated by dependabot.

Hope I didn't forget anything, looking forward to your feedback! :tada:


phaer:
With the new commits it's now running fine in an `act` environment as documented above:
```
act --job "build" \
  --secret HCLOUD_TOKEN="..." \
  --secret SSH_PRIVATE_KEY="$(cat ~/.ssh/kube-hetzner)" \
  --secret SSH_PUBLIC_KEY="$(cat ~/.ssh/kube-hetzner.pub)"
```
creates a new single-node cluster with configuration details currently hardcoded in the test file, waits for HTTP to work and destroys the cluster again. Additional checks and github ci are still to be done

mysticaltech:
I'm looking forward to this @phaer, as it's time-consuming for me to test all the different combinations for each PR.

mysticaltech:
@phaer Please to re-target to the `staging` branch! 🙏

PR #135, opened 2022-03-21
Improve stability of traefik load balancer IP

mnencia:
By default, if the helm chart installation or upgrade fails for any
reason, the whole helm chart is deleted and recreated from scratch,
with a new load balancer bound to a new IP.

By setting the traefik helm chart failure policy to `stop`, we prevent
the load balancer service from being recreated in case of chart
upgrade failure, at the cost of requiring manual user intervention to
restore automatic upgrades.

### DIFF
--- a/templates/traefik_config.yaml.tpl
+++ b/templates/traefik_config.yaml.tpl
@@ -4,7 +4,7 @@ metadata:
   name: traefik
   namespace: kube-system
 spec:
-  failurePolicy: stop
+  failurePolicy: abort
   valuesContent: |-
     service:
       enabled: true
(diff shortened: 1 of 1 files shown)


templates/traefik_config.yaml.tpl
```diff
@@ -4,6 +4,7 @@ metadata:
   name: traefik
   namespace: kube-system
 spec:
+  failurePolicy: stop
```
mysticaltech, line 7: @mnencia Why not set it to `ignore` instead, it seems safer, no?

mnencia: I think that abort is the right word https://github.com/k3s-io/helm-controller/blob/341b2eaa41e5cb9247c119bfa34a7ad131c7ac51/pkg/helm/controller.go#L67

PR #145, opened 2022-04-04
feat(agents): add location, label and taint options to agent-nodepool

HenkVanMaanen:


terraform.tfvars.example
```diff
@@ -92,6 +111,14 @@ load_balancer_type        = "lb11"
 #       "0.0.0.0/0"
 #     ]
 #   },
+#   {
```
phaer, line 114: What's the idea here? I think an example allowing outgoing connections to an external postgres cluster might actually be more useful than the existing one, but having both seems to be a bit redundant, no?

terraform.tfvars.example
```diff
@@ -14,7 +14,6 @@ private_key  = "/home/username/.ssh/id_ed25519"
 
 # These can be customized, or left with the default values
 # For Hetzner locations see https://docs.hetzner.com/general/others/data-centers-and-connection/
-# For Hetzner server types see https://www.hetzner.com/cloud
 location         = "fsn1"       # change to `ash` for us-east Ashburn, Virginia location
```
phaer, line 17: If I read this PR correctly, we should remove `location` here and in `variables.tf` as well, as it won't be used after this is merged? Alternatively we could use the global one as a default for node pools

phaer (approved):
Thanks @HenkVanMaanen, LGTM, useful feature!

terraform.tfvars.example
HenkVanMaanen: It's still used for the control plane and load balancer. I'm going to make another PR after this one so that the control plane can be configured the same way as the agents, with a pool, to be able to select multiple locations for the control plane nodes.

terraform.tfvars.example
HenkVanMaanen: The idea was to show the 'destination_ips' because otherwise you would have to dig into the module to find that option. Do you know a more useful example for outgoing connections?

mysticaltech:
Good PR, @HenkVanMaanen. Good to merge on my end!

terraform.tfvars.example
phaer: No - i'd just remove the example for incoming connections, as I believe people using `kube-hetzner` will have no trouble finding the docs on registry.terraform.io :)

But it's not really important. Good to merge from my point of few. Not entirely sure regarding the permissions in this repo: Are you able to merge yourself with the approvals?

HenkVanMaanen: Changed it to a SMTP port so the example is more clear.

HenkVanMaanen: I don't have permissions for 'Running workflow' and merging.

HenkVanMaanen:
Thanks for merging @phaer !

mysticaltech:
Thanks for this contribution, @HenkVanMaanen! It would be an honor to have you join our project Kube-Hetzner. I've just invited you here. 

And we also have a private chat room using Matrix (If you do not use it already, a good client for it is https://element.io/). Here's the invite link https://matrix.to/#/!flYKuAstLvMyDEZLPs:matrix.org?via=matrix.org :-)

HenkVanMaanen:
> Thanks for this contribution, @HenkVanMaanen! It would be an honor to have you join our project Kube-Hetzner. I've just invited you here.
> 
> And we also have a private chat room using Matrix (If you do not use it already, a good client for it is https://element.io/). Here's the invite link https://matrix.to/#/!flYKuAstLvMyDEZLPs:matrix.org?via=matrix.org :-)


I can't join via the provided url:
image

mysticaltech:
@HenkVanMaanen, please give me your matrix username, so I'll invite you directly.

mysticaltech:
Or try this one please https://matrix.to/#/!RowCRsakuwIKnDNrMe:matrix.org?via=matrix.org

HenkVanMaanen:
Does not work either. My username: @henkvanmaanen:matrix.org

phaer:
image
Haven't seen this one before :eyes: Quite interesting because I could send you a DM, but can't invite you

mysticaltech:
I get the same error. What client are you using @HenkVanMaanen? If it's a new account, did you very your email... etc.?
ksnip_20220405-125139

mysticaltech:
Do you get to this screen at least? 
ksnip_20220405-125439

mysticaltech:
Let's give it some time and retry a few hours from now. I believe some of the matrix servers may be overloaded or something.

mysticaltech:
@HenkVanMaanen I made the room private temporarily, can you try now https://matrix.to/#/!kube-hetzner:matrix.org

HenkVanMaanen:
Very weird, it says 'No known servers', I can join the public 'Matrix' room but can't join the kube-hetzner one, can't find it either in the public room search bar. Lets try again in a few hours indeed.

mysticaltech:
@HenkVanMaanen Am not yet able to send you an invite, but please try that link once more https://matrix.to/#/#kube-hetzner:matrix.org

PR #147, opened 2022-04-06
WIP: control plane pool

HenkVanMaanen:
Can somebody help me? I can't figure out how to fix this problem: control_plane.tf:19

image

### DIFF
--- a/control_planes.tf
+++ b/control_planes.tf
@@ -1,7 +1,8 @@
 module "control_planes" {
   source = "./modules/host"
 
-  count                  = var.control_plane_count
+  for_each = local.control_plane_nodepools
+
   name                   = "${var.use_cluster_name_in_node_name ? "${var.cluster_name}-" : ""}control-plane"
   ssh_keys               = [hcloud_ssh_key.k3s.id]
   public_key             = var.public_key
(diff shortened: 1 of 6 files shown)


control_planes.tf
```diff
@@ -28,33 +29,33 @@ module "control_planes" {
...
+    host           = module.control_planes[each.key].ipv4_address
   }
 
   # Generating k3s server config file
   provisioner "file" {
     content = yamlencode({
-      node-name                = module.control_planes[count.index].name
-      server                   = "https://${element(module.control_planes.*.private_ipv4_address, count.index > 0 ? 0 : 1)}:6443"
+      node-name                = module.control_planes[each.key].name
+      server                   = "https://${local.first_control_plane.private_ipv4_address}:6443"
```
mysticaltech, line 49: This line seems necessary to retain more or less the same logic as before. @mnencia That's your implementation I believe. How do you think we could solve this?

init.tf
```diff
@@ -97,7 +97,7 @@ resource "null_resource" "kustomization" {
         name                       = "${var.cluster_name}-traefik"
         load_balancer_disable_ipv6 = var.load_balancer_disable_ipv6
         load_balancer_type         = var.load_balancer_type
-        location                   = var.location
+        location                   = var.load_balancer_location
```
mysticaltech, line 100: Great

mysticaltech:
@HenkVanMaanen This is the line that does the automatic subnet generation, and in there, we can take into account the control_plane nodepools too. 

https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/4e3f34585e89813d2c268b9fb68f03c7ae9e5a5d/locals.tf#L191

The first subnet is reserved for anything; the second was for the control plane nodes, but now, I suggest we keep that only for the first control plane node, and before the agent nodepools subnets, we generate ones for the control planes nodepools.

The `hcloud network describe k3s` command is very useful for debugging subnets.
ksnip_20220407-134304

mysticaltech:
Supplanted by https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/149.

PR #149, opened 2022-04-09
Control plane pool

mysticaltech:
@HenkVanMaanen, I have recreated the PR with a branch in the team account so that we can all push to it. You can push here directly now.


mysticaltech:
About the progress - for the time being, only works with one control plane nodepool. Am investigating why it still produces the ip_not_available error for more than one control plane nodepools.

mysticaltech:
The IP problem normally is solved, however, there is still an issue with the kustomization with non-first-control-plane nodes.

mysticaltech:
For the IPs, here's how the subnets work:
IMG_20220409_092154965

mysticaltech:
When I SSH into the first control plane (see README, if you haven't done it already), use `systemctl status k3s`, it says that k3s is `activating`:
ksnip_20220409-101429

mysticaltech:
And when I inspect the k3s logs with `journalctl -u k3s`, I see that one of the locations is in state FullDisruption:

ksnip_20220409-101231

mysticaltech:
The config used is:

```
control_plane_nodepools = [
  {
    server_type = "cpx11",
    location    = "fsn1",
    labels      = [],
    taints      = [],
    count       = 2
  },
  {
    server_type = "cpx11",
    location    = "nbg1",
    labels      = [],
    taints      = [],
    count       = 1
  }
]
```

control_planes.tf
```diff
@@ -28,33 +29,33 @@ module "control_planes" {
...
+    host           = module.control_planes[each.key].ipv4_address
   }
 
   # Generating k3s server config file
   provisioner "file" {
     content = yamlencode({
-      node-name                = module.control_planes[count.index].name
-      server                   = "https://${element(module.control_planes.*.private_ipv4_address, count.index > 0 ? 0 : 1)}:6443"
+      node-name                = module.control_planes[each.key].name
+      server                   = "https://${module.control_planes[each.key].private_ipv4_address == "10.1.0.101" ? "10.1.0.102" : "10.1.0.101"}:6443"
```
mysticaltech, line 49: @mnencia Can you confirm this logic is correct, please? As you wrote the initial working code.

mysticaltech:
When I use the same location, k3s on the first control plane run correctly. However, there is still an issue with our control planes. Will investigate further later on. Maybe you can unlock this @HenkVanMaanen :)

control_planes.tf
mnencia: It looks correct. I'll test it this afternoon.

mysticaltech: Awesome! It does not yet work on my end, and I do not know where the ball gets dropped.

HenkVanMaanen:
I think I found the problem, going to test it and then push to this branch. It's a bug introduced by me in the previous branch

Issue #142, opened 2022-04-03
Placement group 10 servers limit


HenkVanMaanen:
First of all, thank you for this amazing project. I created my own terraform module to deploy k3s on Hetzner. But this project is way more organized and really nice! 

If I recall correctly the placement group is limited by 10 servers per group. So [this code adding placement groups](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/34) will only work until there are more than 10 servers.

mysticaltech:
Good find @HenkVanMaanen! It seems to be the case indeed. I have verified the info here https://docs.hetzner.com/cloud/placement-groups/overview.

@phaer @mnencia, I propose we make the placement group optional then. @HenkVanMaanen, please do not hesitate to submit a PR; using a variable flag for this would be ideal.

mnencia:
I would use, by default, one placement group per node group plus one for the control plane. However, having a flag to disable them looks like a good idea if you hit the limits.

HenkVanMaanen:
The solution I used was creating enough placement groups so that all servers would fit into these groups, with this algorithm the servers will get round robinned into the placement groups e.g.: 

```terraform
locals {
  hcloud_placement_group_limit = 10
}

resource "hcloud_placement_group" "workers" {
  count = ceil(var.worker_replicas / local.hcloud_placement_group_limit)
  name = "workers-group-${count.index + 1}"
  type = "spread"
}

resource "hcloud_server" "workers" {
  count       = var.worker_replicas
  name        = "${var.cluster_name}-worker-${count.index + 1}"
  placement_group_id = element(hcloud_placement_group.workers.*.id, ceil(count.index / local.hcloud_placement_group_limit))
}
```

mnencia:
This is an interesting way to support more than ten servers in the same group. With this trick and having different placement groups per node pool, we could avoid having the switch to disable placement groups globally.

HenkVanMaanen:
We still need to consider the 50 placement group limit of Hetzner.

![image](https://user-images.githubusercontent.com/36051232/161519027-7a4ee212-b0cf-44df-a9b0-59e428fbd6bf.png)

mnencia:
Well, if you have 500 servers, you probably have the resources to contribute to the project a global disable placement group patch :-)

mysticaltech:
Nice solution, @HenkVanMaanen!

gor181:
[...]

PR #151, opened 2022-04-13
Placement group fix

mysticaltech:
Folks, this fixes the placement group limitation and uses the method proposed by @HenkVanMaanen in #142.


mysticaltech:
This is important, merging fast!

PR #152, opened 2022-04-13
Network subnet dissociation

mysticaltech:
This creates two groups of subnets, one for the agent and one for the nodepools; that way, the nodepools are a lot easier to scale.

Combined max 255 nodepools, with each 64k nodes. Should be enough to make everyone happy! :) 

Pardon my ugly drawing, but should make it simpler to understand:

IMG_20220413_132450932

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,17 +97,13 @@ _Once you start with Terraform, it's best not to change the state manually in He
 
 ### Scaling Nodes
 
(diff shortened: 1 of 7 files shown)


mysticaltech:
@phaer @mnencia @HenkVanMaanen FYI above, how the subnets now work!

mysticaltech:
Merging without further do, as in an ideal scenario it should have been part of yesterday's PR. We have to keep the master always fresh and optimal.

HenkVanMaanen:
Nice work!

Issue #160, opened 2022-04-24
[REQ] Ability to install Longhorn to use local storage for volumes


karbowiak:
Hello there,

I just tried installing Longhorn onto the cluster, but during driver-deployment it hangs itself and waits forever - which if i remember correctly is because open-iscsi is missing on the system (OR it can't find a path, but that usually comes with a different error 😅)

Is there a way to install open-iscsi onto the nodes so Longhorn can run, or is there something bigger preventing Longhorn from working.

Would be lovely to have some volumes locally and others in hcloud, both money and performance wise 😄

mysticaltech:
@karbowiak Try installing open-iscsi at this level: 

https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/5087c15bf20754d73ee0c42a8d3e5b06721baac7/modules/host/main.tf#L71-L73

Do it by adding the following command to the command array there: 

`transactional-update pkg install open-iscsi
`
And if it works for you, please submit a PR.

mysticaltech:
Hey @karbowiak, do you need help with this? Ideally, please do try it, and if you can make Longhorn work with this, I will commit it to the repo and release a new minor version of the module.

karbowiak:
@mysticaltech hey! thanks for the example - i'll totally give it a try this weekend ! :D

And yes, if it all works out - i'll see if i can't make it nice with a kustomize and some enable/disable flags and make a pull request :)

mysticaltech:
Sounds like a plan!!

skriems:
I'm currently looking into this as well. What I've got so far is

```
# Install open-iscsi
  provisioner "remote-exec" {
    inline = [
      "set -ex",
      "transactional-update pkg install -y open-iscsi",
      "systemctl enable iscsid",
      "systemctl start iscsid",
    ]
  }
```
This worked manually on one node so I think this is fine and doesn't need a restart or anything. Will continue on the others, deploy longhorn and report back how it goes

mysticaltech:
Good to hear; yes, the way MicroOS works, it needs to restart after any installation of packages to commit the new snapshot. It's a transactional system.

Best to do the installation at https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/5087c15bf20754d73ee0c42a8d3e5b06721baac7/modules/host/main.tf#L71-L73, as stated above.

For the reboot, it's already being done after the above point, so that is already taken care of. 

[...]

PR #170, opened 2022-04-29
#160 Install open-iscsi and Longhorn

karbowiak:
ref: #160

This installs open-iscsi and Longhorn using Helmfile.

Not sure this is the optimal way to do it - but it does work 😅

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,12 +54,13 @@ Follow those simple steps, and your world's cheapest Kube cluster will be up and
 
 First and foremost, you need to have a Hetzner Cloud account. You can sign up for free [here](https://hetzner.com/cloud/).
 
-Then you'll need to have [terraform](https://learn.hashicorp.com/tutorials/terraform/install-cli),  [kubectl](https://kubernetes.io/docs/tasks/tools/) cli, and [hcloud](<https://github.com/hetznercloud/cli>) the Hetzner cli. The easiest way is to use the [homebrew](https://brew.sh/) package manager to install them (available on Linux, Mac, and Windows Linux Subsystem).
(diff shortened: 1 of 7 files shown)


skriems:
cool :) I think I get the idea that you use `transactional-update` for starting the deamon, as this also create a snapshopt, right? I wonder if that's necessary 🤔 

With `defaultClass: true` you're setting longhorn as the default StorageClass, right? I wonder if its possible if longhorn can be combined with the hcloud-storage class. Haven't had time to look into that. On the other hand I assume that with longhorn you _want_ to have the volumes on the nodes and an additional backup strategy should push data to hetzner volumes, maybe... 🤔

karbowiak:
@skriems 

Afaik just doing `systemctl enable --now iscsid` will just enable it for now, after the next reboot it'll be gone - hence the transactional update :)

as for `defaultClass: true` it's just what i use myself, i noticed after i made the pull request that hcloud is also setting itself as default 😅 

So by default it'll most likely end up using hcloud since it comes first in the alphabet.
As for backing up Longhorn, that has to happen to an S3 bucket - that's how it's own internal backup solution works.

As for single node clusters, it wouldn't fail - it'd just still try and keep 2 volumes around 😂 
As for setting it to the X amount of worker nodes added, that could work - but you'd end up with very little usable space, longhorn should be smart enough to move a volume over to the node that requires it if it doesn't have it there already

mysticaltech:
@karbowiak Thanks! There are many good things in this PR; I like how you think. However, a few things. 

It needs to be fully optional, with a variable flag to determine if longhorn is used or not.

Also, the whole interaction with hcloud needs to be clarified, and the different major use cases need to be exposed with a flag. 

One of those use cases would be either hcloud or longhorn (mutual exclusion), in which case the Hetzner CSI needs to never get installed (so its installation must be if longhorn is false, for instance).

Last but not least, please retarget this PR to the "staging" branch, so we can test it out. We try to keep the master in sync with the latest release. So closing for now while you re-submit it to the "staging" branch, and from there, we can make the improvements needed before we merge.

mysticaltech:
As for the systemctl enable --now, am almost sure it does not need transactional updates as @skriems mentioned, please double check!

skriems:
@karbowiak I checked the docs and in fact an `enable --now` does both, _enable_ (start the deamon after reboot) and _start_ a given deamon. So I think since nothing is really installed we can omit the transactional-update :)

I you still working on this?

PR #171, opened 2022-04-29
Add private CIDR for calico

### LINKED ISSUE #169: Calico subnet performance
Is it a problem that Calico uses 192.168.0.0 subnet instead of the 10.0.0.0 subnet? For example is it faster to go through the private network or are requests always routed over the private hetznet network.

nupplaphil:
Fixes #169


mysticaltech:
Thanks @nupplaphil !

PR #172, opened 2022-04-30
completed k3s channel list

### LINKED ISSUE #165: [REQ] add more (all) k3s channels for kubernetes version
In my cluster, we are installing components that are not currently running on kubernetes >v1.21. however, the stable channel is already on v1.22.x. K3s itself offers additional channels like v1.21. But these are not allowed in the channel variable. 

I could provide a PR that adds the missing channels to the variable. What do you think?

batthebee:
Fixes #165


mysticaltech:
Well done, @batthebee!

PR #173, opened 2022-05-02
#160 Longhorn

karbowiak:
This merge commit refers to #160 

It brings with it..

- Ability to define the version for Kured
- Ability to enable/disable installation of Hetzner CSI
- Ability to enable/disable installation of Longhorn
- Sort so Longhorn is not default when Hetzner CSI is also enabled

Tested:
- terraform plan
- terraform apply -auto-approve
- terraform destroy -auto-approve
- installation with/without hetzner csi while longhorn was enabled
- installation with/without longhorn while hetzner csi was enabled
- installation with neither enabled


helm/longhorn.yaml
```diff
@@ -0,0 +1,22 @@
+---
```
nupplaphil, line 1: Wouldn't it be better to stick to one type of helm usage, would it? Currently we're using the rancher https://rancher.com/docs/k3s/latest/en/helm/ for helm, f.e. for the traefik config. This would reduce the amount of dependencies as well

nupplaphil (changes requested):

helm/longhorn.yaml
karbowiak: You can define HelmCharts as a manifest for K3? Interesting - definitely would simplify things - i'll have a whack at it :)

nupplaphil: Exactly :) the nice thing is that so you won't have to install helm locally and the helm is managed by k8s itself

modules/host/main.tf
```diff
@@ -69,7 +69,7 @@ resource "hcloud_server" "server" {
   provisioner "remote-exec" {
     inline = [
       "set -ex",
-      "transactional-update shell <<< 'rpm --import https://rpm.rancher.io/public.key;zypper install -y https://github.com/k3s-io/k3s-selinux/releases/download/v0.5.stable.1/k3s-selinux-0.5-1.sle.noarch.rpm'"
+      "transactional-update shell <<< 'rpm --import https://rpm.rancher.io/public.key; zypper install -y open-iscsi https://github.com/k3s-io/k3s-selinux/releases/download/v0.5.stable.1/k3s-selinux-0.5-1.sle.noarch.rpm'"
```
nupplaphil, line 72: If we don't use longhorn, we should avoid installing open-iscsi to minimize the image

nupplaphil (changes requested):

helm/longhorn.yaml
mysticaltech: Very interesting, @nupplaphil; I had not connected these dots myself; indeed, managing through the install through `HelmChartConfig` would be ideal!

modules/host/main.tf
mysticaltech: Thought about that, but was thinking that maybe other storage solutions do use that too, so maybe it does not hurt leaving it. If not started, it won't really add much overhead, either in space or time of installation. But please correct me if I'm wrong.

karbowiak: Not sure other storage solutions need it, but they might.
I agree on not installing it if longhorn ain't needed - but not installing it and then wanting to add longhorn later might complicate things.

Ideally i'm thinking something like an array containing all extra packages to install that can be set in terraform.tfvars maybe.

Not sure how to go about that one tho 😅

mysticaltech: Could very well be done IMHO; that would be ideal. And when Longhorn is selected, the required package is installed and started.

terraform.tfvars.example
```diff
@@ -103,10 +103,18 @@ load_balancer_location = "fsn1"
 
 ### The following values are entirely optional
 
+# To use local storage on the nodes, you can enable Longhorn, default is "false"
```
mysticaltech, line 106: Awesome!!

variables.tf
```diff
@@ -65,6 +65,12 @@ variable "hetzner_csi_version" {
   description = "Version of Container Storage Interface driver for Hetzner Cloud"
 }
 
+variable "kured_version" {
```
mysticaltech, line 68: Useful addition! Thanks.

karbowiak: Honestly needed it, because i kept running into issues where Github would throttle how often i could run terraform 😂

helm/longhorn.yaml
karbowiak: Should be added now, lemme know if it looks alright :)

Tested it and.. it works on my machine 😂

variables.tf
mysticaltech: Interesting... Did not know that did that! You must be a power user! :)

karbowiak: More like "run the thing an abnormal amount of times because you f'ed up writing it correctly the first time" 😂

nupplaphil: I had similar issues until I added a Github token To the Github provider inside my parent main module :)

karbowiak: @nupplaphil and how does one do that ? 😅 

Like the parent main module part, not the token

mysticaltech:
All is now very elegant with the HelmChart, and opens the possibilities to add even more add-ons that way in the future!

Now we just need to install and enable the required packages if longhorn is enabled and we can merge! Later on we can sort out the optional packages feature.

karbowiak:
Alrighty @mysticaltech - i'll get that sorted a bit later today :)

(For some reason, work thinks i should work - what for a ludicrus assumption is that 😂)

skriems:
awesome work so far !! :) Can't wait integrating it on my end. I'm just about to spin up our new setup... gonna steal your work locally :P

karbowiak:
Ok i have to tap out - i got no clue how i'll get it into the main.tf in modules/host that longhorn is enabled/disabled.

Tried passing it in via agents.tf and control_panels.tf like as follows:

```
  location               = each.value.location
  server_type            = each.value.server_type
  ipv4_subnet_id         = hcloud_network_subnet.control_plane[[for i, v in var.control_plane_nodepools : i if v.name == each.value.nodepool_name][0]].id
  enable_longhorn = var.enable_longhorn
  ```
  
  no dice, then like labels is done, just calling it extras - no dice.
  I can get it in via labels, but that's not really nice.
  
  So how in the world do i pass an arbitrary variable into modules/host/main.tf from agents.tf and control_planes.tf? 😂

mysticaltech:
That's exactly how you do it: `enable_longhorn = var.enable_longhorn` and in the variables of the module, you also have to declare enable_longhorn.

Then it should work, same as other variables that are passed to the server module!

mysticaltech:
Ok, folks, now we have a `packages_to_install` variable we can use to install stuff, I added the logic for `open-iscsi` when longhorn is enabled.

This is now merged! Thanks for your good work @karbowiak! 🥳

mysticaltech:
This is also now part of the new v1.1.1 release!

PR #176, opened 2022-05-04
Rancher addon

mysticaltech:
This adds both cert-manager and rancher!

ksnip_20220504-214126

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,6 +229,29 @@ To use Kube-Hetzner on Terraform cloud, use as a Terraform module as mentioned a
 
 </details>
 
+<details>
+
+<summary>Configure add-ons with HelmChartConfig</summary>
+
+For instance, to customize the Rancher install, if you choose to enable it, you can create and apply the following `HelmChartConfig`:
+
+```yaml
+apiVersion: helm.cattle.io/v1
+kind: HelmChartConfig
+metadata:
+  name: rancher
+  namespace: kube-system
+spec:
+  valuesContent: |-
+    **values.yaml content you want to customize**
+```
+
+The helm options for Rancher can be seen here <https://github.com/rancher/rancher/blob/release/v2.6/chart/values.yaml>.
+
+Same goes for all add-ons, like Longhorn, Cert-manager, and Traefik.
+
+</details>
+
 ## Debugging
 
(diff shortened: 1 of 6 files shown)


strowi:
Nice work, can't test it atm though.
Just a small thing: Will `number_control_plane_nodes` somehow override the `allow_scheduling_on_control_plane`, or is it too early and i need more coffee first? If so it might me worth mentioning. ;)

mysticaltech:
@strowi Thanks for diving deep into the code! `number_control_plane_nodes` is just a count, and `allow_scheduling_on_control_plane` is a boolean turned on automatically in the case of single-node clusters to allow for scheduling of payloads on the control plane(s). So both are not related! So either it was too early, you needed coffee, or I'm missing your point! 😂

strowi:
> `allow_scheduling_on_control_plane`  is a boolean turned on automatically in the case of single-node clusters 
That's what i meant to ask or what confused me- ;) 
`number_control_plane_nodes` has nothing to do with the `control_plane_nodepools`-count.
I thought about the recommended 3-node Rancher-HA-Cluster (control-plane on all 3 nodes) as per the [rancher-docs](https://rancher.com/docs/rancher/v2.5/en/installation/resources/k8s-tutorials/how-ha-works/#recommended-architecture), which have the rancher-workload running on control-nodes.

mysticaltech:
Ah yes, I understand. Then all good huh, no change needed? I thought it would make sense to have Rancher on CP nodes.

strowi:
I would just leave it as is.
We never deployed the helm-chart on a single node, since it is equivalent to `docker run rancher...`.

For HA you want 3 control nodes for the rancher-cluster itself (because of etcd + quorum) but don't really have any load on it. So it makes sense to run the rancher-cluster on the same cp nodes. I don't really see a point in adding extra-workers for the rancher-ui.

mysticaltech:
Makes sense! Thanks for clarifying 🙏

strowi:
@mysticaltech regarding the rancher-addon - i "had" to implement a code-only approach to importing the downstream-clusters into the rancher-ui last couple of days. 
It works by using the offical rancher/rancher2 provider with access + secret-key of a user and doesn't need logging into the rancher-ui (after getting the credentials of course). 
It has also the advantage, that it would be possible to install additional rancher-apps (like monitoring etc..) via terraform in one go.  If this seems like a good idea i can open a MR.

mysticaltech:
Definitely, please do! 🙏✨

PR #182, opened 2022-05-09
system-upgrade on the storage node

### LINKED ISSUE #181: BotKube reports errors for system-upgrade job of storage node
Hey,

I've added BotKube to my cluster and it constantly reports errors regarding the update Job of the storage node:

```
v1/pods error
Error Occurred in Pod: system-upgrade/apply-k3s-agent-on-k3s-storage-rrz-with-1b39df95bca3c0-5vwh7 in k3s cluster

0/7 nodes are available: 1 node(s) had taint {server-usage: storage}, that the pod didn't tolerate, 3 node(s) didn't match Pod's node affinity/selector, 3 node(s) had taint {[node-role.kubernetes.io/master](http://node-role.kubernetes.io/master): }, that the pod didn't tolerate.
```

I wonder if this is a real error. I could prevent this by disabling upgrades of the storage node, or disabling botkube to report this error on the _system-upgrade_ namespace.

Another question came to mind now that we have longhorn: If enabled, wouldn't it make more sense to have more storage on each node, instead of a big storage node?

skriems:
Adds `tolerations` so that the system-upgrade job can run on the storage node.

fixes #181

### DIFF
--- a/templates/plans.yaml.tpl
+++ b/templates/plans.yaml.tpl
@@ -16,6 +16,11 @@ spec:
       - {key: k3s_upgrade, operator: NotIn, values: ["disabled", "false"]}
       - {key: node-role.kubernetes.io/master, operator: NotIn, values: ["true"]}
   serviceAccountName: system-upgrade
+  tolerations:
+  - effect: NoSchedule
+    key: server-usage
+    operator: Equal
+    value: storage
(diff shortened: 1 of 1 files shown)


mysticaltech (commented):
Perfect, thanks!

Issue #190, opened 2022-05-18
LB not removed when running as module


strowi:
Hi,

i just ran into an issue where the LB is not removed.
This happened when running as a terraform-module and it seems this is related to the `destroy_traefik_loadbalancer` using `${module.path}/..` for the kubeconfig?

when removing the `on_failure = continue` it results in: 
```
│ Error: local-exec provisioner error
│ 
│   with module.rvg.null_resource.destroy_traefik_loadbalancer,
│   on modules/kube-hetzner/main.tf line 127, in resource "null_resource" "destroy_traefik_loadbalancer":
│  127:   provisioner "local-exec" {
│ 
│ Error running command 'kubectl -n kube-system delete service traefik --kubeconfig modules/kube-hetzner/kubeconfig.yaml
│ ': exit status 1. Output: error: stat modules/kube-hetzner/kubeconfig.yaml: no such file or directory

```

regards,
strowi

mysticaltech:
Good find @strowi, good find, don't hesitate to submit a PR.

Otherwise, will fix ASAP. If I remember correctly there are other places in the code where this problem was fixed.

mysticaltech:
Fixed in #191, and fix included in v1.1.9.

PR #191, opened 2022-05-18
fix lb not being destroy if used as module

strowi:
solves #190 

using kubeconfig.yaml from project-root folder works here.;)

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -76,7 +76,7 @@ resource "null_resource" "destroy_traefik_loadbalancer" {
   provisioner "local-exec" {
     when       = destroy
     command    = <<-EOT
-      kubectl -n kube-system delete service traefik --kubeconfig ${path.module}/kubeconfig.yaml
+      kubectl -n kube-system delete service traefik --kubeconfig kubeconfig.yaml
     EOT
     on_failure = continue
   }


mysticaltech:
Thanks, @strowi! :)

I was waiting for your first PR to ask you if you wanted to join the team? Basically, you'd be able to get a bit more involved; you will be able to create branches, merge PRs, close issues, and get access to our private Matrix channel.

Not a lot is going on, but I love your ideas and know-how! It would be an honor to add you to our team; what do you say?

strowi:
@mysticaltech thx for the invite, happy to help ;) Although i am not sure how much time i can spend besides work and life.
I just noticed, the MR will break if the module is being used multiple times - it will overwrite the same kubeconfig. It probably should contain the culster-name.

mysticaltech:
Welcome on board @strowi! 🤗 Don't worry; we're all participating on a best-effort basis. Like the spot instances on AWS! 😂

You can join our matrix room here https://matrix.to/#/#kube-hetzner:matrix.org (I will delete this discussion after). Personally, I like the Element-io client, but anything works!

About the MR, I understand something breaks if the module is reused in the same directory, but what do you mean by MR? I agree adding the cluster name, to the kubeconfig filename would be great, like kubeconfig-${cluster-name}.yaml, maybe?!

strowi:
Haha, seems just like my recently updated Xubuntu22.04. Thank you for the warm welcome :) 
How do i join? I am getting a not-invited message from element (using my github-account)?

Yeah i was talking about "MR breaking" because the change also might break existing functionality.
but i just noticed other problems with the module-usage.. It seems the vars inside `module "cluster" { ...}` are not being used. ;/

strowi:
@strowi:matrix.org

mysticaltech:
Just sent you the invite!

mysticaltech:
@strowi Looking forward to the minor tweak to support multiple deploys in the same folder! Adding the cluster name to the kubeconfig filename, for instance.

PR #194, opened 2022-05-19
Support for terraform 1.2.0

### LINKED ISSUE #193: Terraform 1.2.0 compatibility
With 1.2.0 i am getting the following when using it as a module.  Not sure yet how the upgrade-path should look like. Have to investigate..

```
The Terraform configuration must be valid before initialization so that
Terraform can determine which modules and providers need to be installed.
╷
│ Error: Module is incompatible with count, for_each, and depends_on
│ 
│   on ../modules/kube-hetzner/control_planes.tf line 4, in module "control_planes":
│    4:   for_each = local.control_plane_nodes
│ 
│ The module at module.rancher is a legacy module which contains its own local provider configurations, and so calls to it may not use the count, for_each, or depends_on arguments.
│ 
```

mysticaltech:
This adds support for the latest version of Terraform by moving the provider params only at the root of the first module call. It fixes #193.


mysticaltech:
Folks, we need to support the latest terraform version fast. Am merging without further do; we can fix it later if needed.

PR #196, opened 2022-05-20
+ use cluster_name as prefix for local kubeconfig.yaml

strowi:
Oh man, this took wayy too long to figure out - can only use `self` an `each.key` on destroy AND you need to specify those als triggers AND environment.
But it seems to be working now. Ofc if there is a better solution... ;)

relates to #190

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -81,7 +81,7 @@ resource "null_resource" "destroy_traefik_loadbalancer" {
 
     environment = {
       kubeconfig_prefix = self.triggers.kubeconfig_prefix
-     }
+    }
   }
 
   depends_on = [
--- a/main.tf
+++ b/main.tf
@@ -70,7 +70,8 @@ resource "null_resource" "destroy_traefik_loadbalancer" {
 
   # this only gets triggered before total destruction of the cluster, but when the necessary elements to run the commands are still available
   triggers = {
-    kustomization_id = null_resource.kustomization.id
+    kustomization_id  = null_resource.kustomization.id
+    kubeconfig_prefix = var.cluster_name
   }
 
   # Important when issuing terraform destroy, otherwise the LB will not let the network get deleted
--- a/main.tf
+++ b/main.tf
@@ -70,14 +70,14 @@ resource "null_resource" "destroy_traefik_loadbalancer" {
 
   # this only gets triggered before total destruction of the cluster, but when the necessary elements to run the commands are still available
   triggers = {
-    kustomization_id  = null_resource.kustomization.id
-    kubeconfig_prefix = var.cluster_name
+    kustomization_id = null_resource.kustomization.id
+    cluster_name     = var.cluster_name
   }
 
   # Important when issuing terraform destroy, otherwise the LB will not let the network get deleted
   provisioner "local-exec" {
     when       = destroy
-    command    = "kubectl -n kube-system delete service traefik --kubeconfig ${self.triggers.kubeconfig_prefix}_kubeconfig.yaml"
+    command    = "kubectl -n kube-system delete service traefik --kubeconfig ${self.triggers.cluster_name}_kubeconfig.yaml"
     on_failure = continue
   }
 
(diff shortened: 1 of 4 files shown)


main.tf
```diff
@@ -75,10 +76,12 @@ resource "null_resource" "destroy_traefik_loadbalancer" {
   # Important when issuing terraform destroy, otherwise the LB will not let the network get deleted
   provisioner "local-exec" {
     when       = destroy
-    command    = <<-EOT
-      kubectl -n kube-system delete service traefik --kubeconfig kubeconfig.yaml
-    EOT
+    command    = "kubectl -n kube-system delete service traefik --kubeconfig ${self.triggers.cluster_name}_kubeconfig.yaml"
```
mysticaltech, line 79: Hey @strowi, well-done man!

I understand this is necessary, and I remember coding the destroy, but it was such a pain in the behind.

Just one thing, where is the triggers defined? Why is it through the environment? And why aren't we using `kubeconfig_prefix` anyways? We just define the environment, but why?

So we `self.triggers.cluster_name` and `self.triggers.kubeconfig_prefix`, both are different right, no the same?

Will try to read more about environments and triggers! :) But great work!

strowi:
Yeah the triggers is hidden, because i messed up the rebase. Good catch! ;)
Also the environment is not needed it seems. I had a dirty test-env, and those things don't change once created which messed everything up for me.
But after some more testing (and github-timeouts) it seems to work as expected. Sorry for the inconvenience.

main.tf
strowi: Yeah the triggers is hidden, because i messed up the rebase. Good catch! ;)
Also the environment is not needed it seems. I had a dirty test-env, and those things don't change once created which messed everything up for me.
But after some more testing (and github-timeouts) it seems to work as expected. Sorry for the inconvenience.

mysticaltech: Ok, perfect! Looking forward to the final version 🚀

mysticaltech: So just to clarify,var.cluster_name is not available during the destroy? Correct?

strowi: Yes, using `${var.cluster_name}` would result in:

```bash
│ Error: Invalid reference from destroy provisioner
│ 
│   on ../terraform-hcloud-kube-hetzner/main.tf line 80, in resource "null_resource" "destroy_traefik_loadbalancer":
│   80:     command    = "kubectl -n kube-system delete service traefik --kubeconfig ${var.cluster_name}_kubeconfig.yaml"
│ 
│ Destroy-time provisioners and their connection configurations may only reference attributes of the related resource, via 'self', 'count.index', or 'each.key'.
│ 
│ References to other resources during the destroy phase can cause dependency cycles and interact poorly with create_before_destroy.
```

mysticaltech: Ah yeah, I remember banging my head at the time! I did not want to do a local exec at the beginning, but had no choice!

mysticaltech (commented):
Still I do not see a declared trigger of the name cluster_name, how does it work? Only one I see declared is for the kustomization_id. Does not make sense to me yet. Please explain if you can 🙏

strowi:
I am so sorry, i work with git everyday but somehow messed up. :worried: 
I triple checked now, and the trigger should hopefully be there.

mysticaltech:
No problem, @strowi, it happens to all of us!

One more thing, can we rename `kubeconfig_prefix` to `cluster_name`? Just to keep it consistent.

And, why is `kustomization_id` in the triggers still needed?

mysticaltech:
Thanks, @strowi! This was a giant bug that would have cost folks a lot of money with all these forgotten LBs that would not have been deleted properly! Also, well done finding the way to pass info via triggers!  🥳 🙏

strowi:
No Problem. I didn't touch anything else like the kustomize trigger, since i still need to have a deeper look into that.

mysticaltech:
Yes, exactly, it was me who wrote that, so that it triggers only when the kustomization null resource gets deleted (very early), so that we can still run kubectl successfully.

PR #198, opened 2022-05-22
import downstream cluster via access + secret token

strowi:
This replaces #176 and allow the programmatical import of clusters by the use of access + secret token.
Furthermore it will allow to use the rancher2-provider to install more rancher-addons from their catalog (e.g. monitoring, logging-charts..).

(By using hdns for the racher dns-records + a local-exec  "while .. nslookup rancher-ui" it would even be possible to do both with 1 terraform-plan).

On destroy this will obviously fail  if the rancher-ui server is non-responsive/has already been deleted. But this usually shouldn't happen. 

@mysticaltech as already mentioned, please check. Especially regarding the provider-setup - i'm not sure i have the new layout already internalized. Hopre this one goes smoother than the previous one.;)

### DIFF
--- a/init.tf
+++ b/init.tf
@@ -67,6 +67,15 @@ resource "null_resource" "first_control_plane" {
   ]
 }
 
+# create the cluster in rancher
+resource "rancher2_cluster" "downstream_cluster" {
+  count = var.enable_rancher_import ? 1 : 0
+
+  provider    = rancher2
+  name        = var.cluster_name
+  description = "imported via terraform"
+}
+
 resource "null_resource" "kustomization" {
   connection {
     user           = "root"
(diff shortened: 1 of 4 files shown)


mysticaltech:
Thanks @strowi , I appreciate you taking the time to submit this, but I do not understand why it is needed. Everything is already working great with the present setup without the need of a new provider!

Can you explain to me exactly what is the added value? I want to understand.

mysticaltech:
The ability to programmatically install rancher addons is interesting, but it's not a scope I think we should support. Folks csn do that from the rancher ui already.

mysticaltech:
I'm sorry @strowi , I will close that PR for now, but it you can help me understand why we really need it, we can reopen it later. Thanks for your understanding!

Issue #167, opened 2022-04-26
[REQ] optionally provide hetzner_ssh_key


batthebee:
If I want to create multiple clusters I currently need several different ssh keys, because hcloud does not allow to store multiple keys with the same fingerprint. 

I could create a PR that optionally passes the id of the key used and then does not store a new key in hcloud.

What do you think?

mysticaltech:
Sounds good; the id could help differentiate the keys.

Also, maybe autogenerating the keys with https://registry.terraform.io/providers/hashicorp/tls/latest/docs/resources/private_key would be an option here, and store them in an array.

For instance, we could do that if both the `public_key` and `private_key` variables are left blank.

Please submit a PR; that's a completely valid use case.

mysticaltech:
That also re-joins your proposal of having the keys as strings. This would indeed be doable; see my answer here https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/161#issuecomment-1109531226.

mysticaltech:
@batthebee In your coming PR(s), please target the `staging` branch 🙏

batthebee:
@mysticaltech you mentioned on #161 for local use, it's easy in bash to makes strings look as files!

How does it Work?

mysticaltech:
Yep, basically you use cat and the string enclosed in EOF. See https://unix.stackexchange.com/questions/505828/how-to-pass-a-string-to-a-command-that-expects-a-file

strowi:
This would be greatly appreciated, since i am currently testing everything in a single project...

mysticaltech:
Done in #204, thanks @batthebee! 👏

Issue #161, opened 2022-04-25
[REQ] set ssh key by string and not by file path


batthebee:
I want to use your environment to set up multiple clusters. The ssh keys should be created dynamically with `tls_private_key`, each host a different one. 

Now I would have to generate the key and write it to a file, which I pass to the module, which generates noise on each ci run and is overall quite unattractive.

Do you have a better suggestion? Or should I adapt the module so that it also accepts ssh keys as string?

batthebee:
I took a closer look and saw that you also need to access the cluster via ssh. So the local ssh key is needed. I close the issue again.

mysticaltech:
Having access to keys as string would be a sweet option to have. 

It needs to be ed25519 and passphrase-less in order to work without SSH agents. 

And for local use, it's easy in bash to makes strings look as files!


PR welcome and appreciated.

mysticaltech:
Done in #204 🎯

PR #204, opened 2022-05-25
changed ssh keys from file to string.

batthebee:
solves #167 #161

was a bit difficult to solve. It's not 100% clean either, but it does what it's supposed to. 

I tried both the way to create the ssh key via pipe & /dev/stdio and via a file created via terraform local_secret_file, which I then delete again. 

The first does not work because the rights must be set to 0600. 
The second is unattractive because changes are displayed on every rerun. 

I have now solved this so that the corresponding file is created under /tmp and then deleted again. The file is created with 0600, which I think should be acceptable from a security point of view.

### DIFF
--- a/agents.tf
+++ b/agents.tf
@@ -8,7 +8,7 @@ module "agents" {
   for_each = local.agent_nodes
 
   name                       = "${var.use_cluster_name_in_node_name ? "${var.cluster_name}-" : ""}${each.value.nodepool_name}"
-  ssh_keys                   = [hcloud_ssh_key.k3s.id]
+  ssh_keys                   = [local.hcloud_ssh_key_id]
   ssh_public_key             = var.ssh_public_key
   ssh_private_key            = var.ssh_private_key
   ssh_additional_public_keys = var.ssh_additional_public_keys
--- a/agents.tf
+++ b/agents.tf
@@ -11,7 +11,7 @@ module "agents" {
   ssh_keys                   = [hcloud_ssh_key.k3s.id]
   ssh_public_key             = var.ssh_public_key
   ssh_private_key            = var.ssh_private_key
-  additional_ssh_public_keys = var.additional_ssh_public_keys
+  ssh_additional_public_keys = var.ssh_additional_public_keys
   firewall_ids               = [hcloud_firewall.k3s.id]
   placement_group_id         = var.placement_group_disable ? 0 : element(hcloud_placement_group.agent.*.id, ceil(each.value.index / 10))
   location                   = each.value.location
--- a/agents.tf
+++ b/agents.tf
@@ -7,17 +7,17 @@ module "agents" {
 
   for_each = local.agent_nodes
 
-  name                   = "${var.use_cluster_name_in_node_name ? "${var.cluster_name}-" : ""}${each.value.nodepool_name}"
-  ssh_keys               = [hcloud_ssh_key.k3s.id]
-  public_key             = var.public_key
-  private_key            = var.private_key
-  additional_public_keys = var.additional_public_keys
-  firewall_ids           = [hcloud_firewall.k3s.id]
-  placement_group_id     = var.placement_group_disable ? 0 : element(hcloud_placement_group.agent.*.id, ceil(each.value.index / 10))
-  location               = each.value.location
-  server_type            = each.value.server_type
-  ipv4_subnet_id         = hcloud_network_subnet.agent[[for i, v in var.agent_nodepools : i if v.name == each.value.nodepool_name][0]].id
-  packages_to_install    = local.packages_to_install
+  name                       = "${var.use_cluster_name_in_node_name ? "${var.cluster_name}-" : ""}${each.value.nodepool_name}"
+  ssh_keys                   = [hcloud_ssh_key.k3s.id]
+  ssh_public_key             = var.ssh_public_key
+  ssh_private_key            = var.ssh_private_key
+  additional_ssh_public_keys = var.additional_ssh_public_keys
+  firewall_ids               = [hcloud_firewall.k3s.id]
+  placement_group_id         = var.placement_group_disable ? 0 : element(hcloud_placement_group.agent.*.id, ceil(each.value.index / 10))
+  location                   = each.value.location
+  server_type                = each.value.server_type
+  ipv4_subnet_id             = hcloud_network_subnet.agent[[for i, v in var.agent_nodepools : i if v.name == each.value.nodepool_name][0]].id
+  packages_to_install        = local.packages_to_install
 
   private_ipv4 = cidrhost(hcloud_network_subnet.agent[[for i, v in var.agent_nodepools : i if v.name == each.value.nodepool_name][0]].ip_range, each.value.index + 101)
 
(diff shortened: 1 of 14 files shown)


mysticaltech:
@batthebee Looks good to me!! Great work!

It should work with Terraform cloud in "local" mode correct?

@phaer, what do you think?

mysticaltech:
@strowi please don't hesitate to have a look too! 🙏

agents.tf
```diff
@@ -7,17 +7,17 @@ module "agents" {
...
-  placement_group_id     = var.placement_group_disable ? 0 : element(hcloud_placement_group.agent.*.id, ceil(each.value.index / 10))
-  location               = each.value.location
-  server_type            = each.value.server_type
-  ipv4_subnet_id         = hcloud_network_subnet.agent[[for i, v in var.agent_nodepools : i if v.name == each.value.nodepool_name][0]].id
-  packages_to_install    = local.packages_to_install
+  name                       = "${var.use_cluster_name_in_node_name ? "${var.cluster_name}-" : ""}${each.value.nodepool_name}"
+  ssh_keys                   = [hcloud_ssh_key.k3s.id]
+  ssh_public_key             = var.ssh_public_key
+  ssh_private_key            = var.ssh_private_key
+  additional_ssh_public_keys = var.additional_ssh_public_keys
```
phaer, line 14: That's a good change, thanks! I'd personally prefer `ssh_additional_public_keys`, so that the `ssh_` prefix is consistently at the beginning. But it's good to merge anyway :)

variables.tf
```diff
@@ -212,6 +213,7 @@ variable "rancher_registration_manifest_url" {
   type        = string
   description = "The url of a rancher registration manifest to apply. (see https://rancher.com/docs/rancher/v2.6/en/cluster-provisioning/registered-clusters/)"
   default     = ""
+  sensitive   = true
```
phaer, line 216: Why is this sensitive and why change it here? I don't think it's related to SSH?

modules/host/locals.tf
```diff
@@ -1,19 +1,16 @@
...
+  # ssh_client_identity is used for ssh "-i" flag, its the private key if that is set, or a public key
   # if an ssh agent is used.
-  ssh_identity_file = var.private_key == null ? var.public_key : var.private_key
-
-  # shared flags for ssh to ignore host keys, to use our ssh identity file for all connections during provisioning.
-  ssh_args = "-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i ${local.ssh_identity_file}"
+  ssh_client_identity = var.ssh_private_key == null ? var.ssh_public_key : var.ssh_private_key
+  # ssh_client_identity_file is used to (temporary) store identity informations (from ssh_client_identity),
+  # used by the ssh client.
+  ssh_client_identity_file = "/tmp/${local.name}-ssh-identity"
```
phaer, line 13: I see that you are correctly setting 0600 as permissions for this file below. I think ideally, we would have a way to randomize a prefix as with `mktemp` in a shell, but that would make it a bit more involved to remember the generated directory name for cleanup later on. Maybe using terraforms random provider could help here?

modules/host/main.tf
```diff
@@ -56,8 +64,8 @@ resource "hcloud_server" "server" {
   # Issue a reboot command and wait for MicroOS to reboot and be ready
   provisioner "local-exec" {
     command = <<-EOT
-      ssh ${local.ssh_args} root@${self.ipv4_address} '(sleep 2; reboot)&'; sleep 3
-      until ssh ${local.ssh_args} -o ConnectTimeout=2 root@${self.ipv4_address} true 2> /dev/null
+      ssh ${local.ssh_args} -i ${local.ssh_client_identity_file} root@${self.ipv4_address} '(sleep 2; reboot)&'; sleep 3
```
phaer, line 67: If I understand this correctly, those lines are the only reason why `local.ssh_client_identity_file` is needed, right? And all of them seem to be executed by the local shell, which we already assume to be bash(-compatible). 
If so, couldn't we just do `ssh ${local.ssh_args} -i <(echo "${local.ssh_client_identity}")` instead? This would avoid having to use a tempfile and just pass a file descriptor directly from bash to ssh. As echo is a builtin, the key wouldn't be visible in the process tree. (see [bash manual: process substitution](https://www.gnu.org/software/bash/manual/html_node/Process-Substitution.html)

phaer:
Hi @batthebee, thanks for the initiative! Looks good to me in general, I just read the code and haven't tested a deployment with your changes yet, but I do believe we could do without writing the private key to a local temp file, see last comment in the review and let me know what you think!

variables.tf
strowi: The `rancher_registration_manifest_url` contains the rancher-ui server + token to register with an  UUID. Not sure if that url persists after the import, but there is always the time in-between and it is better to be safe.

strowi:
Niec work. From a first look this should work. But i won't be able to test it until next week.

variables.tf
phaer: Ah makes sense, thanks! Ideally that would be an extra commit if not PR for visibility in git history, but that's less important than getting it fixed in master then :)

versions.tf
```diff
@@ -1,4 +1,5 @@
 terraform {
+  required_version = ">= 1.2.0"
```
strowi, line 2: Somehow this won't take from the module, can we add this also to the `kube.tf.example` to make it work for new users?

modules/host/main.tf
mysticaltech: @phaer I believe this was tried, but it failed before of the lack of permission (0600) when using this approach.

phaer: Ah right, and even if that wasn't an issue, SSH would close it's FDs too early. Should have checked. So the tempfile is fine, sorry for the noise.

mysticaltech: All good; thank you for the great review!

versions.tf
batthebee: done

modules/host/locals.tf
batthebee: I have now taken it completely out of the locals and just create a random name for it. Is that what you meant?

agents.tf
batthebee: done

batthebee:
@mysticaltech I have now tried this with terraform cloud and remote execution, and it works. Local should also work.

batthebee:
> Nice work. My thoughts initially were to utilize data.hcloud_ssh_key_s_ and then somehow check through the fingerprints. But this should work. ;)
> 
> I can only test it earliest next week.

@strowi  Sorry, I had linked the wrong issue.

I have now at least implemented that you can pass an ssh id and then no extra key is created. Your proposed solution would be more elegant. Maybe we can add it later.

phaer (approved):

strowi (approved):

mysticaltech:
Awesome work @batthebee! 🥳 Will merge and push ASAP this weekend!  🚀

mysticaltech:
Just shipped with v1.3.0! 🥳

Issue #55, opened 2022-02-09
When using allow_scheduling_on_control_plane=true, lb does not point to control planes


leonsteinhaeuser:
**What I did:**

```bash
git clone https://github.com/kube-hetzner/kube-hetzner.git
git checkout staging

export TF_VAR_hcloud_token="my-hcloud-token"
```

Create a terraform.tfvars file:
```hcl
# You need to replace these
public_key   = "~/.ssh/id_ed25519.pub"
# Must be "private_key = null" when you want to use ssh-agent, for a Yubikey like device auth or an SSH key-pair with passphrase
private_key  = "~/.ssh/id_ed25519"

# These can be customized, or left with the default values
# For Hetzner locations see https://docs.hetzner.com/general/others/data-centers-and-connection/
# For Hetzner server types see https://www.hetzner.com/cloud
location                  = "nbg1" # change to `ash` for us-east Ashburn, Virginia location
network_region            = "eu-central" # change to `us-east` if location is ash
agent_server_type         = "cpx21"
control_plane_server_type = "cpx31"
lb_server_type            = "lb11"
servers_num               = 3
agents_num                = 0

# If you want to use a specific Hetzner CCM and CSI version, set them below, otherwise leave as is for the latest versions
# hetzner_ccm_version = ""
# hetzner_csi_version = ""

# If you want to kustomize the Hetzner CCM and CSI containers with the "latest" tags and imagePullPolicy Always, 
# to have them automatically update when the node themselve get updated via the rancher system upgrade controller, the default is "false".
# If you choose to keep the default of "false", you can always use ArgoCD to monitor the CSI and CCM manifest for new releases,
# that is probably the more "vanilla" option to keep these components always updated. 
# hetzner_ccm_containers_latest = true
# hetzner_csi_containers_latest = true

# If you want to use letsencrypt with tls Challenge, the email address is used to send you certificates expiration notices
traefik_acme_tls = true
traefik_acme_email = "my@mail.com"

# If you want to allow non-control-plane workloads to run on the control-plane nodes set "true" below. The default is "false".
allow_scheduling_on_control_plane = true
```

**What happened:**

All resources are properly scheduled, but the load balancer does not point to the control planes.

![lb_overview](https://user-images.githubusercontent.com/42437185/153226525-16e8510c-7922-4063-8145-1b4d65f34293.jpg)

**What I expect:**

I expect a reference from the load balancer to the control planes.

[...]

PR #206, opened 2022-05-26
Allow Hetzner LB to point at schedulable controlplane nodes

geraldwuhoo:
This PR introduces the `LegacyNodeRoleBehavior=false` feature gate to the CCM to allow the Hetzner loadbalancer to point at schedulable controlplane nodes. I'm guarding the feature gate behind the `allow_scheduling_on_control_plane` setting, but if you feel that this is not necessary, I can drop the guard and set it as default for all new clusters.

Resolves https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/55.

### DIFF
--- a/init.tf
+++ b/init.tf
@@ -128,7 +128,8 @@ resource "null_resource" "kustomization" {
     content = templatefile(
       "${path.module}/templates/ccm.yaml.tpl",
       {
-        cluster_cidr_ipv4 = local.cluster_cidr_ipv4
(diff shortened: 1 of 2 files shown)


mysticaltech:
It's perfect the way it is now! Simple and clean, well done! So have you tried it? Does it work even with your removing the "master" role?

mysticaltech:
Just re-read your comment from yesterday; it seems that it's the only change needed! Appreciate your help in fixing this issue! Thanks again!

PR #207, opened 2022-05-27
prefix kustomization-backup with cluster_name

strowi:
Same as #196 but for the missed `kustomization_backup.yaml` .

When using as a module more then once it will overwrite the file in the root-directory.


mysticaltech:
Thanks @strowi, awesome! 🙏

mysticaltech:
Released in 1.3.0 🥳

Issue #209, opened 2022-05-29
Extend Rancher-support to allow bootstrapping


strowi:
Hi,

@mysticaltech would you be open to modifying the `rancher.yaml.tmpl` to allow for 2 more variables like  the following?

```yaml
  valuesContent: |-
    ingress:
      tls:
        source: ${rancher_tls_source}
    hostname: ${rancher_hostname}
    replicas: ${number_control_plane_nodes}
    bootstrapPassword: ${rancher_bootstrap_password}
```

That would allow to set the `bootstrapPassword` and complete the setup (if the variable is set outside the module) by code instead of manually having to check the container logs for the `bootstrapPassword`. 

BTW i had some trouble getting the rancher-import working when the `ingress.tls.source´ was rancher - since `kubectl apply -k..` would fail with a TLS-error. Only way to circumvent was to use `curl -k .. | k apply ..` or save the file locally in `/var/post_install`.

mysticaltech:
@strowi Both would be useful I agree! Please PR. Will merge that one right away! _(Apologies for last time, I just did not want to add another provider to the mix)._

About the cert, I believe the error was maybe because you specifically set enable-cert-manger to false? Because by default it is needed for the current rancher setup to work.

But definitely, having rancher_tls_source to choose between the [three options](https://rancher.com/docs/rancher/v2.0-v2.4/en/installation/resources/advanced/helm2/helm-rancher/#choose-your-ssl-configuration) would be great!

strowi:
@mysticaltech On it, see the PR. ;) No worries, i can understand and was maybe a little too focused on "our" solution/fork. Since we want to automate as much as possible (even the removal of clusters).

Certs: I think when `tls.source=rancher` those certs are self-signed (as per the docs u linked). So kubectl/kustomize etc. will fail on those. If i use cert-manager and `tls.source=letsEncrypt` i get a fully signed cert which works with kubectl/kustomize. But this is not recommended for testing, since there is a exact domain-limit of 5 per week which i found out the hard way.

mysticaltech:
Awesome @strowi ! Just so you know, the rancher self signed works great for me! But it does need cert manager for these too (which gets installed automatically).

strowi:
Hmm okay. Guess i don't understand the docs about that, whats the difference then between letsEncrypt and rancher when they are both signed by LE in the end?
[...]

PR #211, opened 2022-05-29
+ extend rancher-configuration with tls-selection, bootstrap-password

strowi:
solves #209 

@mysticaltech please have a look ;)


mysticaltech:
Looks great man, will give it a quick test ASAP and merge! 👌

mysticaltech:
@strowi For the rancher_tls_source, both letsEncrypt and secret are not possible, because the former needs us to configure cert-manager, and the latter needs us to configure a secret!

mysticaltech:
Hello @strowi, I have removed you from the team. The reason for that is that even though your ideas are good, it feels your goals are not aligned enough with Kube-Hetzner. Your priority is your project, that is understandable, but even in the last PR, there were too many bugs and too little thinking about the consequences of adding things. 

Your PRs are always welcome, I will definitely consider them, but please do test them from all angles before submitting them! 

I am sorry man. Be well!

Issue #221, opened 2022-06-07
Don't set `load_balancer_public_ipv4` if `using_klipper_lb`


phaer:
Hello,

I just noticed the following snippet in our outputs.
```
output "load_balancer_public_ipv4" {
  description = "The public IPv4 address of the Hetzner load balancer"
  value = local.using_klipper_lb ? [
    for obj in module.control_planes : obj.ipv4_address
  ][0] : var.traefik_enabled == false ? null : data.hcloud_load_balancer.traefik[0].ipv4
}
```
...and i think that's non-optimal, as `load_balancer_public_ipv4` suggests, well, a load-balancer. It's fine as long as all nodes are up, but if one uses that output to set DNS records and the first control planes goes down (i.e. during upgrades), services will be unavailable due to how klipper works.
Additionally, using just the first node with klipper pushes all traffic through that node, contradicting another reason to use a load-balancer in the first place besides HA.

I think a better approach would be to use `load_balancer_public_ipv4` only if `using_klipper_lb` is false and to `null` otherwise. This would still allow users to i.e. set multiple DNS A records, one for each control plane IP, but would be less surprising.

Otherwise we should at least explicitly document it. What do you think?

otavio:
Providing the list of control plane modes seems better as it would use same output as a unified interface for any module using it.

phaer:
But using a list would break the public interface for IMO little benefit, as the same information is already available in `control_planes_public_ipv4`

PR #222, opened 2022-06-07
Remove load balancer IP from outputs if klipper is enabled

phaer:
See https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/221

### DIFF
--- a/output.tf
+++ b/output.tf
@@ -19,9 +19,7 @@ output "agents_public_ipv4" {
 
 output "load_balancer_public_ipv4" {
   description = "The public IPv4 address of the Hetzner load balancer"
-  value = local.using_klipper_lb ? [
-    for obj in module.control_planes : obj.ipv4_address
-  ][0] : var.traefik_enabled == false ? null : data.hcloud_load_balancer.traefik[0].ipv4
+  value = var.traefik_enabled == false ? null : data.hcloud_load_balancer.traefik[0].ipv4
 }
 
 output "kubeconfig_file" {


output.tf
```diff
@@ -19,9 +19,7 @@ output "agents_public_ipv4" {
 
 output "load_balancer_public_ipv4" {
   description = "The public IPv4 address of the Hetzner load balancer"
-  value = local.using_klipper_lb ? [
-    for obj in module.control_planes : obj.ipv4_address
-  ][0] : var.traefik_enabled == false ? null : data.hcloud_load_balancer.traefik[0].ipv4
+  value       = var.traefik_enabled == false ? null : data.hcloud_load_balancer.traefik[0].ipv4
```
mysticaltech, line 22: @phaer `traefik_enabled` can still be true if we are using Klipper, so many you need to test add "OR local.using_klipper_lb" in the condition to set to null, no?

otavio: Also, it'd be good to squash the commits.

phaer: @mysticaltech Ah, thanks! Don't use the klipper stuff myself and thought it wasn't possible to use it together with traefik. Hope the logic is sound now! :)

phaer: @otavio I believe it's best to get to know the history and customs of a given open source project a bit before commenting on such things. While I pretty much always use GitHubs "squash and merge" function during merge, please have quick look at the somewhat messy commit history of this project and see lots of unsquashed commits ;)

mysticaltech:
Awesome @phaer! Merging to staging will give one quick test ASAP before releasing a patch version. You can do it yourself if you feel like it! :)

PR #225, opened 2022-06-08
Feat/ccm extra vars

oujonny:


mysticaltech:
What happened, @oujonny? Interesting work there! But it wasn't needed, right? The load_balancer_location variable is enough to set the LB location, correct?

Don't hesitate; we can always help!

oujonny:
I messed up which my branches and accidently open this PR directly on the orginal repo instated within my fork. 😆 
But I will create a new PR now.

PR #226, opened 2022-06-09
Set default lb location to hcloud ccm

oujonny:
Hi
This PR sets these two environments variables to the hcloud ccm. The settings are the same as for the treafik lb, but this way they are default values for the hole cluster. 
More informations about these env here: https://github.com/hetznercloud/hcloud-cloud-controller-manager/blob/main/docs/load_balancers.md 

It is related to #175 .


mysticaltech:
Thanks, @oujonny, it looks good to me! Did you test it? Let me know; if it deploys fine on your side, I will merge it.

mysticaltech:
@oujonny Did you test it? Please confirm, thanks!

Issue #224, opened 2022-06-07
Feature request: Offer ways to configure reverse DNS


otavio:
### Discussed in https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/discussions/223

<div type='discussions-op-text'>

<sup>Originally posted by **otavio** June  7, 2022</sup>
On other deployments we've been doing, using standalone servers in Hetzner Cloud, we've done the reverse DNS for the nodes. This would be something we'd like to do for our K8S cluster as well. However, it seems we cannot do it without hacking the module. Are we missing something?</div>

mysticaltech:
I have tried giving this a quick shot and am blocking the conditional requirement - only create the [rdns ressource](https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs/resources/rdns) if the base_domain variable is present. I do not think it is doable (as Terraform doesn't support if-else blocks, in my understanding).

That would mean that if we need to support that feature, we would always need to apply the `rdns` and force users to provide a base domain. That is not something I want to impose on them, as it is not that important.

Maybe there is a way to accomplish this if-and-only if the base_domain is defined? For now, it's not obvious to me personally, but ideas are welcome!

mysticaltech:
Ah, the `count` meta-argument can be used to accomplish the conditional logic.

mysticaltech:
This feature has been implemented in the above PR and released as part of v1.4.0.

PR #230, opened 2022-06-11
added support for reserve dns domain setting

mysticaltech:
Implements https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/224.


mysticaltech:
@otavio FYI

otavio:
Worked just fine here, thanks!

PR #242, opened 2022-07-10
Remove icmp from WWW

ricristian:
Since ICMP doesn't have any active role for cluster creation it should be disabled because there is no point for bots which scans randomly ip's to find any of my public ip's available

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -137,15 +137,6 @@ locals {
       ]
     },
 
-    # Allow ping on ipv4
-    {
-      direction = "in"
-      protocol  = "icmp"
-      source_ips = [
-        "0.0.0.0/0"
-      ]
-    },
-
     # Allow basic out traffic
     # ICMP to ping outside services
     {


phaer:
Could you confirm whether adding a blocking rule to `var.extra_firewall_rules` works for you?
If not, we should make that one possible and maybe mention it in the example definition.

Users should definitely be able to disallow ICMPv4 if preferred, but I like having it on by default as it can ease debugging and things like "wait until this server is reachable". And we listen on tcp/22 for sshd anyway, so bots can just scan for that ;)

phaer (changes requested):
Should be easy to block, but allowed by default IMO.

ricristian:
> 

I can confirm that by adding this block in extra_firewall_rules icmp starts working again. 
`{
       direction       = "in"
       protocol        = "icmp"
       port            = ""
       destination_ips = []
     }`
     
    I was forgetting about 22 ... also for 22 and 6443 I should be able to disable those for cases where I haven't cloned this project. The main problem is that I can survive with updating locals.tf from .terraform/modules/kube-hetzner/locals.tf but it would mean that I have to push into my git project entire .terraform folder which is not quite nice and also in case that someone is updating this file for some purposes I'll have to maintain it.

phaer:
@ricristian just a heads up, i just edited my comment to add additional info, while you replied.

Regarding extra_firewall_rules: I meant the other way round; does it work for you, i.e. block icmpv4 if you use our master branch and add a rule to block it into your `var.extra_firewall_rules`? I am not sure if it would overrule the default rule, but it should. We might need to fine-tune our terraform expression here for it to work though.

ricristian:
I haven't saw any documentation regarding blocking some ports on https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs/resources/firewall and also I don't see any option in hetzner console for DROP / ALLOW selector. I think that the way of how is implemented right now firewall ( If source and range exists than allow )

phaer:
If the rules are "merged" on the backend, just setting `source_ips      = []`in your rule above could work? If not we should try to allow that to work in our terraform code and fall back to making the rule optional with a flag like `var.allow_icmpv4` or so. :)

ricristian:
Unfortunately if I try to set source_ip's to [] it will fail with `source_ips needs to have at least one entry` but if I try to set source_ips=["127.0.0.1/32"] the ICMP rule won't be merged instead it will be added as an extra rule

phaer:
I believe the most flexible way would be to replace "concat" by something smarter - merging both lists on items with same direction & protocol should suffice? 
https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/master/main.tf#L40
```terraform
  dynamic "rule" {
    for_each = concat(local.base_firewall_rules, var.extra_firewall_rules)
    content {
      direction       = rule.value.direction
      protocol        = rule.value.protocol
      port            = lookup(rule.value, "port", null)
      destination_ips = lookup(rule.value, "destination_ips", [])
      source_ips      = lookup(rule.value, "source_ips", [])
    }
  }
}
```

ricristian:
I think so. It should do the trick

mysticaltech:
All for the flag seems like the cleanest option!!

mysticaltech:
@phaer The flag could be very quick to implement, but I was thinking about places in the code where we ping the server in the past to see if it was up; I am not sure we still use that logic; it might have been replaced by pure SSH try and repeat. Will check ASAP.

phaer:
@mysticaltech Grepping for "ping" in the code base only yields the firewall rule, so we are  probably good here. It should still deploy cleanly if a flag is added and icmpv4 is forbidden.

mysticaltech:
@phaer Exactly; I am looking at it now and we do not use pinging in the deploy anymore, so will just add the flag quickly.

Issue #240, opened 2022-07-07
Expose loadbalancer IPv4 when Traefik disabled.


geraldwuhoo:
Currently the code in `output.tf` returns `null` for `load_balancer_public_ipv4` if Traefik is disabled. I can understand the reasoning, because without an ingress controller to provision an IP from the Hetzner LB, there is no LB IP to speak of. However, I'm disabling Traefik to install nginx-ingress instead. I'm sure many others disable Traefik to instead install their own preferred ingress controller.

Since there is no way for the Terraform module to detect whether or not a user will provision an ingress controller *after* Terraform provisioning, I think it makes sense to err on the side of caution, and return proper `load_balancer_public_ipv4` value even when Traefik is disabled.

mysticaltech:
Hello @geraldwuhoo, it makes sense! PRs welcome, that should be a quick fix.

@phaer FYI.

mysticaltech:
Moving this to a discussion in the ideas section, @geraldwuhoo, If you decide to take this on, I will convert it back to an issue to properly close it and give proper attributions.

PR #243, opened 2022-07-11
Expose loadbalancer IPv4 when Traefik disabled.

geraldwuhoo:
This PR removes the check for Traefik enabled when exposing the loadbalancer IPv4. Addresses https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/240.

### DIFF
--- a/output.tf
+++ b/output.tf
@@ -19,7 +19,7 @@ output "agents_public_ipv4" {
 
 output "load_balancer_public_ipv4" {
   description = "The public IPv4 address of the Hetzner load balancer"
-  value       = (local.using_klipper_lb == true || var.traefik_enabled == false) ? null : data.hcloud_load_balancer.traefik[0].ipv4
+  value       = local.using_klipper_lb ? null : data.hcloud_load_balancer.traefik[0].ipv4
 }
 
 output "kubeconfig_file" {


mysticaltech:
@geraldwuhoo Now while reviewing your change, I remembered why we have this.

The hcloud_load_balancer data just fetches the IP of the load balancer, it does not create anything, it is the CCM, on the request of an SVC load balancer that creates a load balancer on Hetzner. 

So THE ONLY WAY, TO KNOW FOR SURE THAT THERE WILL BE ONE, is to test if Traefik is enabled. If it is, we know that there will be an IP to fetch. If it is not enabled, we do not have any info on other ingress controllers that would have eventually created an svc load balancer too like nginx. We could very well be in a case where someone just does not wish to have a load balancer at all or gave it a name other than "Traefik", in which case the data.hcloud_load_balancer.traefik[0] would be null, and data.hcloud_load_balancer.traefik[0].ipv4 would fail to execute.

All that to say that the check on `traefik_enabled` is actually essential, as it ensures that data.hcloud_load_balancer.traefik[0] will not be null and executing data.hcloud_load_balancer.traefik[0].ipv4 would not result in an error. And since we basically have no control over other ingress controllers, and the name of the load_balencer they may choose to create, we can't do anything to fetch their IPs.

Hope I was clear enough?! If you can think of another way, please let me know.

mysticaltech:
Now maybe instead of a data call to hcloud_load_balencer of name Traefik, like in https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/944e5567d721a0f2070820a200fd208291466c90/main.tf#L63, we could instead make a call to https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs/data-sources/load_balancers, which is hcloud_load_balancer**s** plural, without passing any name attributes. In which case, we would test if there is one, if so return the first element, and that the first value to grab the IP to display in the output. That in fact could work IMHO.

So we would delete line 63 in `main.tf`, and ideally move the logic to `data.tf`, as this is where it belongs anyway, it was an initial error on our part not to have it there.

If you can try that, it would be great, if not I may have time to work on that by the end of the week.

mysticaltech:
Hey @geraldwuhoo, what do you think of the above? Can you do it?

mysticaltech:
Hey @geraldwuhoo, this has been inactive and the change proposed is not valid due to the reasons explained above. Closing this for now, but please do not hesitate to submit a new PR that considers the needs above. You are correct that we need to show any created LB, not just those by Traefik.

PR #244, opened 2022-07-11
Add volume storage option to longhorn

ricristian:
Now we can have an option where we can specify that we need more packages installed on VM's out of the box. 

I've updated a little bit longhorn part because:
1. Using only local storage is not reliable due to the fact that max size for storage dependes on flavour size ( max 360GB ) + extra pricing involved on scaling vertically  
2. If rancher is not enabled there is no UI option to see longhorn statistics like degraded volumes / replicas or issues inside it

Advantage: 
1. Volumes are created and mounted automatically inside VM's 
2. In case that after some time we need to scale up volumes we only change the "longhorn_volume_size" and resize is done automatically for both sides ( cloud level and vm level ) 

Disadvantage:
1. Volumes cannot be scaled down due to cloud provider so we cannot impact here anything but it worth mentioned
2. Default behaviour is the initial one with local storage but for example if someone want's to move longhorn from local storage to volume storage this is not possible ... so this "Longhorn on volumes" should be implemented at the initial bootstrap of the cluster. 

Thanks,


ricristian:
> Thank you, @ricristian, excellent contribution! Sorry for the delay I somehow missed the email notification. I will give it one test deploy ASAP, then merge. In the meantime, if you could just run "terraform format" and push again, it would be great! 🙏

Done.

karbowiak:
So in kube.tf we just add `longhorn_volume_size` and then it'll create a volume pr. node, mount it and use it for longhorn ?
Fancy!

Is it possible to use both local storage AND/OR the mounted volumes ?

ricristian:
So in kube.tf we just add longhorn_volume_size and then it'll create a volume pr. node, mount it and use it for longhorn ? - YES

Is it possible to use both local storage AND/OR the mounted volumes ? - As far as I know not by default because longhorn will be looking into a path /var/longhorn or something. 
Since we’re on cloud area still thinking to  storage as we’re used on plain VPS. ( what happens if we change flavour or we need to change OS or stuff like that ? I think that data will be lost ? ) Fortunately  these days volume storage is quite cheap no matter the cloud. But I must to say that I’m not an expert so I might say something wrong so please let me know.

karbowiak:
@ricristian oh yeah true - so it mounts it into the host at /var/longhorn

I was just kinda hoping it would be toggleable on a pr. node basis (or group of nodes)..
Would it make sense to make it dependent on a flag in the kube.tf that enables/disables it for a group of nodes perhaps?

Kinda like the storage node in the example
```
    {
      name        = "storage",
      server_type = "cpx21",
      location    = "fsn1",
      labels = [
        "node.kubernetes.io/server-usage=storage"
      ],
      taints = [
        "server-usage=storage:NoSchedule"
      ],
      count = 1
    }
 ```
 
 Maybe make it something like..
 
 ```
    {
      name        = "storage",
      server_type = "cpx21",
      location    = "fsn1",
      labels = [
        "node.kubernetes.io/server-usage=storage"
      ],
      taints = [
        "server-usage=storage:NoSchedule"
      ],
      count = 1,
      local_storage_longhorn = true # true to use local storage, false to use mounted hcloud volume
    }
  ```

What do you think @mysticaltech :)

**Comment by ricristian** [comment]
It could but I don’t recommend to use local storage anymore. The Hetzner CSI Is good for databases for example but in cases where we need something like EFS for readwritemany ( multiple pods reading / writing on same path ) in those cases we would need longhorn or open ebs , rook whatever but the storage will be consumed quite fast since there will be 3 replicas and for moving longhorn from local storage to Hetzner Volumes requires manual action which breaks the idea of automation.

**Comment by karbowiak** [comment]
@ricristian was more so it could be enabled/disabled on a "pool" of nodes at a time - size wise Longhorn should be able to auto configure.

Afaik it even has some labels that could be used in a split setup.

https://longhorn.io/docs/1.3.0/volumes-and-nodes/storage-tags/
No idea how feasible it is to add this tho 😅 

That said, i think it looks fine - personally i would prefer a way to enable/disable it (With default being disabled so old setups don't suddenly lose all their volumes)

**Comment by mysticaltech** [comment]
You both have valid points, folks, but since we are not going to bump a major version here, we need to stay as backward compatible as possible, @ricristian. If you could have the switching logic to choose either at cluster set up time it would be great! 🙏

**Comment by mysticaltech** [comment]
I myself hoped in the past that all of this would be configurable via the rancher longhorn GUI, is it not so?

**Comment by karbowiak** [comment]
@mysticaltech don't think so, you could view it but that's about it 😅

**Comment by mysticaltech** [comment]
@ricristian If you need a hand in adding the option to have both the "old" setup and your new one, please don't hesitate. But I would be more at ease if you did it because you seem to know Longhorn very well and could maybe enhance the "old" local storage setup while keeping it backward compatible.

**Comment by ricristian** [comment]
Yes I would need some help because now I’m kinda lost when it comes to how you want to look this on the end regarding backwards compatibility

**Comment by mysticaltech** [comment]
Alright, will give this one a shot ASAP.

File: templates/longhorn.yaml.tpl [hunk 1]
```diff
@@ -23,7 +25,7 @@ metadata:
 spec:
   chart: longhorn
   # Using this repo makes it compatible with Rancher
-  repo: https://charts.rancher.io
+  repo: https://charts.%{ if rancher_enabled ~}rancher%{ else ~}longhorn%{ endif ~}.io
```
mysticaltech, line 28: @ricristian I for instance, deploy a separate 1 node cluster for rancher, so in that case, I won't have compatibility for it. So will just keep the rancher repo if you don't mind. They manage the project anyways, so it will always have the latest and greatest.

agents.tf
```diff
@@ -87,3 +87,46 @@ resource "null_resource" "agents" {
     hcloud_network_subnet.agent
   ]
 }
+
+resource "hcloud_volume" "volume" {
+  for_each = { for key, val in local.agent_nodes : key => val if val.longhorn_volume_size >= 10 }
```
mysticaltech, line 92: @karbowiak Basically, I will just need to check on a variable flag in the if here to enable to old behavior (local disk) and the same for the other added resource. Please correct me if I am wrong?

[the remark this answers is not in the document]
ricristian: Sorry for late reply. I'm a little bit caught. I'll try to do this tomorrow :)

mysticaltech:
@ricristian @karbowiak Wouls that new approach improve Longhorn's performance in your opinion?

@PurpleBooth FYI.

agents.tf
karbowiak: @mysticaltech sorry for the _LATE_ reply 😅 

But yes that does look correct - for each agent_node check if the longhorn_volume_size is larger than (or equals) 10GB, if it is then it's true otherwise false.

karbowiak:
@mysticaltech probably not improve performance outright, the local disks should be faster than the mounted ones - since they're added from a datacenter-storage-thingymajig.

Afaik it used to be based on CEPH but was changed..

That said, it will allow larger storage blobs to be created for Longhorn :)

ricristian:
> @ricristian @karbowiak Would that new approach improve Longhorn's performance in your opinion?
> 
> @PurpleBooth FYI.

Depends on the situation ... yes local volumes might be faster han attached one but what happens when you need only more storage not necessarily more IOPS ? 

I'm so sorry for being so much away

mysticaltech:
Got it, thanks for the info @karbowiak 🙏

karbowiak:
@ricristian agree 100%, if you need more storage than the local disk gives, then it makes total sense to switch to the Hetzner provided volumes.

If you need higher IOPS tho, the local disks are the way forward.

Also why i think it's a good compromise that it's a selectable pr. node, all we really need is a way to label it so Longhorn can assign volumes to appropriate nodes, but that's something that can come later, for now it's 10/10 that it's enable/disable for the nodes :)

mysticaltech:
Thanks for the input @ricristian and @karbowiak! 

@PurpleBooth The above should answer your questions. I guess a local disk could be what you need to maximize performance. This PR will ultimately allow using either volumes or local-disk (which is the current way) for Longhorn.

mysticaltech:
@ricristian I need to merge this functionality ASAP; been dangling for too long! :) Please just share with me your kube.tf without sensitive values so that I am sure how you use it.

Also, do you thing the resize2fs would work the same with zfs?

mysticaltech:
Thanks, @ricristian appreciate it!

mysticaltech (approved):
All good for now, will merge to staging and do the final tweaks there.

mysticaltech:
Thanks again for this PR @ricristian; I did not have to make many changes, and it added lots of value to the project! Now people can use Longhorn to the fullest of its possibilities.

Issue #245, opened 2022-07-13
Hetzner Load Balancer still gets created when using use_klipper_lb = "true"


andi0b:
I'm trying to create a deployment without a Hetzner Cloud Load Balancer. The setting `use_klipper_lb = "true"` seems to work well, and the k8s LoadBalancer gets the all the nodes IPs assigned, instead of a LoadBalancer IP.

But there are still Hetzner Cloud Load Balancers created. I see them in the Hetzner Console and I get charged for them.

Any idea how to fix that?

mysticaltech:
@andi0b Please post the out of `kubectl get svc -A -o wide` here. It could be something else requesting it from the Hetzner CCM.

andi0b:
Do I misunderstand the functionality of `use_klipper_lb`? I expect it to replace the Hetzner LB and expose a Service of type LoadBalancer to all cluster nodes IPs. So I can point a round-robin DNS record to it and do some random/basic load balancing, without paying for a Hetzner LB.

```
❯ kubectl get nodes -o wide
NAME                     STATUS   ROLES                       AGE     VERSION        INTERNAL-IP    EXTERNAL-IP    OS-IMAGE           KERNEL-VERSION     CONTAINER-RUNTIME
control-plane-nbg1-gjd   Ready    control-plane,etcd,master   4m52s   v1.23.8+k3s2   10.254.0.102   162.55.57.0    openSUSE MicroOS   5.18.9-2-default   containerd://1.5.13-k3s1
control-plane-nbg1-swr   Ready    control-plane,etcd,master   5m25s   v1.23.8+k3s2   10.254.0.101   162.55.56.19   openSUSE MicroOS   5.18.9-2-default   containerd://1.5.13-k3s1

❯ kubectl get services -o wide -A
NAMESPACE         NAME                            TYPE           CLUSTER-IP      EXTERNAL-IP                PORT(S)                      AGE     SELECTOR
cert-manager      cert-manager                    ClusterIP      10.43.183.227   <none>                     9402/TCP                     84s     app.kubernetes.io/component=controller,app.kubernetes.io/instance=cert-manager,app.kubernetes.io/name=cert-manager
cert-manager      cert-manager-webhook            ClusterIP      10.43.216.62    <none>                     443/TCP                      84s     app.kubernetes.io/component=webhook,app.kubernetes.io/instance=cert-manager,app.kubernetes.io/name=webhook
[...]

PR #246, opened 2022-07-15
Disable CCM Load Balancers when using Klipper, fix default LB type

andi0b:
1. Setting `HCLOUD_LOAD_BALANCERS_ENABLED` for the Hetzner Cloud CCM to `false` when klipper LB is enabled, and otherwise to `true`.
    - This fixes the problem of getting unneeded and not properly working Hetzner LBs on single node clusters, see #245 

1. The default for `load_balancer_type` was `cpx11` which is not a valid load balancer type. Set is to `lb11` like in `kube.tf.example`
    - Keeping the LB type unset in the kube.tf got the deployment stuck at `Waiting for load-balancer to get an IP...` and the error `hcloud/loadBalancers.EnsureLoadBalancer: hcops/LoadBalancerOps.Create: load balancer type cpx11 not found (invalid_input)`

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ There are some limitations (to scaling down mainly) that you need to be aware of
 
 _Once the cluster is up; you can change any nodepool count and even set it to 0 (in the case of the first control-plane nodepool, the minimum is 1); you can also rename a nodepool (if the count is to 0), but should not remove a nodepool from the list after once the cluster is up. That is due to how subnets and IPs get allocated. The only nodepools you can remove are those at the end of each list of nodepools._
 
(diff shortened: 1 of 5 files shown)


andi0b:
@phaer  About `using_hetzner_lb`, I did it the same way as the other template in the block before does it (`traefik_config.yaml.tpl`). And tested that it actually works in true and false cases. So I would rather like to not touch it any more. But feel free to optimize it :)

https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/36549df991cda5d3d61b8a192e8ffccc60a92a65/init.tf#L118-L146

mysticaltech:
@andi0b, that works; thank you! And good catch for the LB default type! 😅 I would have used `!using_klipper_lb` too, but it's not an issue. Merging! 🙏

Issue #249, opened 2022-07-28
Changing the DNS server leads to unroutable storage box


PurpleBooth:
When the DNS servers are changed it leads to the storage box product being unreachable, as the IPv6 address they are connectable on is different when using the Hetzner provided DNS and the Google/Cloudflare ones.

Woud you be open to a pull request making, defaulting to on, the DNS server change optional?

mysticaltech:
@PurpleBooth We don't have official support for IPv6 yet! 

If you want to contribute a PR, it would be most welcome.

PurpleBooth:
Cool, will submit something in the upcoming days. Thanks

mysticaltech:
Please explain more, and share your kube.tf without the sensitive values if you can.

phaer:
@mysticaltech As I understand it:

We are currently using Googles & Cloudflares public DNS servers for nodes (see https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/81e73580db795484c00ed0a1dadc836cd0e437f8/modules/host/templates/userdata.yaml.tpl#L57). 

Hetzner runs its own DNS Servers (https://docs.hetzner.com/dns-console/dns/general/authoritative-name-servers) which seem to announce different IPs for some of their services (e.g. the "storage box") than other DNS servers do. At least on ipv6, probably on ipv4 as well.

So two issues
* ipv6 support: we don't need much. ingress via ipv6 is already working if using hetzner load balancers. Egress traffic to (some?) ipv6 addresses seems to be broken.
* custom dns servers: we should either allow users two customize used DNS servers and/or default to hetzners instead of Google & Cloudflare. 

is this correct @PurpleBooth? Would your PR address both of this issues? We don't have much maintainer bandwidth available atm to implement feature requests ourselves, but try to help new contributors where we can :)

PurpleBooth:
This is spot on @phaer!

mysticaltech:
Awesome @phaer. Thanks for the clear explanation!

PurpleBooth:
Closing in favour of the PR

PR #250, opened 2022-07-29
Allow the dns to be configurable

PurpleBooth:
Hetzner as authorative DNS servers on their cloud servers that resolve
different IP addresses for some of their products, such as Storage Box,
when we overwrite them, we stop those services from being connectable
from Hetzner Cloud.

This change allows users to choose their own DNS Servers (or default to
the current Google/Cloudflare combo), or even provide an empty list and
use Hetzner.

Relates-to: #249

### DIFF
--- a/modules/host/templates/userdata.yaml.tpl
+++ b/modules/host/templates/userdata.yaml.tpl
@@ -54,8 +54,10 @@ runcmd:
 - [sed, '-i', 's/NETCONFIG_NIS_SETDOMAINNAME="yes"/NETCONFIG_NIS_SETDOMAINNAME="no"/g', /etc/sysconfig/network/config]
 - [sed, '-i', 's/DHCLIENT_SET_HOSTNAME="yes"/DHCLIENT_SET_HOSTNAME="no"/g', /etc/sysconfig/network/dhcp]
 
-# We set Cloudflare DNS servers, followed by Google as a backup
-- [sed, '-i', 's/NETCONFIG_DNS_STATIC_SERVERS=""/NETCONFIG_DNS_STATIC_SERVERS="1.1.1.1 1.0.0.1 8.8.8.8"/g', /etc/sysconfig/network/config]
+%{ if length(dnsServers) > 0 }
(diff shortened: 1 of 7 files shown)


modules/host/templates/userdata.yaml.tpl
```diff
@@ -54,8 +54,10 @@ runcmd:
 - [sed, '-i', 's/NETCONFIG_NIS_SETDOMAINNAME="yes"/NETCONFIG_NIS_SETDOMAINNAME="no"/g', /etc/sysconfig/network/config]
 - [sed, '-i', 's/DHCLIENT_SET_HOSTNAME="yes"/DHCLIENT_SET_HOSTNAME="no"/g', /etc/sysconfig/network/dhcp]
 
-# We set Cloudflare DNS servers, followed by Google as a backup
-- [sed, '-i', 's/NETCONFIG_DNS_STATIC_SERVERS=""/NETCONFIG_DNS_STATIC_SERVERS="1.1.1.1 1.0.0.1 8.8.8.8"/g', /etc/sysconfig/network/config]
+%{ if length(var.dns_servers) > 0 }
```
phaer, line 57: Are you using this condition to explicitly allow users setting `var.dns_servers` to `[]` in order to set no static DNS servers and rely on DHCP?
If so, please add that feature to the variables description. I only noticed it when i wondered why you don't do 

```
  validation {
    condition     = length(var.dns_servers) == 0
    error_message = "Please set DNS servers."
  }
```

or something similar.

phaer (approved):
Small remark regarding explicitly empty values, but looks good to me! Thanks :tada: :rocket: :+1:

modules/host/templates/userdata.yaml.tpl
PurpleBooth: Should be in the docs now

mysticaltech:
Simple and clean @PurpleBooth. I will do a quick test deploy and merge!

mysticaltech (approved):

PR #251, opened 2022-07-29
Add cilium

PurpleBooth:
Cilium offers some fantastic observability functionality I would like to
take advantage of, along side DNS based filtering in network policies

### DIFF
--- a/control_planes.tf
+++ b/control_planes.tf
@@ -64,10 +64,10 @@ resource "null_resource" "control_planes" {
       advertise-address           = module.control_planes[each.key].private_ipv4_address
       node-label                  = each.value.labels
       node-taint                  = each.value.taints
-      disable-network-policy      = var.cni_plugin == "calico" ? true : var.disable_network_policy
+      disable-network-policy      = var.cni_plugin != "flannel" ? true : var.disable_network_policy
       write-kubeconfig-mode       = "0644" # needed for import into rancher
       },
-      var.cni_plugin == "calico" ? {
+      var.cni_plugin != "flannel" ? {
         flannel-backend = "none"
     } : {}))
 
--- a/init.tf
+++ b/init.tf
@@ -21,9 +21,9 @@ resource "null_resource" "first_control_plane" {
       advertise-address           = module.control_planes[keys(module.control_planes)[0]].private_ipv4_address
       node-taint                  = local.control_plane_nodes[keys(module.control_planes)[0]].taints
       node-label                  = local.control_plane_nodes[keys(module.control_planes)[0]].labels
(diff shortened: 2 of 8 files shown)


PurpleBooth:
Note: I had 5min to do this, so I did. It's the most bare minimum, and really some thought needs to be given to allowing access to the many options Cilium gives you. The ability to enable wireguard, or disable kube proxy from the get go would be of interest. Something is always better than nothing though.

mysticaltech:
@PurpleBooth Awesome! We can propose a sample HelmChartConfig to allow them to do more configurations. This project started with Cilium at the very beginning, so in the git history, we have a few working configs for Cilium.

Anyways, as you said, something is always better than nothing! I will play around with this PR ASAP. I just want to test it and see if we can improve anything (based on previous work).

PurpleBooth:
I got a bit of time upcoming so shout if you want some changes

PurpleBooth:
So I changed the approach a bit, I used the cilium binary, rather than helm crd because I couldn't figure out how to get it to have the right tolerations on the pods it launches, so they were never being scheduled.

The way it's installed basically means it's doing a helm install.

mysticaltech:
Oh man, I know how to fix this non-scheduling issue via helm, just need to pass the right values. Have them all in the git history. Just haven't had time to focus on that. Will try ASAP. 

I prefer like it was before, as the same format we use for everything else, and easier to reconfigure via helm chart config. Also probably easier to update! At least the flow is the same as other addons that we offer.

Will find some time, hopefully tomorrow.

PurpleBooth:
I have a copy of the old way here if it's useful https://github.com/PurpleBooth/terraform-hcloud-kube-hetzner/tree/add-cilium

mysticaltech:
Perfect, definitely 🙏

mysticaltech:
@PurpleBooth Thank you so much for your contributions to this project. We would like to invite you to be a member of our team on Github. It will allow you to create branches (which will make it easier for us to work together on a PR), approve stuff and manage issues if you wish so. Whatever you see fit! ✨

It would also allow you to better influence the direction of this project. We are six people, but most of us do not do much. The current most active members are Phaer and me. I am usually the one to do the releases and was the one who originated the project. It would be an honor to have you be one of us! We all have day jobs, so don't worry; nothing is expected of you; some of our members do not show up for weeks or months! 😁

PurpleBooth:
Lets chat, I tried to login to that room, however I get a "You are not invited to this room.", I am "`@BillieCodes:matrix.org
`" on matrix

mysticaltech:
Hey @PurpleBooth, I just invited you!

PR #252, opened 2022-07-30
Check for formatting in subdirectories

PurpleBooth:
Currently only the main directory is being linted. This will make sure
that the whole project is formatted nicely


mysticaltech:
Beautiful!

PR #253, opened 2022-07-31
Allow additional packages to be installed on the nodes

PurpleBooth:
When using some csi, such as the smb csi, you need additional packages
on the host nodes. Currently you need an additional null resource to
install these, and that will cause problems if autoscaling is ever
implemented. This change allows the user to specify additional packages
to include.

This has already been implemented in #244 however that seems to have
stalled, so this change reinstates the idea.


mysticaltech:
@PurpleBooth Thanks for this, the storage PR just needs some little work that the submitter doesn't have time to implement. So was going to do it myself.

I love the idea of compartimentalizing packages, between agents and control plane, but we are not using that in the cose, are we?

PurpleBooth:
Not a good reason, I did it just in case. I was a little torn on it myself. I will simplify it to one

mysticaltech:
Ok perfect :)

PurpleBooth:
Moved to #257

PR #254, opened 2022-07-31
Loadbalance control plane

PurpleBooth:
Currently all api-server connections go to the first control-plane node.
This means that if kured is restarting that node, an API user can
experience downtime. This change loadbalances the incoming requests via
a loadbalancer, meaning the service never appears down.

To facilitate that, we also add a few new labels to resources to allow
for easy identification.

This should mean that the control plane is highly available.


mysticaltech:
Will review ASAP! Thanks for that, beautiful 🙏🙏✨

PurpleBooth:
moved to repository branch #256

PR #256, opened 2022-08-01
Loadbalance control plane

PurpleBooth:
Currently all incoming connections go to the first control-plane node.
This means that if kured is restarting that node, an API user can
experience downtime. This change loadbalances the incoming requests via
a loadbalancer, meaning the service never appears down.

To facilitate that, we also add a few new labels to resources to allow
for easy identification.

This also means the cluster will be highly available, as we can use the
loadbalancer rather than node IPs for the server ip address.

### DIFF
--- a/agents.tf
+++ b/agents.tf
@@ -23,10 +23,7 @@ module "agents" {
 
   private_ipv4 = cidrhost(hcloud_network_subnet.agent[[for i, v in var.agent_nodepools : i if v.name == each.value.nodepool_name][0]].ip_range, each.value.index + 101)
 
-  labels = {
-    "provisioner" = "terraform",
-    "engine"      = "k3s"
-  }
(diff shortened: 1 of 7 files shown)


mysticaltech:
@PurpleBooth I like this solution because it's clean! And would cost just a tiny bit more than implementing a bastion. Even if we implement the latter, it would be great to keep this possibility.

mysticaltech:
Ok, to merge on my end, would you like to add anything for now @PurpleBooth?

otavio (approved):

PurpleBooth:
Cool! Nope, go for it, I need to pop out but you can fix the conflicts or I will do it when I am back home in a few hours

mysticaltech:
@PurpleBooth When I try to deploy with the flag on, I get this issue, and it doesn't succeed.

ksnip_20220801-154529

ksnip_20220801-154954

PR #257, opened 2022-08-01
Allow additional packages to be installed on the nodes

PurpleBooth:
When using some csi, such as the smb csi, you need additional packages
on the host nodes. Currently you need an additional null resource to
install these, and that will cause problems if autoscaling is ever
implemented. This change allows the user to specify additional packages
to include.

This has already been implemented in #244 however that seems to have
stalled, so this change reinstates the idea.


mysticaltech:
Awesome!

PR #258, opened 2022-08-01
Add cilium

PurpleBooth:
Cilium offers some fantastic observability functionality take advantage
of, along side DNS based filtering in network policies

### DIFF
--- a/cilium_values.yaml.example
+++ b/cilium_values.yaml.example
@@ -1,0 +1,30 @@
+# All Cilium helm values can be found at https://github.com/cilium/cilium/blob/master/install/kubernetes/cilium/values.yaml
+# The following is a personal working and optimized example of Cilium helm values.
+
+ipam:
+  # -- Configure IP Address Management mode.
+  mode: kubernetes
+
+# -- Specify which network interfaces can run the eBPF datapath. This means
+# that a packet sent from a pod to a destination outside the cluster will be
+# masqueraded (to an output device IPv4 address), if the output device runs the
+# program. When not specified, probing will automatically detect devices.
+devices: "eth1"
+
+# -- Configure the encapsulation configuration for communication between nodes.
+# Possible values:
+#   - disabled (native routing works, however it feels that geneve is more stable, but we may be wrong)
+#   - vxlan
+#   - geneve (it's basically a better version of vxlan)
+tunnel: geneve
+
+# -- Configure Kubernetes specific configuration
+k8s:
+  # -- requireIPv4PodCIDR enables waiting for Kubernetes to provide the PodCIDR
+  # range via the Kubernetes node resource
+  requireIPv4PodCIDR: true
+
+# -- Configure the kube-proxy replacement in Cilium BPF datapath
+# Valid options are "disabled", "probe", "partial", "strict".
+# ref: https://docs.cilium.io/en/stable/gettingstarted/kubeproxy-free/
+kubeProxyReplacement: strict
--- a/control_planes.tf
+++ b/control_planes.tf
@@ -94,19 +94,16 @@ resource "null_resource" "control_planes" {
       token                       = random_password.k3s_token.result
       disable-cloud-controller    = true
       disable                     = local.disable_extras
-      flannel-iface               = "eth1"
       kubelet-arg                 = ["cloud-provider=external", "volume-plugin-dir=/var/lib/kubelet/volumeplugins"]
       kube-controller-manager-arg = "flex-volume-plugin-dir=/var/lib/kubelet/volumeplugins"
       node-ip                     = module.control_planes[each.key].private_ipv4_address
       advertise-address           = module.control_planes[each.key].private_ipv4_address
       node-label                  = each.value.labels
       node-taint                  = each.value.taints
-      disable-network-policy      = var.cni_plugin == "calico" ? true : var.disable_network_policy
       write-kubeconfig-mode       = "0644" # needed for import into rancher
       },
-      var.cni_plugin == "calico" ? {
-        flannel-backend = "none"
-    } : {}))
+      lookup(local.cni_k3s_settings, var.cni_plugin, {}),
+    ))
 
     destination = "/tmp/config.yaml"
   }
--- a/control_planes.tf
+++ b/control_planes.tf
@@ -96,6 +96,7 @@ resource "null_resource" "control_planes" {
       disable                     = local.disable_extras
       kubelet-arg                 = ["cloud-provider=external", "volume-plugin-dir=/var/lib/kubelet/volumeplugins"]
       kube-controller-manager-arg = "flex-volume-plugin-dir=/var/lib/kubelet/volumeplugins"
+      flannel-iface               = "eth1"
       node-ip                     = module.control_planes[each.key].private_ipv4_address
       advertise-address           = module.control_planes[each.key].private_ipv4_address
       node-label                  = each.value.labels
@@ -103,6 +104,9 @@ resource "null_resource" "control_planes" {
       write-kubeconfig-mode       = "0644" # needed for import into rancher
       },
       lookup(local.cni_k3s_settings, var.cni_plugin, {}),
+      var.use_control_plane_lb ? {
+        tls-san = [hcloud_load_balancer.control_plane.*.ipv4[0], hcloud_load_balancer_network.control_plane.*.ip[0]]
+      } : {}
     ))
 
     destination = "/tmp/config.yaml"
--- a/init.tf
+++ b/init.tf
@@ -16,6 +16,7 @@ resource "null_resource" "first_control_plane" {
       disable                     = local.disable_extras
       kubelet-arg                 = ["cloud-provider=external", "volume-plugin-dir=/var/lib/kubelet/volumeplugins"]
       kube-controller-manager-arg = "flex-volume-plugin-dir=/var/lib/kubelet/volumeplugins"
+      flannel-iface               = "eth1"
       node-ip                     = module.control_planes[keys(module.control_planes)[0]].private_ipv4_address
       advertise-address           = module.control_planes[keys(module.control_planes)[0]].private_ipv4_address
       node-taint                  = local.control_plane_nodes[keys(module.control_planes)[0]].taints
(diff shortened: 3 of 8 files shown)


mysticaltech:
Hey @PurpleBooth, it's very important to have  `flannel-iface  = "eth1"` as it forces the cluster to only function on the Hetzner private network, as this is what we want. Don't worry about the name, it does not just apply to flannel, but all CNIs :)

For instance here with this set to "none" as was done, Cilium is considering both interface, which is not ideal. Have fixed it, along with a few other proposals. Pushing soon.
ksnip_20220802-033252

cilium_values.yaml.example
```diff
@@ -0,0 +1,30 @@
+# All Cilium helm values can be found at https://github.com/cilium/cilium/blob/master/install/kubernetes/cilium/values.yaml
```
mysticaltech, line 1: @PurpleBooth This file is just an example of helm values that can be configured to seek the desired result. Here, for instance, we use Cilium for KubeProxy functionality.

In the past, I could get host routing via BPF without the Hetzner CSI crashing but couldn't reproduce it. As if relies on IPtables or something.

mysticaltech: So for now, no BPF host routing unless one would use Longhorn, for instance. Then there are a few parameters that would enable it. But not needed unless an app is huge with a crazy amount of hits that iptables can't handle.

```
  bpf:
    masquerade: true
```

control_planes.tf
```diff
@@ -96,13 +96,17 @@ resource "null_resource" "control_planes" {
       disable                     = local.disable_extras
       kubelet-arg                 = ["cloud-provider=external", "volume-plugin-dir=/var/lib/kubelet/volumeplugins"]
       kube-controller-manager-arg = "flex-volume-plugin-dir=/var/lib/kubelet/volumeplugins"
+      flannel-iface               = "eth1"
```
mysticaltech, line 99: I saw you removed these, but they are very important because otherwise, the CNI uses all interfaces, including the public one, which is not needed or wanted. I know it's supposed to be for flannel, but best to set it all the time (as not it configures Kube underneath, I guess).

control_planes.tf
```diff
@@ -96,13 +96,17 @@ resource "null_resource" "control_planes" {
...
       node-ip                     = module.control_planes[each.key].private_ipv4_address
       advertise-address           = module.control_planes[each.key].private_ipv4_address
       node-label                  = each.value.labels
       node-taint                  = each.value.taints
       write-kubeconfig-mode       = "0644" # needed for import into rancher
       },
       lookup(local.cni_k3s_settings, var.cni_plugin, {}),
+      var.use_control_plane_lb ? {
```
mysticaltech, line 107: This is important to have in all control planes, whatever init has, control_planes must have too, as both become interchangeable.

init.tf
```diff
@@ -161,7 +162,7 @@ resource "null_resource" "kustomization" {
     content = templatefile(
       "${path.module}/templates/cilium.yaml.tpl",
       {
-        cluster_cidr_ipv4 = local.cluster_cidr_ipv4
+        values = indent(4, trimspace(fileexists("cilium_values.yaml") ? file("cilium_values.yaml") : local.default_cilium_values))
```
mysticaltech, line 165: Either use the default config, or the helm values read from the optional `cilium_values.yaml` one can now add next to `kube.tf`.

locals.tf
```diff
@@ -255,17 +255,24 @@ locals {
...
+      flannel-backend        = "none"
     }
   }
+
+  default_cilium_values = <<EOT
+ipam:
+ operator:
+  clusterPoolIPv4PodCIDRList:
+   - ${local.cluster_cidr_ipv4}
+devices: "eth1"
```
mysticaltech, line 275: Important for Cilium to stick to the private net :)

PurpleBooth: oooh, nice! I never found this option and ended up doing horrible in my own setup

control_planes.tf
PurpleBooth: I wonder if we should just make a local for all these settings

control_planes.tf
PurpleBooth: TIL

control_planes.tf
mysticaltech: Yeah, not very DRY, huh 😅  This thought crossed my mind a lot! Now, I saw you used `lookup`, which could work in these situations, maybe?!

mysticaltech:
So whenever you test and all is good on your side, we can merge and release 1.4.4!

PurpleBooth:
Not quite sure what's happened but this doesn't seem to work for me, looking at the first control plane the helm chart resource is never created for it

```hcl
module "kube-hetzner" {
  providers = {
    hcloud = hcloud
  }
  source = "../terraform-hcloud-kube-hetzner"
  ssh_public_key = file("~/.ssh/id_rsa.pub")
  ssh_private_key = file("~/.ssh/id_rsa")
  control_plane_nodepools = [
    {
      name        = "control-plane",
      server_type = "cx21",
      location    = "fsn1",
      labels      = [],
      taints      = [],
      count       = 2
    }
  ]
  agent_nodepools = [
    {
      name        = "agent",
      server_type = "cx21",
      location    = "fsn1",
      labels      = [],
      taints      = [],
      count       = 2
    }
  ]
  cni_plugin = "cilium"
  use_control_plane_lb = true
  dns_servers          = []
}
```

mysticaltech:
What's the error? With k9s it's super easy to dive down the helm cilium chart installer pod and grab the logs. That will tell you what went wrong. I am suspecting the helm values in the file if present.

You have two use cases:

- Without the cilium_config.yaml file, you will use defaults which work as the same settings we had before. 

- And with that will be the path chosen if the file is found, and in it, you need to have the helm values or your choosing and must know what you are doing, as not all options would work together and out of the box. So best to start with the `cilium_values.yaml.example`. 

In which case were you above?

mysticaltech:
Hey @PurpleBooth, I would like to release this before the week's end! It's too good to keep on the back burner. So please give it a shot again whenever you can (see explanations above) 🙏 

The default mode (when no extra file is added) works and uses almost the same settings as you first proposed, apart from the addition of `devices`, which is important.

If you don't have the bandwidth now, we can just release and wait to see if folks open issues or not, and then start climbing the optimization mountain! ;)

PurpleBooth:
Unfortunately I am out of contact from a laptop the next few days. Will have a look after that though!

mysticaltech:
@PurpleBooth No problem, I will give it one more test, and if all is good, merge and release. Then you can test whenever you can, and if further fixes/enhancements are needed from your end, we can push a patch quickly.

mysticaltech:
Everything checks out again, this time with no `cilium_config.yaml` file (yesterday did multiple tests with it).

`Kube.tf` is as follows:

```
locals {
  hcloud_token = "xxxxx"
}

module "kube-hetzner" {
  providers = {
    hcloud = hcloud
  }

  source = "../kube-hetzner/"

  ssh_public_key  = file("/home/xxxxx/.ssh/id_ed25519.pub")
  ssh_private_key = file("/home/xxxxx/.ssh/id_ed25519")

  network_region = "eu-central"

  control_plane_nodepools = [
    {
      name        = "control-plane",
      server_type = "cx21",
      location    = "nbg1",
      labels      = [],
      taints      = [],
      count       = 2
    }
  ]

  agent_nodepools = [
    {
      name        = "agent",
      server_type = "cx21",
      location    = "nbg1",
      labels      = [],
      taints      = [],
      count       = 2
    }
  ]


  cluster_name = "test12"

  extra_firewall_rules = [
    # For Postgres
    {
      direction       = "in"
      protocol        = "tcp"
      port            = "5432"
      destination_ips = [] # Won't be used for this rule 
    },
    # To Allow ArgoCD access to resources via SSH
    {
      direction       = "out"
      protocol        = "tcp"
      port            = "22"
    }
  ]

  cni_plugin  = "cilium"
  base_domain = "test12.mystical.com"
}

provider "hcloud" {
  token = local.hcloud_token
}

terraform {
  required_providers {
    hcloud = {
      version = ">= 1.0.0"
    }
  }
}
```

ksnip_20220803-184010

mysticaltech:
Just released in 1.4.4 adding Cilium support. Thanks for the great work @PurpleBooth! 🥳 And rest well if you are on vacation!

mysticaltech:
@phaer @PurpleBooth We are now mentioned on the Cilium [USERS.md](https://github.com/cilium/cilium/blob/master/USERS.md) page! 😍

t1murl:
Hi guys, great work implementing Cilium as a CNI option in this awesome project! 🙏 
While testing out Cilium with this project I've noticed that after node reboots through e.g. kured many pods become unmanaged by Cilium and stop working correctly as a result.
Would we want to set a Cilium specific taint per default when CNI = cilium as mentioned in the [Cilium docs](https://docs.cilium.io/en/v1.12/gettingstarted/taints/#taint-effects) to prevent the aforementioned from happening?
I'll have time to try out the mentioned taint later today or tomorrow and am happy to submit a short PR for you to review if everything works.
Just wanted to reach out and give a heads up 👍 
FYI: @mysticaltech , @phaer , @PurpleBooth

t1murl:
FYI: Simply adding the taint "node.cilium.io/agent-not-ready=true:NoExecute" to the nodepools section obviously does not play well with how the kustomizations (and Cilium) are bootstrapped. I'm thinking something like we would need to boostrap Cilium first, then wait for Cilium to remove the taint once fully initialized before applying the other kustomizations.
Maybe strike the '**short** PR' from my previous message 😆 
Any input / opinion on this is more than welcome 👍

mysticaltech:
@t1murl Thanks for finding and taking this on! What you said last seems like our best shot, first Cilium, wait for it to finish then the rest! I'm sure it's possible to test for label repetitively. Can't wait to review your PR 🙏

phaer:
> I'm thinking something like we would need to boostrap Cilium first, then wait for Cilium to remove the taint once fully initialized before applying the other kustomizations (if we want to keep the current implementation of waiting for certain resources to become ready).

Maybe this could be done with `kubectl wait`? Or if not, a shell loop + curl call to the kubernetes API?

mysticaltech:
@t1murl Have you been able to do any progress? If you have started working on this, please do push a draft PR, and I will do my best to squeeze in some time into this, as it's kind of urgent; I would like to fix this ASAP. Thanks! 🙏

mysticaltech:
If not, sharing your ideal flow about how this should go to succeed (the label logic, I guess) would help a lot!

t1murl:
Find my reply in #270 .

PR #260, opened 2022-08-01
Remove deprecated taint

PurpleBooth:
This taint should match the new role name "control-plane"

Kubeadm [will remove this with 1.25][issue]. It's a bit of an awkward timing, because that looks like it will come out at the [end of this month](https://github.com/kubernetes/sig-release/tree/master/releases/release-1.25#summary)

[issue]: https://github.com/kubernetes/kubeadm/issues/2200


mysticaltech:
Good catch!

Issue #241, opened 2022-07-09
Features to use only private IP


jodhi:
Hello, I just saw the news on [hetzner page regarding flexible networking](https://www.hetzner.com/news/06-22-flexibles-networking/)

And since the kube-hetzner already using private ip by default, i think that should be straightforward to implement worker nodes only with private ip (still not sure).
What do you guys think about worker nodes without public ip?

ps: I can try to create PR if that make sense

phaer:
Good idea, PR more then welcome I think!

jodhi:
cool, i will try this one

ricristian:
And idea would be to extend this from worker nodes also to control plane ones due to the fact that there might be other possibilities to connect to control planes maybe via a VPN and so on . In this way we can basically close completely any chance from receiving requests from outside ...

jodhi:
> An idea would be to extend this from worker nodes also to control plane ones due to the fact that there might be other possibilities to connect to control planes maybe via a VPN and so on . In this way we can basically close completely any chance from receiving requests from outside ...

Yeah, make sense as well, this one will require bastion/jumphost to access the control plane right.
Now you mention it, we need to access the worker node to bootstrap them as well 😓. Looks like we need internal bastion/jumphost to bootstrap the node  🤔

ricristian:
Or at initial bootstrap of the cluster we use public IPv4 and after we finish the job drop those public IP's ?

jodhi:
@ricristian that is smart idea. really

ricristian:
I'm not 100% sure how terraform works but I don't think that we can run in a sequence manner tf modules so a 2nd step can be implemented. I don't see that much hassle for that instead of trying to use jump hosts and so on ...

jodhi:
yeah right, looks like terraform cannot create machine with ip, provision, then destroy it with one go.

what i have in mind 
1. 2-step
```
we put  the machine as usual with public ip in tfvars
terraform apply
then we change the tfvars to remove public ip
terraform apply
```

2. 1-step
```
create control plane nodes with public ip, then create worker nodes without public ip, then get inside the worker nodes through control planes nodes
```

3. 1-step
Use  jumphost, then the terraform module can access all nodes internally. [some docs](https://dev.to/jaysonsantos/connecting-to-services-that-require-jumphost-from-terraform-1aog)

cmiiw. feel free to add

ricristian:
[...]

PR #261, opened 2022-08-02
Setup bastion host

PurpleBooth:
Currently this does not work as we have no way to get rescue system in place so we can install MicroOS


mysticaltech:
Beautiful code right there; I am sure @phaer will be inspired by it. There must be a way to make this work, like setup with the public interface, then just take it down, all via userdata, or remote-exec (unless am not seeing things clearly).

bastion.tf
```diff
@@ -0,0 +1,38 @@
+module "bastion" {
```
mysticaltech, line 1: @PurpleBooth What is the role of this module?

modules/host/main.tf
```diff
@@ -30,6 +30,16 @@ resource "hcloud_server" "server" {
   placement_group_id = var.placement_group_id
   user_data          = data.cloudinit_config.config.rendered
 
+  network {
+    ip         = var.private_ipv4
+    network_id = var.ipv4_subnet.network_id
+  }
+
+  public_net {
```
mysticaltech, line 38: Won't work that way, but not an issue. We can deploy a node and kill its public interface and access it through the private network.

mysticaltech: But the node needs to be deployed as explained at other places via bootstrapping through ignition/combusion. Very smart mechanism there, we created partition, and it auto-detected the scripts to execute before-boot, to enable or kill instances for that matter :)

mysticaltech:
The beauty is that this project has solved a lot of similar scenarios, and has shared a few [links](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/tree/8634a1dc2fd5b05872cf2b35a5aae219ac691282) already to a past state in git, which should hopefully inspire you. The whole thing is there, the image, how to create the partitions, load the script, the reboot logic. With this, you just change a few lines to disable the interface and get your fully isolated agent nodes!

mysticaltech:
Something important though, why? Do that? Why do we need to setup a bastion host. It's probably completely doable, but what benefit will we gain from it as a project, is I understand the need for agent nodes to be private and isolated, is the reason why? And in that case, why not just kill the public interface after everything is done with a remote exec? An attack during the terraform install is highly unlikely.

modules/host/main.tf
mysticaltech: Interesting article started in the private ip only agent nodes task https://www.hetzner.com/news/06-22-flexibles-networking/. 
ksnip_20220802-164508

mysticaltech:
@PurpleBooth As shared here, if possible, we can do private agents more simply, with bare Kubernetes, without having an additional bastion host. See my comment here https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/241#issuecomment-1203991195. 

The simpler, the better! But definitely, I would love to hear your arguments if you don't agree.

modules/host/versions.tf
```diff
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     hcloud = {
       source  = "hetznercloud/hcloud"
-      version = ">= 1.33.2"
+      version = ">= 1.34.2"
```
4ND3R50N, line 5: Hey 👍 There are some bugs in this version related to private network only server!

I highly recommend updating the version to the latest one.

Please have a look here:
https://github.com/hetznercloud/terraform-provider-hcloud/releases/tag/v1.35.1

mysticaltech: Thanks, @4ND3R50N; please don't hesitate to PR so we'll bump that version globally. Otherwise, I'll do it ASAP.

mysticaltech:
@PurpleBooth While merging the latest changes, I noticed that you were probably assuming that `rebootmgr` takes care of all reboots including dealing with `kured`. This is not the case, in order to disable `kured`, you have to add a `REBOOT_METHOD=none` to the transactional_update.conf file, located somewhere in /etc.

More on this here https://en.opensuse.org/Kubic:Update_and_Reboot#Reboot_Strategy_Options and https://github.com/openSUSE/transactional-update/blob/master/etc/transactional-update.conf.

mysticaltech:
Now I understand better what this is, thanks to your topology description. It also made me realize that node upgrades, k3s upgrades, and container fetching would disappear, which is not really ideal. Or am I mistaken?

mysticaltech:
Ah, they do have a gateway that the private subnet can route traffic to in all likelihood. So the nodes would not be cut out from OUT traffic (but needs to be tested, of course).

mysticaltech:
@PurpleBooth Interesting article https://docs.hetzner.com/cloud/servers/getting-started/connecting-via-private-ip/

mysticaltech:
I believe the next step here is to use snapshots made with "normal" nodes. So yes, the bastion setup will take longer, but that's a small initial price to pay for the added sec.

mysticaltech:
Just stumbled on this, seems interesting, but probably not needed (at least for now) https://github.com/inlets/inlets-pro

mysticaltech:
@PurpleBooth I will close this PR for now, because it has stalled and the base is significantly outdated. But the branch will remain, so it can be picked up again later without issue.

Issue #277, opened 2022-08-18
Use different for SSH


otavio:
We would like to provide a service which uses SSH port so we would like to use an alternate port for it.

mysticaltech:
Sounds useful, PR welcome @otavio.

otavio:
No idea where to start. Any pointers?

mysticaltech:
The default SSH port for nodes needs to be conditionally changed at the cloud-init level, so editing the cloud-init template and the var file should be all that is needed, along with a small explanation in the kube.tf.example.

PR #279, opened 2022-08-23
Allow setting alternative SSH port for host access

otavio:
The use of alternative SSH port is useful for different use cases or
when we need to provide the port for other service.

Fixes: #277.
Signed-off-by: Otavio Salvador <otavio@ossystems.com.br>

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ _Once you start with Terraform, it's best not to change the state manually in He
 
 ## CNI
 
-The default is flannel, but you can also choose Calico or Cilium, by setting the cni_plugin variable in `kube.tf` to `calico` or `cilium`. For both, you can create a HelmChartConfig to edit their respective configuration post cluster deploy!
+The default is flannel, but you can also choose Calico or Cilium, by setting the cni_plugin variable in `kube.tf` to `calico` or `cilium`.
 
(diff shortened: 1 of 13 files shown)


mysticaltech:
Looking good @otavio, thanks for this, will try ASAP and merge it.

mysticaltech:
@otavio I tried it, and it works with the default port. However, when I tried setting a ssh_port of 2222, it could not connect to the hosts. Please see the screenshot below. Please try it on your end. Probably a minor fix is needed. Awaiting a subsequent PR, thanks.

ksnip_20220826-140114

mysticaltech:
@otavio @phaer What do you think of having the default SSH port not be 22 and 2222 instead? 

I see no harm in it and it would simplify the life of folks wanting to use 22 in a subsequent firewall tweak via a second `terraform apply`.

otavio:
This is failing at:
https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/279/commits/b44417a2b6ba648ee547700b20ee404b456a8f7d#diff-74e90ce3d374ada5572c1bdaa5f631c6df60251c74d4fdbe0afb372fd9fe79b8R63-R72
However, it is strange as it starts in rescue using port 22, and once booting inside MicroOS, it uses the assigned port.

otavio:
> @otavio @phaer What do you think of having the default SSH port not be 22 and 2222 instead?
> 
> I see no harm in it, and it would simplify the life of folks wanting to use 22 in a subsequent firewall tweak via a second `terraform apply` (or the initial one, for that matter).

I want to use 60022 and 2222 is too common and we avoid it as well.

phaer:
@mysticaltech I'd prefer to stay at 22 as a default because it's well established and the possible space for alternative ports is pretty big, e.g. @otavio wanting to use 60022.

@otavio Why is that strange? The rescue system is hetzners debian, it's got nothing to do with microos?

otavio:
> @mysticaltech I'd prefer to stay at 22 as a default because it's well established and the possible space for alternative ports is pretty big, e.g. @otavio wanting to use 60022.

Agreed.
 
> @otavio Why is that strange? The rescue system is hetzners debian, it's got nothing to do with microos?

The error for @mysticaltech as it worked for me. Can you check at your side? I opened #281 so we can center discussion on the new PR.

mysticaltech:
Ok @phaer, let's stay with a default of port 22. 

@otavio I use a standard kube.tf, will share it in the new PR and try to investigate more also.

PR #281, opened 2022-08-26
Allow setting alternative SSH port for host access

otavio:
The use of alternative SSH port is useful for different use cases or
when we need to provide the port for other service.

Fixes: #277.
Refs: #279.

### DIFF
--- a/modules/host/templates/userdata.yaml.tpl
+++ b/modules/host/templates/userdata.yaml.tpl
@@ -10,6 +10,7 @@ write_files:
 
 # Disable ssh password authentication
 - content: |
+    Port ${sshPort}
     PasswordAuthentication no
     X11Forwarding no
     MaxAuthTries 2
--- a/modules/host/templates/userdata.yaml.tpl
+++ b/modules/host/templates/userdata.yaml.tpl
@@ -51,6 +51,8 @@ preserve_hostname: true
 
 runcmd:
 
+- [sed, '-i', '/Port /d', /etc/ssh/sshd_config]
+
 # As above, make sure the hostname is not reset
 - [sed, '-i', 's/NETCONFIG_NIS_SETDOMAINNAME="yes"/NETCONFIG_NIS_SETDOMAINNAME="no"/g', /etc/sysconfig/network/config]
 - [sed, '-i', 's/DHCLIENT_SET_HOSTNAME="yes"/DHCLIENT_SET_HOSTNAME="no"/g', /etc/sysconfig/network/dhcp]
--- a/modules/host/variables.tf
+++ b/modules/host/variables.tf
@@ -8,6 +8,16 @@ variable "base_domain" {
   type        = string
 }
 
+variable "ssh_port" {
+  description = "SSH port"
+  type        = number
+
+  validation {
+    condition     = var.ssh_port >= 0 && var.ssh_port <= 65535
+    error_message = "The ssh_port must use a valid range from 0 to 65535.."
+  }
+}
+
 variable "ssh_public_key" {
   description = "SSH public Key"
   type        = string
--- a/variables.tf
+++ b/variables.tf
@@ -4,6 +4,17 @@ variable "hcloud_token" {
   sensitive   = true
 }
 
+variable "ssh_port" {
+  description = "SSH port."
+  type        = number
+  default     = 22
+
+  validation {
+    condition     = var.ssh_port >= 0 && var.ssh_port <= 65535
+    error_message = "The ssh_port must use a valid range from 0 to 65535.."
+  }
+}
+
 variable "ssh_public_key" {
   description = "SSH public Key."
   type        = string
--- a/agents.tf
+++ b/agents.tf
@@ -10,6 +10,7 @@ module "agents" {
   name                       = "${var.use_cluster_name_in_node_name ? "${var.cluster_name}-" : ""}${each.value.nodepool_name}"
   base_domain                = var.base_domain
   ssh_keys                   = [local.hcloud_ssh_key_id]
+  ssh_port                   = var.ssh_port
   ssh_public_key             = var.ssh_public_key
   ssh_private_key            = var.ssh_private_key
   ssh_additional_public_keys = var.ssh_additional_public_keys
@@ -42,6 +43,7 @@ resource "null_resource" "agents" {
     private_key    = var.ssh_private_key
     agent_identity = local.ssh_agent_identity
     host           = module.agents[each.key].ipv4_address
+    port           = var.ssh_port
   }
 
   # Generating k3s agent config file
--- a/control_planes.tf
+++ b/control_planes.tf
@@ -10,6 +10,7 @@ module "control_planes" {
   name                       = "${var.use_cluster_name_in_node_name ? "${var.cluster_name}-" : ""}${each.value.nodepool_name}"
   base_domain                = var.base_domain
   ssh_keys                   = [local.hcloud_ssh_key_id]
+  ssh_port                   = var.ssh_port
   ssh_public_key             = var.ssh_public_key
   ssh_private_key            = var.ssh_private_key
   ssh_additional_public_keys = var.ssh_additional_public_keys
@@ -80,6 +81,7 @@ resource "null_resource" "control_planes" {
     private_key    = var.ssh_private_key
     agent_identity = local.ssh_agent_identity
     host           = module.control_planes[each.key].ipv4_address
+    port           = var.ssh_port
   }
 
   # Generating k3s server config file
--- a/init.tf
+++ b/init.tf
@@ -4,6 +4,7 @@ resource "null_resource" "first_control_plane" {
     private_key    = var.ssh_private_key
     agent_identity = local.ssh_agent_identity
     host           = module.control_planes[keys(module.control_planes)[0]].ipv4_address
+    port           = var.ssh_port
   }
 
   # Generating k3s master config file
@@ -81,6 +82,7 @@ resource "null_resource" "kustomization" {
     private_key    = var.ssh_private_key
     agent_identity = local.ssh_agent_identity
     host           = module.control_planes[keys(module.control_planes)[0]].ipv4_address
+    port           = var.ssh_port
   }
 
   # Upload kustomization.yaml, containing Hetzner CSI & CSM, as well as kured.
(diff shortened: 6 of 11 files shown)


modules/host/templates/userdata.yaml.tpl
```diff
@@ -50,6 +51,8 @@ preserve_hostname: true
 
 runcmd:
 
+- [sed, '-i', '/Port /d', /etc/ssh/sshd_config]
```
mysticaltech, line 54: @otavio What is this line for? Isn't the above enough?

modules/host/variables.tf
```diff
@@ -8,6 +8,11 @@ variable "base_domain" {
   type        = string
 }
 
+variable "ssh_port" {
+  description = "SSH port"
```
mysticaltech, line 12: It would be great to change this type to a number instead of a string and add a condition check to verify it's within the allowed range.

variables.tf
```diff
@@ -4,6 +4,12 @@ variable "hcloud_token" {
   sensitive   = true
 }
 
+variable "ssh_port" {
+  description = "SSH port."
+  type        = string
+  default     = 22
+}
```
mysticaltech, line 11: Same as above, 
It would be great to change this type to a number instead of a string and add a condition check to verify it's within the allowed range.

mysticaltech:
@otavio Please could you give it a try with the following `kube.tf` if it works for you, then maybe it's because we are using different OSes; I'm on Fedora.
```
locals {
  hcloud_token = "xxxxxx"
}

module "kube-hetzner" {
  providers = {
    hcloud = hcloud
  }

  source = "../kube-hetzner/"

  network_region = "eu-central"

  control_plane_nodepools = [
    {
      name        = "control-plane",
      server_type = "cx21",
      location    = "nbg1",
      labels      = [],
      taints      = [],
      count       = 2
    }
  ]

  agent_nodepools = [
    {
      name        = "agent",
      server_type = "cx21",
      location    = "nbg1",
      labels      = [],
      taints      = [],
      count       = 2
    }
  ]


  cluster_name = "test12"

  extra_firewall_rules = [
    # For Postgres
    {
      direction       = "in"
      protocol        = "tcp"
      port            = "5432"
      destination_ips = [] # Won't be used for this rule 
    },
    # To Allow ArgoCD access to resources via SSH
    {
      direction       = "out"
      protocol        = "tcp"
      port            = "22"
    }
  ]

  ssh_port                          = 2222
}

provider "hcloud" {
  token = local.hcloud_token
}

terraform {
  required_providers {
    hcloud = {
      version = ">= 1.0.0"
    }
  }
}
```

mysticaltech:
This is how it fails on my end; no ports are mentioned!?
ksnip_20220826-175013

mysticaltech:
@otavio Please note that from reading the [provisioner docs](https://www.terraform.io/language/resources/provisioners/connection), it expects a number and not a string. And this behavior maybe is tolerated on some OSes like yours, so best to change to number types across the board and do a cast to string whenever needed.

modules/host/templates/userdata.yaml.tpl
otavio: I'm redoing the tests without this. I'll push the removal once I confirm this works.

otavio:
Ok, I reproduced the error here. I am preparing the fix for it.

otavio:
I fixed all the above comments and the deploy error; I did fixup commits, so it is easier to review. Once it is Ok, I squash them and mark them for final review.

modules/host/variables.tf
```diff
@@ -10,7 +10,12 @@ variable "base_domain" {
 
 variable "ssh_port" {
   description = "SSH port"
-  type        = string
+  type        = number
+
+  validation {
```
mysticaltech, line 15: Awesome!

variables.tf
```diff
@@ -6,8 +6,13 @@ variable "hcloud_token" {
 
 variable "ssh_port" {
   description = "SSH port."
-  type        = string
+  type        = number
```
mysticaltech, line 9: Yes Sir!

mysticaltech:
@otavio I just tried now, and the first part connects correctly via port 22, but it still doesn't respond for the final stages. I think maybe you put the "sed" for a reason?

ksnip_20220826-190833

Please let me know when it works on your end for final testing! 🙏

otavio:
Here it worked. See:
```console
...
module.infra.null_resource.kustomization: Still creating... [40s elapsed]
module.infra.null_resource.kustomization (remote-exec): deployment.apps/system-upgrade-controller condition met
module.infra.null_resource.kustomization (remote-exec): + sleep 5
module.infra.null_resource.kustomization (remote-exec): + kubectl -n system-upgrade apply -f /var/post_install/plans.yaml
module.infra.null_resource.kustomization (remote-exec): plan.upgrade.cattle.io/k3s-agent created
module.infra.null_resource.kustomization (remote-exec): plan.upgrade.cattle.io/k3s-server created
module.infra.null_resource.kustomization (remote-exec): + timeout 120 bash
module.infra.null_resource.kustomization (remote-exec): Waiting for load-balancer to get an IP...
module.infra.null_resource.kustomization (remote-exec): Waiting for load-balancer to get an IP...
module.infra.null_resource.kustomization: Still creating... [50s elapsed]
module.infra.null_resource.kustomization (remote-exec): Waiting for load-balancer to get an IP...
module.infra.null_resource.kustomization: Creation complete after 55s [id=1285535057785919446]
module.infra.data.hcloud_load_balancer.traefik[0]: Reading...
module.infra.data.remote_file.kustomization_backup: Reading...
module.infra.null_resource.destroy_traefik_loadbalancer: Creating...
module.infra.null_resource.destroy_traefik_loadbalancer: Creation complete after 0s [id=395963813960186084]
module.infra.data.hcloud_load_balancer.traefik[0]: Read complete after 1s [name=lab-traefik]
digitalocean_record.lb: Creating...
digitalocean_record.lb: Creation complete after 1s [id=333216929]
module.infra.data.remote_file.kustomization_backup: Read complete after 4s [id=5.161.152.121:60022:/var/post_install/kustomization.yaml]
module.infra.local_file.kustomization_backup: Creating...
module.infra.local_file.kustomization_backup: Creation complete after 0s [id=6374b71a7f689dbb3108a2a1f49c6f2aab37eff1]

Apply complete! Resources: 31 added, 0 changed, 0 destroyed.
```

mysticaltech:
This is strange. On my end, it still can't connect to MicroOS. I used the Hetzner console to further debug, and here's what is happening. The OpenSSH server is not starting after cloud-init, so the ssh_config file is probably left in an invalid state. I will try to investigate more.

ksnip_20220826-233743

otavio:
Can you check if sed helps? added it in fixup.

mysticaltech:
@otavio I tried myself and retried now with your change. Same, unfortunately. If I remember correctly, cloud-init has a debug mode that prints more info; going to give this a shot.

mysticaltech:
Trying with the following
ksnip_20220827-005611
 will post the output.

mysticaltech:
@otavio I added this command to the runcmd part `- [journalctl, '-u', sshd]` and we now know the error, it cannot bind to the port. After googling, it turns out it could be SELinux blocking it. Now trying the fix.

ksnip_20220827-011912

mysticaltech:
@otavio Here's the situation from my point of view, I have tried hitting it from many angles, and the solution seems to remain to execute this command `- [semanage, port, '-a', '-t', ssh_port_t, '-p', tcp, '${sshPort}']`. 

However, there is a significant issue, `semanage` is not installed by default on MicroOS images, `setsebool` is, so I tried many flags but with no success. The only solution is to install the missing `semanage` command, which can only be installed with `transactional-updates`, and when this is done, it does not work because it needs to reboot for the snapshot to come into place. And executing `zypper` also does not work.

This means that the only solution that I see from my limited vantage point at this time is to switch back to a MicroOS image that supports [combustion](https://en.opensuse.org/Portal:MicroOS/Combustion),  basically the execution of a shell script before boot, and this allows for plain `zypper` to be used. We were using this before moving to a cloud-init image. But with this we would be dropping the advantages of cloud-init, and the possibility to quickly adopt an eventual Hetzner own MicroOS image in the future if they support them.

@sysrich Your guidance would be very much appreciated; basically, we are just trying to modify the default SSH port, but SELinux is blocking it, and since `semanage` is not installed by default and cannot be installed, then used within cloud-init (unless I am mistaken) we are stuck. Do you see any way we could have an image with `semanage` installed or another way around this without dropping our cloud-init setup?

mysticaltech:
@otavio Maybe we could create a Hetzner snapshot of a base working MicroOS image with the SSH default port of 22 and `semanage` installed on it. And then deploy that snapshot, everywhere. Creating a snapshot was already in the pipes to allow for autoscaling. But it just complicates things, and if we can find a more direct way, it would be best for the time being.

mysticaltech:
I have at least confirmed manually that two command do the fix, the `semanage` one, and the `semodule -i` module installation after creating a module via `audit2allow -a -M sshd_t` after causing the issue purposefully on an accessible machine.

The latter's advantage is that it allows you to move the binary file to a different machine. In that branch https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/commit/d239ce1c2a25c8f4dd51d94d87daaff247cb5c3e I attempted writing via base64 !!binary file, and applying it with `semodule`. 

I can confirm that the files seem created correctly, but I cannot yet confirm that the semodule applies correctly, and I still have no success via Terraform. 

Am going to pause on this for now, unless I find more inspiration.

mysticaltech:
Confirmed that the `sshd_t.pp` module file is correctly recreated on the node via `cloud-init`, and works when applied with `semodule -vi /etc/semodule/sshd_t.pp`.

mysticaltech:
It worked, finally! Just needed to restart `sshd` from within cloud-init.

mysticaltech:
Merging to staging, where I will add the final tweaks.

mysticaltech:
@otavio It has been released in v1.4.5; thank you for the excellent contribution! This last phase was engaging with the SELinux issue, but it was good learning. 

Also, I informed Richard Brown of the need to have essential tools like `semanage` shipped with the base OS. Hopefully, he'll consider the suggestion!

otavio:
@mysticaltech thank you for doing this debug. I am wondering why it worked for me ... glad the PR ended succeeding.

mysticaltech:
Me too, like really have no idea at all how it worked on your end 🤯 

The 'semodule port -l' command clearly showed enforcement of SELinux of port 22 by default.

Maybe you tested with an old branch or something without realizing it!? 😅

 But anyways, we figured it out finally! :)

Issue #293, opened 2022-09-01
Expose loadbalancer IPv4 when Traefik disabled.


otavio:
### Discussed in https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/discussions/284

<div type='discussions-op-text'>

<sup>Originally posted by **geraldwuhoo** July  6, 2022</sup>
Currently the code in `output.tf` returns `null` for `load_balancer_public_ipv4` if Traefik is disabled. I can understand the reasoning, because without an ingress controller to provision an IP from the Hetzner LB, there is no LB IP to speak of. However, I'm disabling Traefik to install nginx-ingress instead. I'm sure many others disable Traefik to instead install their own preferred ingress controller.

Since there is no way for the Terraform module to detect whether or not a user will provision an ingress controller *after* Terraform provisioning, I think it makes sense to err on the side of caution, and return proper `load_balancer_public_ipv4` value even when Traefik is disabled.</div>

mysticaltech:
Just FYI future readers, with the new system for managing ingresses in #298, adding new ingresses via HelmChart config is now more or less a piece of cake. So don't hesitate to send us your PRs!

Kong, HAProxy, Caddy... You-name-it Ingress Controller! If it has both a decent user base and a helm chart, we want it! And we can integrate it quite easily, just follow in the footsteps of Nginx-ingress. Issuing a Find "nginx" in your code editor will illuminate the path you need to take to add the Ingress controller of your liking.

mysticaltech:
@geraldwuhoo FYI, the above. We know support Traefik, Nginx, and none - And we also have an easy flow to add more Ingress controllers if someone needs them, and they are popular in the cloud-native community to justify supporting them. This should solve your original issue #254, indirectly at least.

PR #294, opened 2022-09-01
Allow use of load balancer even if traefik is disabled

otavio:
Fixes: #293.
Signed-off-by: Otavio Salvador <otavio@ossystems.com.br>

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -60,14 +60,14 @@ resource "hcloud_placement_group" "agent" {
   type  = "spread"
 }
 
-data "hcloud_load_balancer" "traefik" {
+data "hcloud_load_balancer" "cluster" {
   count = local.using_klipper_lb ? 0 : var.traefik_enabled == false ? 0 : 1
-  name  = "${var.cluster_name}-traefik"
+  name  = "${var.cluster_name}-lb"
 
   depends_on = [null_resource.kustomization]
 }
 
-resource "null_resource" "destroy_traefik_loadbalancer" {
+resource "null_resource" "destroy_cluster_loadbalancer" {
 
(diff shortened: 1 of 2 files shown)


mysticaltech:
Great initiative @otavio, we just need to continue using klipper, that gets activated automatically for single node clusters for instance.

otavio:
Do we need klipper in this case?

otavio:
I wonder as traefik is [creating the load balancer](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/master/templates/traefik_config.yaml.tpl#L13) and without it, we would need to make it externally as otherwise, we wouldn't know the IP of it.

mysticaltech:
We need klipper; it's an on-metal lb, which gives all the nodes IPs, not just one. Traffic in the current setup can be used either with the Hetzner LB, or with Klipper. So we would need to maintain klipper functionality.

mysticaltech:
I haven't checked, but normally anything that requests an LB can be used with klipper, but maintaining it tied to traefik as it is now is ok.

mysticaltech:
It's not a lot @otavio, the way I see it, just re-add the checks with `local.using_klipper_lb` that you removed and it should be good to merge.

otavio:
@mysticaltech I did as you asked but it is kinda confusing for me now. How I'd use the load balancer IP this way?

mysticaltech:
@otavio If that seems ok but a lot, you can share your Nginx config, and I will try to make the changes.

mysticaltech:
@otavio I'm taking this on, will merge until here, as it's a valid naming change, and continue adding Nginx on a separate branch. I will tag you in the next PR.

PR #298, opened 2022-09-03
Add Nginx ingress controller + various tweaks | ⚠️ Breaking changes in some variable names

### LINKED ISSUE #293: Expose loadbalancer IPv4 when Traefik disabled.

mysticaltech:
This PR adds the Nginx ingress controller and the possibility to use a helm `_values.yaml` file for both it and Rancher. It also tweaks the variable names to make them more consistent.

Last but not least, a lot of the logic related to load-balancers, including Klipper, has been tweaked and cleaned up!

This fixes #293.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
   <h2 align="center">Kube-Hetzner</h2>
 
   <p align="center">
-    A highly optimized and auto-upgradable, HA-default & Load-Balanced, Kubernetes cluster powered by k3s-on-MicroOS and deployed for peanuts on <a href="https://hetzner.com" target="_blank">Hetzner Cloud</a> 🤑 🚀
+    An optimal, easy-to-use, auto-upgradable, HA-default & Load-Balanced, Kubernetes cluster powered by k3s on MicroOS and deployed for peanuts on <a href="https://hetzner.com" target="_blank">Hetzner Cloud</a> 🤑 🚀
   </p>
   <hr />
 </p>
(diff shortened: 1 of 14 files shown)


mysticaltech:
@otavio This turned out to be a big PR with various tweaks, and because of some breaking changes in the variable names, was the occasion to push v1.5.0. 

Bottom line, you can now use Nginx and also have the possibility to tweak its default settings via a `nginx_values.yaml` values file if you want to) or via a `HelmChartConfig` after deployment. Here's its default config, which seems to work great. 

https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/851ac92eb4ca8c383362d24acdb9ae28360d3f27/locals.tf#L318-L333.

To activate Nginx ingress, just set these variables in your `kube.tf`:

```
enable_traefik = false
enable_nginx = true
```

However, if you see possible tweaks or adjustments, PRs always welcome! 🙏

otavio:
I didn't add haproxy yet. I am wondering if I would be able to use flux to update haproxy settings, if need.

mysticaltech:
Hey @otavio, good to hear from you! I thought you were using Nginx; anyways, PR welcome for HAProxy; just please follow the same path as Nginx, so grep for "nginx" in any code editor to see what was done for it.

About flux, what is it? If needed, along with HAProxy and it is a standard tool that can be installed via HelmChat definitions, yes, sure, please add it; if not, then if it can be installed separately by the user, it is wiser to leave it out as anything that is not needed for the deployment phase shouldn't be included.

PR #323, opened 2022-09-26
Add LB IPv6 to the outputs

captnCC:
First of all a word of thanks, this is project is awesome!

This PR simply adds the IPv6 address of the load balancer to the outputs.

### DIFF
--- a/output.tf
+++ b/output.tf
@@ -22,6 +22,11 @@ output "load_balancer_public_ipv4" {
   value       = local.using_klipper_lb || local.ingress_controller == "none" ? null : data.hcloud_load_balancer.cluster[0].ipv4
 }
 
+output "load_balancer_public_ipv6" {
+  description = "The public IPv6 address of the Hetzner load balancer"
+  value       = local.using_klipper_lb || local.ingress_controller == "none" || var.load_balancer_disable_ipv6 ? null : data.hcloud_load_balancer.cluster[0].ipv6
+}
+
 output "kubeconfig_file" {
   value       = local.kubeconfig_external
   description = "Kubeconfig file content with external IP address"
--- a/output.tf
+++ b/output.tf
@@ -19,12 +19,12 @@ output "agents_public_ipv4" {
 
 output "load_balancer_public_ipv4" {
   description = "The public IPv4 address of the Hetzner load balancer"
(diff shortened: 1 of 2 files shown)


output.tf
```diff
@@ -22,6 +22,11 @@ output "load_balancer_public_ipv4" {
   value       = local.using_klipper_lb || local.ingress_controller == "none" ? null : data.hcloud_load_balancer.cluster[0].ipv4
 }
 
+output "load_balancer_public_ipv6" {
+  description = "The public IPv6 address of the Hetzner load balancer"
+  value       = local.using_klipper_lb || local.ingress_controller == "none" || var.load_balancer_disable_ipv6 ? null : data.hcloud_load_balancer.cluster[0].ipv6
```
phaer, line 27: I'd personally prefer adding brackets here, e.g. 

```
(local.using_klipper_lb || local.ingress_controller == "none" || var.load_balancer_disable_ipv6) ? null : data.hcloud_load_balancer.cluster[0].ipv6
 ```

As feature set has become quite big, the amount of flags grew. I am not sure all contributors could answer whether `||` or `? :` has higher precedence in terraform without looking it up ;)

phaer (approved):
I have **not** tested it, but code-wise looks good IMO.

Could be a good first step, while ipv6 support in other places (control nodes, etc) is still missing,.

output.tf
mysticaltech: @phaer Makes sense, but he's just copied the logic of the IPv4 LB output above, so it's ok in that case!

captnCC: I tested it this far with enabled v6 and can confirm its working (the output itself and the LB handling incoming IPv6 traffic including proper forwarding headers).
Because this is restricted to the output of the LB IPs only, the state of IPv6 support in the other components shouldn't be of much concern. 
I added the brackets for improved readability of the code 👌🏻

otavio (approved):

mnencia (approved):
Tested. It works like a charm

output.tf
phaer: @captnCC Thanks! I just wanted to note the general state of ipv6-support could still be improved further. But this is a great first step as it fixes ipv6 for pretty much all users of public services hosted on top of kube-hetzner :tada:

mysticaltech:
Great work folks, and awesome first contribution @captnCC! Just released under v1.5.4 🚀

Issue #325, opened 2022-09-28
Set tls-san option to the array of control-plane node IPs both public and private


mysticaltech:
### Discussed in https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/discussions/324

<div type='discussions-op-text'>

<sup>Originally posted by **captnCC** September 27, 2022</sup>
Hello folks out there,
I'm currently playing around with this module to create clusters and deploy applications on it.

The straight forward path seemed to be to use the cluster module together with the Kubernetes provider.

So I take the retuned `kubeconfig` from the cluster module and put it into the Kubernetes provider:
```terraform
provider "kubernetes" {
  host                   = module.cluster.kubeconfig.host
  client_certificate     = module.cluster.kubeconfig.client_certificate
  client_key             = module.cluster.kubeconfig.client_key
  cluster_ca_certificate = module.cluster.kubeconfig.cluster_ca_certificate
}
```

This works well after the cluster is created, but on the first run most resources will fail with an cert error  
`Failed to create deployment: Post "https://142.132.239.194:6443/apis/apps/v1/namespaces/default/deployments": x509: certificate is valid for 10.255.0.101, 10.43.0.1, 127.0.0.1, ::1, not 142.132.239.194`. 
Weirdly enough in some instances a part of the resources would succeed while the rest fails. Even adding a delay of 3 minutes before creating the resources didn't resolve that issue.

As base of the infrastructure I needed some additional components to be installed via helm. So I take the same config and put it in the helm provider:
```terraform
provider "helm" {
  kubernetes {
    host                   = module.cluster.kubeconfig.host
    client_certificate     = module.cluster.kubeconfig.client_certificate
    client_key             = module.cluster.kubeconfig.client_key
    cluster_ca_certificate = module.cluster.kubeconfig.cluster_ca_certificate
  }
}
```

And to my surprise this just works fine. 

Has anyone faced the same issue and what is your workaround? And the more general question: Keep infra and application in the same place or split it up in 2 separated terraform workflows? 

#### Update

By playing around with the order and dependencies of resources I found that generally the first request(s) to the Kubernetes API fail both with Kubernetes or Helm resources. </div>

captnCC:
Thanks @mysticaltech for the hint, this was an easy fix, but making the setup much more reliable!

mysticaltech:
Released in v1.5.5 🚀

PR #327, opened 2022-09-28
Set tls-san to external ip for control planes

captnCC:
In response to #325 

This PR sets the external IP of the control planes to the `tis-san` for the certificate to prevent certificate errors on the first request to the Kubernetes API.
When the control plane load balancer is used, nothing should change. I tested it with 1 and 3 control planes and the connection to each cp works properly.
I'm still a bit unsure about `tls-san` in `control-planes.tf`, should it include only the IP of the current server or the list of all CP IPs?

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ brew install hcloud
 
 _A complete reference of all inputs, outputs, modules etc. can be found in the [terraform.md](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/master/docs/terraform.md) file._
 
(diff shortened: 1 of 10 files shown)


mysticaltech:
Thanks for that @captnCC! 🚀

Please use `terraform fmt` to pass the validations. Or if you use VSCode, you can install the [Terraform](https://marketplace.visualstudio.com/items?itemName=HashiCorp.terraform) extension, which does that for you automatically.

captnCC:
@mysticaltech Done 👌🏻

mysticaltech:
Perfect! Merging to staging, and will escalate to master and release in a few minutes if all is good.

PR #331, opened 2022-09-30
Extra manifests

captnCC:
Following up to #328 this PR implements support for a user defined kustomization.

The extra step is triggered if `extra-manifests/kustomization.yaml.tpl` exists. 
It will be executed after `kustomization` was successful, so it can't break the cluster build.
If this step fails (e.g. `kubectl -k apply) the manifest can be fixed and run again, without rebuilding the cluster!


mysticaltech:
@captnCC Wonderful, will review ASAP today or tomorrow!!

mysticaltech:
EXCELLENT is the only word that comes to mind! Merging to master directly! 🚀 🙏🏻🙏🏻

mysticaltech:
Just released in v1.5.6, awesome contribution @captnCC!!

PR #333, opened 2022-10-02
Fix placement groups

nupplaphil:
I'm working with ~20-22 nodes and everytime, I re-created the whole cluster, it failed because of wrong placement_group mappings (it tried to map >10 nodes at one placement_group)

As far as I could calculate, the placement_group logic is starting one ID above 0 (1) and accidentally used the terraform overflow-logic (if n+1 is longer than the array-size, start from 0)

I think I solved it by slightly replace the calculation :-)

### DIFF
--- a/agents.tf
+++ b/agents.tf
@@ -15,7 +15,7 @@ module "agents" {
   ssh_private_key            = var.ssh_private_key
   ssh_additional_public_keys = var.ssh_additional_public_keys
   firewall_ids               = [hcloud_firewall.k3s.id]
-  placement_group_id         = var.placement_group_disable ? 0 : element(hcloud_placement_group.agent.*.id, ceil(each.value.index / 10))
+  placement_group_id         = var.placement_group_disable ? 0 : hcloud_placement_group.agent[floor(each.value.index / 10)].id
   location                   = each.value.location
   server_type                = each.value.server_type
   ipv4_subnet_id             = hcloud_network_subnet.agent[[for i, v in var.agent_nodepools : i if v.name == each.value.nodepool_name][0]].id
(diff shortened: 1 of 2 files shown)


mysticaltech (changes requested):
Glad to see such a decent size cluster being deployed with our tool. Thanks for your suggestion @nupplaphil.

agents.tf
```diff
@@ -15,7 +15,7 @@ module "agents" {
   ssh_private_key            = var.ssh_private_key
   ssh_additional_public_keys = var.ssh_additional_public_keys
   firewall_ids               = [hcloud_firewall.k3s.id]
-  placement_group_id         = var.placement_group_disable ? 0 : element(hcloud_placement_group.agent.*.id, ceil(each.value.index / 10))
```
mysticaltech, line 18: There is a +1 issue indeed for the index because we did not consider that it starts at 0. So just doing -1 should fix it, instead of using the floor function!

Please try this, and if good, do submit the changes so we can merge 🙏🏻

placement_group_id   = var.placement_group_disable ? 0 : element(hcloud_placement_group.agent.*.id, ceil(each.value.index / 10) - 1)

mysticaltech: @nupplaphil, Any updates on this? It is an important fix.

nupplaphil:
I cannot update this PR before tomorrow evening (in about 20hours), sry :-/

mysticaltech:
No worries @nupplaphil, thanks good, thanks!!

mysticaltech:
Am awaiting confirmation to merge and release! Just try the above proposed changes with +20 setup please, and confirm all good! :) Whenever you can...

mysticaltech:
@nupplaphil What I would like you is test the above. With 20-something nodes, the bill can go up fast! But since you use that already... And it's your PR, I though you could test it.

agents.tf
mysticaltech: You are right, the floor function works great here! Thanks.

mysticaltech:
I did a small test, and the floor function works great here, indeed. Thanks for this fix @nupplaphil!

PR #335, opened 2022-10-03
Allow `0` in cluster name

captnCC:
Just a minor problem: the cluster name isn't allowed to contain a `0` which may cause problems with auto-generated names. 
It don't see any reason why the `0` should be excluded (correct me if im wrong)


mysticaltech:
Sounds legit - I have a vague memory of this 0 thing as if I disabled it on purpose, but I can't remember the reason why now. So it's good, we can add it.

Issue #332, opened 2022-09-30
No public IP output if load balancer is disabled


captnCC:
Currently there are only IP(v4/v6) outputs if the external load balancer from Hetzner is used. 
For single-node clusters, the IPs need to extracted from the `control_planes_public_ipv4` output (and no IPv6).

To make the usage of the Ingress IPs more consistent down the line, they should have a consistent output for both v4 and v6.

The implementation is fairly simple, just one decision needs to be done: should the behaviour of the existing `load_balancer_public_ipv4`|`load_balancer_public_ipv6` be extended to include the kilpper-lb IP or should new outputs `ingress_public_ipv4`|`ingress_public_ipv6` be added?

mysticaltech:
@captnCC Not sure I follow you all the way here. The problem with Klipper is that it gives many IPs, so it's an array, with the external IPs of all the nodes, as it's an on-metal LB provider.

mysticaltech:
You are right; for a single node cluster, it should be outputted! Please don't hesitate to take this on.

captnCC:
@mysticaltech Yes especially the single-node where my main point. For dev/prod-setups this is very useful (when the dev cluster is a dev-cluster is single node, and prod a full-blown cluster with load balancer). Then there is no need for extra logic to set DNS records for the cluster! 

As usual, the PR will follow soon!

mysticaltech:
"No need for extra logic", exactly @captnCC !! 🙏🏻

PR #336, opened 2022-10-03
Output the cp ip when using klipper

captnCC:
Follow up to #332.

This adds 2 new outputs `ingress_public_ipv4` and `ingress_public_ipv6`.
They support `hetzner_lb` and `klipper` and are more consistent to use between single-node clusters with `klipper` and full blown clusters.

To be conservative I added them as new outputs instead of extending the existing `load_balancer_public_ipv4|6`.

Since I haven't experience with klipper setups a few questions remain: 
How to handle setups besides single-node clusters where it is save to assume the first control-plane accepts the traffic?

My best guess would be something like this: (just some rough code for understanding)
```
if(!local.using_kilpper_lb){
   return  data.hcloud_load_balancer.cluster[0].ipv4
}

if(local.is_single_node_cluster){
  return module.control_planes[keys(module.control_planes)[0]].ipv4_address
}

return module.agents[keys(module.agents)[0]].ipv4_address 
```

In addition there is still #284 open. To solve this in one go, I would drop `local.ingress_controller == "none"` completely and make this output solely rely on the load balancing config and not the ingress controller itself.

### DIFF
--- a/output.tf
+++ b/output.tf
@@ -7,14 +7,14 @@ output "control_planes_public_ipv4" {
   value = [
     for obj in module.control_planes : obj.ipv4_address
   ]
-  description = "The public IPv4 addresses of the controlplane server."
+  description = "The public IPv4 addresses of the controlplane servers."
 }
 
 output "agents_public_ipv4" {
   value = [
     for obj in module.agents : obj.ipv4_address
   ]
-  description = "The public IPv4 addresses of the agent server."
+  description = "The public IPv4 addresses of the agent servers."
 }
 
 output "load_balancer_public_ipv4" {
@@ -37,4 +37,16 @@ output "kubeconfig" {
   description = "Structured kubeconfig data to supply to other providers"
   value       = local.kubeconfig_data
   sensitive   = true
+}
+
+
+output "ingress_public_ipv4" {
+  description = "The public ingress IPv4 of the cluster (external/internal loadbalancer)"
+  value       = (local.ingress_controller == "none" ? null : (local.using_klipper_lb ? module.control_planes[keys(module.control_planes)[0]].ipv4_address : data.hcloud_load_balancer.cluster[0].ipv4))
+}
+
+
+output "ingress_public_ipv6" {
+  description = "The public ingress IPv6 of the cluster (external/internal loadbalancer)"
+  value       = (local.ingress_controller == "none" || var.load_balancer_disable_ipv6 ? null : (local.using_klipper_lb ? module.control_planes[keys(module.control_planes)[0]].ipv6_address : data.hcloud_load_balancer.cluster[0].ipv6))
 }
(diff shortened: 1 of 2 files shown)


output.tf
```diff
@@ -38,3 +38,15 @@ output "kubeconfig" {
...
+
+output "ingress_public_ipv4" {
+  description = "The public ingress IPv4 of the cluster (external/internal loadbalancer)"
+  value       = (local.ingress_controller == "none" ? null : (local.using_klipper_lb ? module.control_planes[keys(module.control_planes)[0]].ipv4_address : data.hcloud_load_balancer.cluster[0].ipv4))
+}
+
+
+output "ingress_public_ipv6" {
+  description = "The public ingress IPv6 of the cluster (external/internal loadbalancer)"
+  value       = (local.ingress_controller == "none" || var.load_balancer_disable_ipv6 ? null : (local.using_klipper_lb ? module.control_planes[keys(module.control_planes)[0]].ipv6_address : data.hcloud_load_balancer.cluster[0].ipv6))
```
mysticaltech, line 51: @captnCC Are you sure that nodes get IPv6 addresses too? I do not remember seeing those.

mysticaltech: In the case of Klipper that is! Needs to be tested.

mysticaltech:
@captnCC Thank you for this. In the case of klipper, all nodes can receive traffic, so it would be a simplification to mention only the first node; however why not! At the end of the day, I do not really see klipper really used in production.

captnCC:
I think its a simplification in default usage, using klipper in production would need an advanced setup anyway

mysticaltech:
Exactly... Only thing am not sure about is the IPv6 thing for Klipper, am not sure that nodes have those currently, if you can test it would be great.

output.tf
mysticaltech: Ok, it should be ok then! Thanks for the info.

mysticaltech:
Thanks again for this @captnCC, just released along with the other PR in v1.5.7 🙏🏻

mysticaltech:
@captnCC Sorry long day today, local.ingress_controller == "none", yes we can remove that in a subsequent PR. Please shoot whenever you want.

mysticaltech:
Hey Johannes, I have invited you to join Kube-Hetzner; if you accept, it would be an honor! I understand that your contributions are for work because you probably use it, and it's really what it is for; we all use this tool for our work. But I love the way you think, and it would be an honor to give you more power to shape the future of this project. Cheers!!

captnCC:
@mysticaltech I feel honoured, thanks for the kind words and trust! 

To continue on the topic, indeed you were right: klipper itself doesn't accept IPv6 (at least in the current configuration).
I will create a new PR handling the IPv6 output accordingly and remove the `local.ingress_controller == "none"` portion!

mysticaltech:
Awesome @captnCC, thanks!

mysticaltech:
@captnCC We have a small chat room for occasional direct discussions, please feel free to join..

Join my space Kube-Hetzner https://matrix.to/#/#kube-hetzner:matrix.org

PR #337, opened 2022-10-04
Ignore ingress config for ip outputs

captnCC:
Followup to #336

And this should be the final one!
- `ingress_public_ipv4` is now always populated, either with the lb ip or with the ip of the first control plane.
- `ingress_public_ipv6 ` is only populated when a load balancer is used and IPv6 is not disabled

This should clear #284!

### DIFF
--- a/data.tf
+++ b/data.tf
@@ -18,7 +18,7 @@ data "github_release" "kured" {
 }
 
 data "hcloud_load_balancer" "cluster" {
-  count = local.using_klipper_lb ? 0 : local.ingress_controller == "none" ? 0 : 1
+  count = local.has_external_load_balancer ? 0 : 1
   name  = var.cluster_name
 
   depends_on = [null_resource.kustomization]
--- a/init.tf
+++ b/init.tf
@@ -269,7 +269,7 @@ resource "null_resource" "kustomization" {
         "sleep 5", # important as the system upgrade controller CRDs sometimes don't get ready right away, especially with Cilium.
         "kubectl -n system-upgrade apply -f /var/post_install/plans.yaml"
       ],
-      local.using_klipper_lb || local.ingress_controller == "none" ? [] : [
+      local.has_external_load_balancer ? [] : [
         <<-EOT
       timeout 180 bash <<EOF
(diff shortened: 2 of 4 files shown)


mysticaltech:
@captnCC Looking good man, but what about the case when there is no ingress controller defined, so no load balencers deployed, this will result in an error, no? Maybe we can use the try() function?

otavio:
Couldn't the load balancer be deployed by Terraform even when no chart is installed? This allows for my use case where I install it using fluxcd but load balancer IP would be handled by Terraform and allow for use with other plugins (DNS, for example).

mysticaltech:
@otavio Yes, totally possible, especially with the new Kustomization ability (see readme). The above is just to make sure we do not get into errors when no load balancers are present at all.

captnCC:
Oh yes I see your point, I oversaw that it is possible to configure it to neither use klipper nor an external load balancer. 
I refactored some internals to unify the logic of "do we have a load balancer". So either there is an external load balancer or we will output the control planes ip.

But I don't know if that would help @otavio.

mysticaltech:
That's great @captnCC, thank you!! 

@otavio That may still not address your use case, please do not hesitate to submit a PR for what you need exactly.

otavio:
It doesn't indeed. I'd like to have load balancer created using Terraform so IPs can be known either way.

captnCC:
@otavio The main point of the hetzner ccm is, that the load balancer is provisioned from within the cluster to give it the proper configs. So far I know it is not possible to instruct the ccm to reuse an existing load balancer.

But is there something that prevents you from creating the LB yourself in terraform and use it as you need?

otavio:
No, I just expected this module to offer this option as it offers an opportunity for load balancer creation, and it does the provisioning of control plane load balancer; we could have the same for agents.

captnCC:
But that are 2 completely different use cases. The cp load balancer just serves the `tcp` traffic on port `6443` to the control planes, no further configuration needed. The ingress load balancer is part of the hetzner integration via the ccm to deliver proper interoperability. So at this point it is not possible to provision it via terraform.

I assume in your setup you set `enable_nginx = false` and `enable_traefik = false` and `enable_klipper_metal_lb = false`.

So the load balancer will be provisioned after you installed your own ingress. At this point you can get the lb ip the same way the module does it:

```
data "hcloud_load_balancer" "cluster" {
  count = local.has_external_load_balancer ? 0 : 1
  name  = var.cluster_name

  depends_on = [null_resource.kustomization]
}
```

It just needs to be timed right with the provisioning.

Issue #338, opened 2022-10-05
Error on create single node cluster


panakour:
I try to create a single node cluster using the count of 0 to nodepools:

```
  control_plane_nodepools = [
    {
      name        = "control-plane-fsn1",
      server_type = "cpx31",
      location    = "nbg1",
      labels      = [],
      taints      = [],
      count       = 1
    }
  ]

  agent_nodepools = [
    {
      name        = "agent-small",
      server_type = "cpx11",
      location    = "fsn1",
      labels      = [],
      taints      = [],
      count       = 0
    }
  ]
```

But i get this error
<img width="1461" alt="image" src="https://user-images.githubusercontent.com/12927166/194012070-576a394d-30f7-4a72-b0b2-e7d5295454fb.png">

In the past it was working

mysticaltech:
@panakour Please try again recreating the cluster with `hetzner_ccm_version="v1.12.1"`, which will ensure that workloads can be executed on the control plane and ensure your setup works. 

More on the matter here https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/311

panakour:
@mysticaltech Thanks a lot now its working

otavio:
@mysticaltech the version could be fixed when using single node cluster, automatically.

mysticaltech:
@otavio The above is just a temporary patch. If you know of a real fix, please explain or PR, thank you! 🙏

Issue #311, opened 2022-09-19
The "allow_scheduling_on_control_plane = true" flag causes Timeout during creation


cc-nogueira:
Hi, I am having a timeout issue when creating a cluster. This is the tail of the output:

> module.kube-hetzner.null_resource.kustomization (remote-exec): Waiting for the system-upgrade-controller deployment to become available...
> module.kube-hetzner.null_resource.kustomization (remote-exec): + kubectl -n system-upgrade wait --for=condition=available --timeout=180s deployment/system-upgrade-controller
> module.kube-hetzner.null_resource.kustomization: Still creating... [30s elapsed]
> ...
> module.kube-hetzner.null_resource.kustomization: Still creating... [3m21s elapsed]
> module.kube-hetzner.null_resource.kustomization (remote-exec): error: timed out waiting for the condition on deployments/system-upgrade-controller
> ╷
>  │ Error: remote-exec provisioner error
>  │
>  │   with module.kube-hetzner.null_resource.kustomization,
>  │   on .terraform/modules/kube-hetzner/init.tf line 231, in resource "null_resource" "kustomization":
>  │  231:   provisioner "remote-exec" {
>  │
>  │ error executing "/tmp/terraform_39240884.sh": Process exited with status 1
> ╵

I have retried a few times. Also destroyed and retried again.

Here is my kube.tf:

```
locals {
  hcloud_token = "<token>"
}

module "kube-hetzner" {
  providers = {
    hcloud = hcloud
  }
  hcloud_token = local.hcloud_token

  source = "kube-hetzner/kube-hetzner/hcloud"

  ssh_public_key = file("/home/username/.ssh/id_ed25519.pub")
  ssh_private_key = file("/home/username/.ssh/id_ed25519")

  network_region = "us-east"

  control_plane_nodepools = [
    {
      name        = "control-plane",
      server_type = "cpx11",
      location    = "ash",
      labels      = [],
      taints      = [],
      count       = 1
    }
  ]

  agent_nodepools = [
    {
      name        = "agent-small",
      server_type = "cpx11",
      location    = "ash",
      labels      = [],
      taints      = [],
      count       = 0
    }
  ]

  load_balancer_type     = "lb11"
  load_balancer_location = "ash"

  allow_scheduling_on_control_plane = true
  enable_cert_manager = true
  cluster_name = "k3s"

  extra_firewall_rules = [
    # For Postgres
    {
      direction       = "in"
[...]

PR #339, opened 2022-10-05
readme: single node setup

panakour:
related to: https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/338 and https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/311

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,7 +301,7 @@ spec:
 
 <summary>Single-node cluster</summary>
 
-Running a development cluster on a single node without any high availability is also possible. You need one control plane nodepool with a count of 1 and one agent nodepool with a count of 0.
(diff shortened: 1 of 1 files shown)


captnCC:
Looks good, thank you!

phaer:
Thanks for caring about documentation, but I am not sure whether we just want to recommend using an older version to "prevent some errors". We should rather get a better understanding of the bug, open it upstream and link the issue in the README so people know when it applies.

captnCC:
There is already an upstream issue on the ccm repo: hetznercloud/hcloud-cloud-controller-manager#282
But yes the upstream issue and #311 should be mentioned in the read me as it is already done in the `kube.tf.example`

mysticaltech:
Thanks, folks! @phaer You are right; we will delete that detail as soon as the issue is fixed; currently looking into it.

PR #340, opened 2022-10-05
Add switch to disable automatic os updates

captnCC:
This adds `automatically_upgrade_os` as an option to control the automatic system updates. 
As disabling them is needed for non HA clusters to work properly. This is far more convenient than the "ssh in each node" method described right now in the docs.

For convince it might be even better to set this automatically depended on the node counts, but it is contrasted by the security risk. There for I personally would tend to leaving it as user choice.

### DIFF
--- a/modules/host/main.tf
+++ b/modules/host/main.tf
@@ -135,7 +135,7 @@ resource "hcloud_server" "server" {
     EOT
   }
 
-  # Cleanup ssh identity file 
+  # Cleanup ssh identity file
   provisioner "local-exec" {
     command = <<-EOT
       rm /tmp/${random_string.identity_file.id}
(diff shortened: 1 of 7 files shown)


modules/host/main.tf
```diff
@@ -152,6 +152,18 @@ resource "hcloud_server" "server" {
       EOT
     ]
   }
+
+  provisioner "remote-exec" {
+    inline = var.automatically_upgrade_os ? [<<-EOT
+      echo "Automatic updates stay enabled"
```
otavio, line 158: Could here it enable the service? so if user changes the variable, it'd re-enable it.

captnCC: As its part of the node provisioning, the nodes would need to be re-create/provisioned for this to work

captnCC:
@mysticaltech TIL: `terraform fmt` doesn't work recursively, it needs to be executed in the submodule directory itself!

phaer:
> @mysticaltech TIL: `terraform fmt` doesn't work recursively, it needs to be executed in the submodule directory itself!

There's a `-recursive` flag.

PR #350, opened 2022-10-15
Opt out of local files and add kubeconfig as an output to the kube.tf example

### LINKED ISSUE #349: Don't use local_sensitive_file for kubeconfig
Hi,

I'm trying to integrate the terraform run into a CI/CD pipeline alongside other terraform modules. This causes the kubeconfig file to be lost on each run and it always detects it as a change. Would it be possible to disable the creation of that file and use outputs to handle that?

At first glance this file doesn't seem to be required for the module to work.

Probably this can be moved to an output of the module and the kube.tf could include something like:
```tf
output "kubeconfig" {
    value = module.kube-hetzner.kubeconfig_file
    sensitive = true
}
```

Then the docs could state to export the kubeconfig by executing:
```
terraform output --raw kubeconfig > kubeconfig.yaml
```

WolfspiritM:
Fixes #349 by removing the local kubeconfig file and documented how to export it from the terraform output

### DIFF
--- a/kubeconfig.tf
+++ b/kubeconfig.tf
@@ -15,7 +15,7 @@ locals {
   kubeconfig_external = replace(data.remote_file.kubeconfig.content, "127.0.0.1", var.use_control_plane_lb ? hcloud_load_balancer.control_plane.*.ipv4[0] : module.control_planes[keys(module.control_planes)[0]].ipv4_address)
   kubeconfig_parsed   = yamldecode(local.kubeconfig_external)
   kubeconfig_data = {
-    host                   = var.use_control_plane_lb ? hcloud_load_balancer.control_plane.*.ipv4[0] : local.kubeconfig_parsed["clusters"][0]["cluster"]["server"]
+    host                   = local.kubeconfig_parsed["clusters"][0]["cluster"]["server"]
     client_certificate     = base64decode(local.kubeconfig_parsed["users"][0]["user"]["client-certificate-data"])
     client_key             = base64decode(local.kubeconfig_parsed["users"][0]["user"]["client-key-data"])
     cluster_ca_certificate = base64decode(local.kubeconfig_parsed["clusters"][0]["cluster"]["certificate-authority-data"])
@@ -23,6 +23,7 @@ locals {
 }
 
 resource "local_sensitive_file" "kubeconfig" {
+  count           = var.create_kubeconfig ? 1 : 0
   content         = local.kubeconfig_external
   filename        = "${var.cluster_name}_kubeconfig.yaml"
   file_permission = "600"
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ It will take around 5 minutes to complete, and then you should see a green outpu
 
 When your brand new cluster is up and running, the sky is your limit! 🎉
 
-You can immediately kubectl into it (using the `clustername_kubeconfig.yaml` saved to the project's directory after the installation). By doing `kubectl --kubeconfig clustername_kubeconfig.yaml`, but for more convenience, either create a symlink from `~/.kube/config` to `clustername_kubeconfig.yaml` or add an export statement to your `~/.bashrc` or `~/.zshrc` file, as follows (you can get the path of `clustername_kubeconfig.yaml` by running `pwd`):
+You can immediately kubectl into it (by exporting the kubeconfig to a file via `terraform output --raw kubeconfig > clustername_kubeconfig.yaml` after the installation). By doing `kubectl --kubeconfig clustername_kubeconfig.yaml`, but for more convenience, either create a symlink from `~/.kube/config` to `clustername_kubeconfig.yaml` or add an export statement to your `~/.bashrc` or `~/.zshrc` file, as follows (you can get the path of `clustername_kubeconfig.yaml` by running `pwd`):
 
 ```sh
 export KUBECONFIG=/<path-to>/clustername_kubeconfig.yaml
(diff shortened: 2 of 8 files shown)


mysticaltech:
@WolfspiritM Good work here, but the thing is that we need to keep some form of backward compatibility. Can't you just add the output and leave the rest?

WolfspiritM:
@mysticaltech But the rest is actually the problem not the output. The output is just for convenience so users can still get the kubeconfig. The problem is that in CI/CD pipelines the missing file is detected as a change on each run cause it's not commited. I think that creating the file in a version controlled git terraform repository might even be a security risk as the kubeconfig might get committed.

See also https://registry.terraform.io/providers/hashicorp/local/latest/docs/resources/sensitive_file
```
When working with local files, Terraform will detect the resource as having been deleted each time a configuration is applied on a new machine where the file is not present and will generate a diff to re-create it. This may cause "noise" in diffs in environments where configurations are routinely applied by many different users or within automation systems.
```
So a local file should not be used unless really needed. Another "hacky" solution might be to use a null_resource and create the file just once via the null resource. When deleting the file it will then not recreate it on each run if it is missing.

A workaround for now could be a new variable "create_kubeconfig" which defaults to true and the local file:

```tf
resource "local_sensitive_file" "kubeconfig" {
  count           = var.create_kubeconfig ? 1 : 0
  content         = local.kubeconfig_external
  filename        = "${var.cluster_name}_kubeconfig.yaml"
  file_permission = "600"
}
```
That will however create the resource as an array which terraform should detect that it moved (I hope).
If not will just end up deleting and recreating the file once on the next apply.

So in CI runs this can be set to false which prevents the resource from being created.

Would that be a valid solution?

mysticaltech:
I love your last proposal! 🙏

mysticaltech:
@WolfspiritM This is not yet complete is it? The new variable is not used anywhere..

mysticaltech:
@WolfspiritM What are your thoughts on that? What you are proposing is sweet. 

We are moving to 1.6 soon, so that could be a breaking change if you make your point. But I would like to possibly keep it "working out of the box" without the extra terraform command by default. Then for folks like you that integrate this, we could disable it. 

Please share your thoughts if you think the choices are not optimal.

WolfspiritM:
> @WolfspiritM What are your thoughts on that? What you are proposing is sweet.
> We are moving to 1.6 soon, so that could be a breaking change if you make your point. But I would like to possibly keep it "working out of the box" without the extra terraform command by default. Then for folks like you that integrate this, we could disable it.

At first: It should be done now. Can you take a look? 😃 

My thought is that for new people who copy the kube.tf.example it should be already disabled. That's why I added it there with false. As I think it can be a security issue if someone accidentially commits that file and isn't aware of it.

For people who copied the kube.tf.example before that it should still work now as I set it to true as default. Terraform should mark the resource as moved and otherwise not touch it.

I also had to do the same for the kustomize backup. In general I think the backup doesn't make much sense as the real templates aren't even backuped at all. But that's for another PR I think.

mysticaltech:
Awesome @WolfspiritM! Sweet work there 🙏 Thanks a lot, I will release ASAP, in 1.6, and will set to `false` by default!

About the kustomization, I was thinking about it too; great you added support for it! You are right, that is file is not very useful in its current state, but better than nothing for now.

dimisjim:
An improvement to this would be to generate the kubecg yaml by abiding to the `cluster_name` module input var

timowevel1:
Hey, on the latest commit I receive the following when I try to do terraform validate:
╷
│ Error: Unsupported argument
│
│   on kube.tf line 319, in module "kube-hetzner":
│  319:   create_kubeconfig = false
│
│ An argument named "create_kubeconfig" is not expected here.

WolfspiritM:
> Hey, on the latest commit I receive the following when I try to do terraform validate: ╷ │ Error: Unsupported argument │ │ on kube.tf line 319, in module "kube-hetzner": │ 319: create_kubeconfig = false │ │ An argument named "create_kubeconfig" is not expected here.

@timowevel1 

The reason is probably that you're using the master kube.tf.example file with the terraform registry which is not yet the master version but the released version.

Either use `source = "github.com/kube-hetzner/terraform-hcloud-kube-hetzner"` to be on the development/master version

Or use the kube.tf.example from the tagged release here:
https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/tree/v1.5.10

mysticaltech:
@timowevel1 Yes, as @WolfspiritM stated, use the github release and run `terraform init -upgrade`.

@WolfspiritM It is not yet available in a tagged release, probably tomorrow, it will drop with 1.16, with the new autoscaling capability.

kubeconfig.tf
```diff
@@ -15,15 +15,16 @@ locals {
   kubeconfig_external = replace(data.remote_file.kubeconfig.content, "127.0.0.1", var.use_control_plane_lb ? hcloud_load_balancer.control_plane.*.ipv4[0] : module.control_planes[keys(module.control_planes)[0]].ipv4_address)
   kubeconfig_parsed   = yamldecode(local.kubeconfig_external)
   kubeconfig_data = {
-    host                   = var.use_control_plane_lb ? hcloud_load_balancer.control_plane.*.ipv4[0] : local.kubeconfig_parsed["clusters"][0]["cluster"]["server"]
```
mysticaltech, line 18: @WolfspiritM I had missed this, it is actually important to keep it as is. We have the possibility to have a dedicated LB for the control plane nodes, to load balance calls to the kube-api for instance. Or using kubectl in an "HA" manner!

PR #351, opened 2022-10-15
Autoscaling

mysticaltech:
@MagnusSvendsen @Nigma1337 Please let's continue your good work on the autoscaling feature here. There will be less noise as on the issue.


agents.tf
```diff
@@ -133,3 +133,37 @@ resource "null_resource" "configure_longhorn_volume" {
     hcloud_volume.longhorn_volume
   ]
 }
+
+resource "null_resource" "configure_autoscaling" {
+  count = var.max_number_nodes_autoscaler ? 1 : 0
+
+  provisioner "hcloud_snapshot" "autoscale_image"{
+    server_id = hcloud_server.service.id
```
mysticaltech, line 141: @MagnusSvendsen How does it select which agent from which agent node pool to create a snapshot from? I propose we either add an attribute of autoscale = true in the node pool definition and look for it, or just take the last node pool definition, and snapshot the first or last node.

Probably something can be done through the count and the dependencies of this null-resource. What do you think?

agents.tf
```diff
@@ -133,3 +133,37 @@ resource "null_resource" "configure_longhorn_volume" {
...
+  }
+  # Create autoscaling configfile
+  provisioner "file" {
+    content = templatefile(
+      "templates/hetzner_autoscaler_config.yaml.tpl",
+      {
+        cloudinit_config = data.cloudinit_config
+        nodepool_name = var.name
+        server_type = var.server_type
+        location = var.location
```
mysticaltech, line 155: All the vars in here need to be copied automatically from the node pool/node which was snapshotted so that they match the actual "shape and form" of the snapshot.

PR #352, opened 2022-10-15
Autoscaling

mysticaltech:
@MagnusSvendsen @Nigma1337 I built upon your branch. Please have a look and let's discuss this here.


mysticaltech:
@Nigma1337 It's now executing at least, let's see if it works. Maybe you can take it on from here. If you happen to not be able to push directly to this branch, please open a new PR pointing to the `autoscaling` branch.

mysticaltech:
For now the snapshot is correctly created:

ksnip_20221015-131440

And the autoscaler gets installed, but the pod crashes, I do not yet know why. Any help in debugging this would be very much appreciated!

ksnip_20221015-131734

mysticaltech:
Ah, it's failing because of this error, any ideas?
ksnip_20221015-132406

mysticaltech:
@mnencia @phaer Any ideas on the certificate issue above? 🙏

phaer:
Error seems to imply that mounting a single file in this context is forbidden? Maybe it's not a bind mount? Just switching from the file to the parent directory should be safe, as all contents of `/etc/ssl/certs/` are public anyways.

templates/autoscaler.yaml.tpl
```diff
@@ -0,0 +1,189 @@
...
+                  key: token
+#         - name: HCLOUD_SSH_KEY
+#            value: <optional SSH Key Name or ID>
+          - name: HCLOUD_NETWORK
+            value: ${ipv4_subnet_id}
+          - name: HCLOUD_IMAGE
+            value: ${snapshot_label}
+          volumeMounts:
+            - name: ssl-certs
+              mountPath: /etc/ssl/certs/ca-certificates.crt
```
phaer, line 183: Changing this to `/etc/ssl/certs` should fix the certificate error.

templates/autoscaler.yaml.tpl
```diff
@@ -0,0 +1,189 @@
...
+              readOnly: true
+          imagePullPolicy: "Always"
+      volumes:
+        - name: ssl-certs
+          hostPath:
+            path: "/etc/ssl/certs/ca-certificates.crt"
```
phaer, line 189: This should be /etc/ssl/certs as well

mysticaltech:
Thanks @phaer, appreciate it! Will try the above later this week-end.

mnencia:
I've just tried the patch. It creates the snapshot, but the cluster autoscaler pod cannot get scheduled because the control plane has "node-role.kubernetes.io/control-plane" taint instead of "node-role.kubernetes.io/master".

mnencia:
I've pushed a few changes. Now the autoscaler pod boots, but I didn't test if it works.

mysticaltech:
Beautiful! So good to see you commit again, @mnencia; it's been so long! Thanks for that 🙏 💙

templates/autoscaler.yaml.tpl
mysticaltech: Marco did that, and it worked! Thanks, @phaer 🙏 ✨

mysticaltech:
Closing in favor of #353.

PR #353, opened 2022-10-17
Autoscaler added

ifeulner:
Hi @mysticaltech I merged the autoscaling branch with my working branch, and I got the first successful node-creation by autoscaler when I apply a random workload like this:

```
apps/v1
kind: Deployment
metadata:
  labels:
    app: hello-world
  name: hello-world
spec:
  replicas: 4
  selector:
    matchLabels:
      app: hello-world
  strategy:
    rollingUpdate:
      maxSurge: 1
      maxUnavailable: 0
    type: RollingUpdate
  template:
    metadata:
      labels:
        app: hello-world
    spec:
      containers:
      - image: rancher/hello-world
        imagePullPolicy: Always
        name: hello-world
        ports:
        - containerPort: 80
          protocol: TCP
        resources:
          limits:
            cpu: 1000m
            memory: 1024Mi
          requests:
            cpu: 1000m
            memory: 1024Mi
```

A special cloud-init is necessary, as the old k3s stuff has to be wiped out. I am using the control-plan, as this will allow also for a single-control-plane cluster with just autoscaling nodes.

Still a lot to clean up, also we should support multiple autoscale-node groups. 

What do you think?


mysticaltech:
Very happy to hear, @ifeulner ! I would guess, let's refine this for now, test it well, release and then we can add multiple autoscaling nodepool. But my preference is always to keep it simple, if there is no immediate necessity of making it complex.

I will test this ASAP, and you let me know when you feel it's ready.

ifeulner:
There are still some strange log-messages by k3s (probably from the installation before), so some work to do... 

for that it looks actually too good: 
```
❯ k get nodes -o wide
NAME                           STATUS   ROLES                       AGE   VERSION        INTERNAL-IP    EXTERNAL-IP       OS-IMAGE           KERNEL-VERSION    CONTAINER-RUNTIME
autoscaling-50d004da190858f8   Ready    <none>                      20m   v1.24.6+k3s1   10.255.0.1     142.132.227.144   openSUSE MicroOS   6.0.1-1-default   containerd://1.6.8-k3s1
k3s-agent-small-rtv            Ready    <none>                      88m   v1.24.6+k3s1   10.0.0.101     49.12.215.218     openSUSE MicroOS   6.0.1-1-default   containerd://1.6.8-k3s1
k3s-control-plane-fsn1-dfh     Ready    control-plane,etcd,master   88m   v1.24.6+k3s1   10.254.0.101   168.119.159.232   openSUSE MicroOS   6.0.1-1-default   containerd://1.6.8-k3s1
❯ k get pods -A -o wide
NAMESPACE        NAME                                               READY   STATUS    RESTARTS   AGE   IP             NODE                           NOMINATED NODE   READINESS GATES
default          hello-world-5c7d777db8-9jt22                       1/1     Running   0          22m   10.42.1.10     k3s-agent-small-rtv            <none>           <none>
default          hello-world-5c7d777db8-bj88d                       1/1     Running   0          20m   10.42.4.7      autoscaling-50d004da190858f8   <none>           <none>
default          hello-world-5c7d777db8-bsk4c                       1/1     Running   0          20m   10.42.4.8      autoscaling-50d004da190858f8   <none>           <none>
default          hello-world-5c7d777db8-x5srr                       1/1     Running   0          20m   10.42.4.9      autoscaling-50d004da190858f8   <none>           <none>
kube-system      cluster-autoscaler-59c65995bb-vx7nc                1/1     Running   0          35m   10.42.0.10     k3s-control-plane-fsn1-dfh     <none>           <none>
kube-system      coredns-b96499967-hjvxb                            1/1     Running   0          88m   10.42.0.4      k3s-control-plane-fsn1-dfh     <none>           <none>
kube-system      hcloud-cloud-controller-manager-57d697cf98-bqp49   1/1     Running   0          88m   10.254.0.101   k3s-control-plane-fsn1-dfh     <none>           <none>
kube-system      hcloud-csi-controller-0                            5/5     Running   0          87m   10.42.1.8      k3s-agent-small-rtv            <none>           <none>
kube-system      hcloud-csi-node-8zklp                              3/3     Running   0          88m   10.42.0.2      k3s-control-plane-fsn1-dfh     <none>           <none>
kube-system      hcloud-csi-node-nllp8                              3/3     Running   0          20m   10.42.4.2      autoscaling-50d004da190858f8   <none>           <none>
kube-system      hcloud-csi-node-xlr6v                              3/3     Running   0          88m   10.42.1.2      k3s-agent-small-rtv            <none>           <none>
kube-system      kured-2dj7x                                        1/1     Running   0          87m   10.42.1.6      k3s-agent-small-rtv            <none>           <none>
kube-system      kured-8b6xc                                        1/1     Running   0          20m   10.42.4.5      autoscaling-50d004da190858f8   <none>           <none>
kube-system      kured-kvdg8                                        1/1     Running   0          87m   10.42.0.6      k3s-control-plane-fsn1-dfh     <none>           <none>
kube-system      metrics-server-668d979685-npvs6                    1/1     Running   0          88m   10.42.0.3      k3s-control-plane-fsn1-dfh     <none>           <none>
kube-system      traefik-7b7d4549db-p52w7                           1/1     Running   0          87m   10.42.0.7      k3s-control-plane-fsn1-dfh     <none>           <none>
system-upgrade   system-upgrade-controller-5d4ff9f49-zp6ft          1/1     Running   0          88m   10.42.0.5      k3s-control-plane-fsn1-dfh     <none>           <non
```

mysticaltech:
So it did end up using 10.255 on its own; good to see!! Well done.

Don't hesitate to post the k3s logs!

mysticaltech:
@ifeulner Beautiful! Will do my own tests ASAP then merge, probably tomorrow.

Nice work!! 🙏 ✨

mnencia:
Can we keep this open for a bit more time? I'd like to test it but unfortunately, I've been too busy today.

mysticaltech:
Yes, @mnencia, absolutely! Just FYI merged another unrelated PR in `master` for 1.6.

If we can push the release before the end of the week, it would be awesome! 🚀

main.tf
```diff
@@ -15,13 +15,21 @@ resource "hcloud_network" "k3s" {
 }
 
 # We start from the end of the subnets cird array, 
-# as we would have fewer control plane nodepools, than angent ones.
+# First the the autoscaling subnet, then the control planes subnets
+# And finally starting from the beginning of the array, the agents subnets
+resource "hcloud_network_subnet" "autoscaling" {
```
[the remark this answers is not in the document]
mysticaltech: @ifeulner Finally, did you understand the logic that the autoscaler uses for IP allocation? Because it does not know about this subnet? So I get it jumps on any subnet randomly, and because we have the first 100 free for each (for LBs), it just works.

If that is the case, we can just delete the autoscaling subnet.

mysticaltech (commented):
@ifeulner I have forgotten to post my review. Please, if you can take a quick look and address those quickly, it would be amazing!! Sorry for the bother.

templates/autoscale-cloudinit.yaml.tpl
```diff
@@ -0,0 +1,138 @@
+instance-id: iid-abcde001
+
+#cloud-config
+
+debug: True
+
+write_files:
+
+# Configure the private network interface
```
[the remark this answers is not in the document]
ifeulner: Should be as idempotent as possible here - taking a snapshot of a running server is in general not a good solution.
So we should ensure that the server is in a proper state. E.g. if you apply the autoscaler when creating the cluster initially, the control_plane node is in another / earlier state as if you enable the autoscaler after creation. 
Best would be to use an externally / especially created snapshot with a proper state for the autoscaler.

autoscale-agents.tf
```diff
@@ -0,0 +1,103 @@
...
+    host           = module.control_planes[keys(module.control_planes)[0]].ipv4_address
+    port           = var.ssh_port
+  }
+
+  # Upload the Rancher config
+  provisioner "file" {
+    content = templatefile(
+      "${path.module}/templates/autoscaler.yaml.tpl",
+      {
+        #cloudinit_config - we have to check if this is necessary, if so we need to recreate it, or somehow extract it from server module, up to a higher level
```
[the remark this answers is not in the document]
ifeulner: done already in a previous commit.

main.tf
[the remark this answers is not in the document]
ifeulner: remove the subnet for now. But hope the bug gets fixed soon.

ifeulner:
@mysticaltech updated the pull request.

main.tf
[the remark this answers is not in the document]
mysticaltech: @ifeulner Is it a bug or a feature? I believe the latter, unfortunately, but we should indeed request support for subnets, which would make much more sense!

templates/autoscale-cloudinit.yaml.tpl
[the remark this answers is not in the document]
mysticaltech: Got it, thanks for the clarification. I trust your good judgment on this.

mysticaltech:
Thank you @ifeulner!! Looking good.

@mnencia The ball is fully in your camp now! ✌️

ifeulner:
> It would get us closer to the format already used for the node pools. What do you think?

@mysticaltech sure, would work, but then we could also support immediately multiple nodeppols ;) Maybe with an update? So that you could do a release now?

mysticaltech:
@ifeulner Definitely... Multiple node pools feel like a low-hanging fruit to me too; however, what about the node types? We could create the snapshot from cpx11 and in clout-init would expand the partitions automatically (or we could specifically do it).

If you can, go man! Let's move to the final format sooner rather than later! :) 🙏

mysticaltech:
@ifeulner You are fast my friend. Beautiful!! 🚀 🙏

mnencia:
I've found that it is complicated to upgrade from 5.10 because it misses some data, even with autoscaling disabled:

```
╷
│ Error: Invalid index
│
│   on ../terraform-hcloud-kube-hetzner/autoscaler-agents.tf line 6, in locals:
│    6:       cloudinit_config = base64encode(data.cloudinit_config.autoscaler-config[0].rendered)
│     ├────────────────
│     │ data.cloudinit_config.autoscaler-config is empty tuple
│
│ The given key does not identify an element in this collection value: the collection has no elements.
╵
╷
│ Error: Invalid index
│
│   on ../terraform-hcloud-kube-hetzner/autoscaler-agents.tf line 11, in locals:
│   11:       snapshot_id      = hcloud_snapshot.autoscaler_image[0].id
│     ├────────────────
│     │ hcloud_snapshot.autoscaler_image is empty tuple
│
│ The given key does not identify an element in this collection value: the collection has no elements.
╵
```

We need to provide a clean way to upgrade, at least if you do not enable the autoscaling feature with the upgrade.

mnencia:
It looks like the problem is that `autoscaler_nodepools` need to have at least one element. We should support having it undefined or empty to smoothly upgrade from previous versions.

mysticaltech:
@mnencia Thanks for your review and for pinpointing the problem above. @ifeulner FYI, we are very close to merging, but this needs to be sorted out first.

variables.tf
```diff
@@ -339,6 +356,29 @@ variable "extra_kustomize_parameters" {
   description = "All values will be passed to the `kustomization.tmp.yml` template."
 }
 
+variable "autoscaler_max_nodes" {
```
mysticaltech, line 359: @ifeulner All of these variables should be removed, no? I can't seem to find how they re-appeared.

autoscaler-agents.tf
```diff
@@ -0,0 +1,90 @@
+locals {
+  autoscaler_yaml = templatefile(
+    "${path.module}/templates/autoscaler.yaml.tpl",
+    {
+      #cloudinit_config - we have to check if this is necessary, if so we need to recreate it, or somehow extract it from server module, up to a higher level
+      cloudinit_config = base64encode(data.cloudinit_config.autoscaler-config[0].rendered)
```
mysticaltech, line 6: It seems we just need to add the `length(var.autoscaler_nodepools) > 0` condition at line 6 and 11. And then all should be good. Correct @ifeulner @mnencia ?

variables.tf
ifeulner: Yes, right, this change was missing in the commit. Done.

autoscaler-agents.tf
ifeulner: Hi, yes, actually one statement is enough.

ifeulner:
@mysticaltech @mnencia resolved the issues and also added code to clean the snapshot so that it is usable when done at a later stage. Also grows now `/var`partition according to the disk size of the used server type. 

Please check.

mnencia:
Thank you. I'll review the PR ASAP

variables.tf
```diff
@@ -80,6 +80,30 @@ variable "agent_nodepools" {
...
 
+variable "cluster_autoscaler_image" {
+  type        = string
+  default     = "k8s.gcr.io/autoscaling/cluster-autoscaler"
+  description = "Image of Kubernetes Cluster Autoscaler for Hetzner Cloud to be used."
+}
+
+variable "cluster_autoscaler_version" {
+  type        = string
+  default     = "v1.25.0"
```
mysticaltech, line 91: @ifeulner Why not use "latest" here? Because if we set this, we have to manually monitor and update the version.

mysticaltech: I understand it is safer, but I would prefer it breaks and user signal us, then to have to update this manually from time to time.

ifeulner: Because there is no "latest" in the official image distribution. And the reason for this is described [here](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler#releases) - depending on the used kubernetes version the corresponding cluster autoscaler version should be used.

mysticaltech: Ok, I see; thanks for the info!

mysticaltech:
On my end, all ready for ignition! @mnencia should we launch this baby into orbit? Waiting for your final review! 🚀

mysticaltech:
@mnencia I'll merge this for now, and we can fix it later on if you see something is still not right. That way you can take your time, thanks for catching the upgrade issue, this was so valuable! 🙏

PR #355, opened 2022-10-18
Update locals.tf

Maxon4eg:
Add 'IdentitiesOnly yes'  to ssh_args, helps me to fix "Too many authentication failures"

### DIFF
--- a/modules/host/locals.tf
+++ b/modules/host/locals.tf
@@ -3,7 +3,7 @@ locals {
   # For terraforms provisioner.connection.agent_identity, we need the public key as a string.
   ssh_agent_identity = var.ssh_private_key == null ? var.ssh_public_key : null
   # shared flags for ssh to ignore host keys for all connections during provisioning.
-  ssh_args = "-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no"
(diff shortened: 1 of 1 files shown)


phaer:
This is a good default IMHO, as we are always specifying a a key. I have it enabled globally in my ~/.ssh/config.

http://man.openbsd.org/ssh_config.5

> [IdentitiesOnly](http://man.openbsd.org/ssh_config.5#IdentitiesOnly)
    Specifies that [ssh(1)](http://man.openbsd.org/ssh.1) should only use the configured authentication identity and certificate files (either the default files, or those explicitly configured in the ssh_config files or passed on the [ssh(1)](http://man.openbsd.org/ssh.1) command-line), even if [ssh-agent(1)](http://man.openbsd.org/ssh-agent.1) or a PKCS11Provider or SecurityKeyProvider offers more identities. The argument to this keyword must be yes or no (the default). This option is intended for situations where ssh-agent offers many different identities.
```

mysticaltech:
@Maxon4eg A tiny PR would be appreciated for this, will merge ASAP.

mysticaltech:
Thanks @Maxon4eg, @phaer!; pretty helpful indeed. It will fail fast and not ban the IP!

PR #356, opened 2022-10-20
Support adding ssh keys from hcloud by label

WolfspiritM:
This allows setting a label selector to add the public ssh keys stored in the hetzner cloud to the provisioned machines.

I will update terraform docs later today.

@mysticaltech Would be great if you could take a look and maybe give it a try it if still works when no label is provided.

### DIFF
--- a/agents.tf
+++ b/agents.tf
@@ -9,11 +9,11 @@ module "agents" {
 
   name                       = "${var.use_cluster_name_in_node_name ? "${var.cluster_name}-" : ""}${each.value.nodepool_name}"
   base_domain                = var.base_domain
-  ssh_keys                   = [local.hcloud_ssh_key_id]
+  ssh_keys                   = length(var.ssh_hcloud_key_label) > 0 ? concat([local.hcloud_ssh_key_id], data.hcloud_ssh_keys.keys_by_selector[0].ssh_keys.*.id) : [local.hcloud_ssh_key_id]
   ssh_port                   = var.ssh_port
   ssh_public_key             = var.ssh_public_key
   ssh_private_key            = var.ssh_private_key
-  ssh_additional_public_keys = var.ssh_additional_public_keys
+  ssh_additional_public_keys = length(var.ssh_hcloud_key_label) > 0 ? concat(var.ssh_additional_public_keys, data.hcloud_ssh_keys.keys_by_selector[0].ssh_keys.*.public_key) : var.ssh_additional_public_keys
   firewall_ids               = [hcloud_firewall.k3s.id]
   placement_group_id         = var.placement_group_disable ? 0 : hcloud_placement_group.agent[floor(each.value.index / 10)].id
   location                   = each.value.location
--- a/control_planes.tf
+++ b/control_planes.tf
@@ -9,11 +9,11 @@ module "control_planes" {
 
   name                       = "${var.use_cluster_name_in_node_name ? "${var.cluster_name}-" : ""}${each.value.nodepool_name}"
   base_domain                = var.base_domain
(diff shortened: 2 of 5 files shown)


mysticaltech (approved):
@WolfspiritM Thanks for this, folks did ask about it in the past.

All looks good on my end, it's simple and I tested it, the "old" way still deploys like a charm!

mysticaltech:
@WolfspiritM I just realized the docs were already in, and it's simple enough to merge directly. It will ship with our upcoming 1.6 release, hopefully, this weekend.

mysticaltech:
## TEMPORARY MESSAGE ##

@WolfspiritM I've seen enough good things from you that it would be an honor to invite you to join the project. Work on this is voluntary, and it gives you commit rights and more influence to decide the future of this endeavor.

You are also welcome to join our private chat room where we sometimes discuss various aspects of the project, share ideas, or ask for input: https://matrix.to/#/#kube-hetzner:matrix.org

Cheers! ✨ 🍾

codeagencybe:
@mysticaltech I think you still don't understand my question. I can't make any PR to improve if I don't understand the usage of this first?

@WolfspiritM Can you please kindly elaborate how to use this feature if I only want to use the labels to add the ssh keys? 
I already have added my ssh keys in Hetzner in the project and added my label "role=admin", so that part is done. 

But how do I change the other settings in the kube.tf? 
I'm referring to this section;


```
  # * Your ssh public key
  ssh_public_key = file("/home/username/.ssh/id_ed25519.pub")
  # * Your private key must be "ssh_private_key = null" when you want to use ssh-agent for a Yubikey-like device authentification or an SSH key-pair with a passphrase.
  ssh_private_key = file("/home/username/.ssh/id_ed25519")

```  

Perhaps the documentation can be made more clear by calling explicit how these options play "together" if one wants to use only the label feature? 
I'm willing to make those changes to the documentation, but I first need to understand the scope of the features. 
So far still nobody explained to me which options need to be there and which not. 
The kube.tf file has several features that related to another one like eg nginx vs traefik. 
If one is set to true, the other one needs to be false as example. 
That's easy to understand for everybody

But what about the ssh keys? I see basically 3 options:
1. you either specificy the local path for public and private key (I don't want to use this option in my case)
2. you call the specific ID from a key from Hetzner cloud, so you can use existing SSH keys
3. you call the specific label from a key from Hetzner cloud, so you can use existing SSH Keys

2 and 3 are mentioned as commented lines and optional. This is clear. 
But what about 1? Is this minimal requirement even if I want to use the label feature? Do I always have to minimal specify 1 local SSH key?
Or does it need override as null if I want to use label? Null seems to be required only for yubikey but again, its not said how "not" to use it? 
Or does it need commented if I want to use label? 

If somebody can explain, then I can PR for improving the docs. 
Thanks

mysticaltech:
@codeagencybe You can use hetzner labels, but you also need one local to work with the remote_execs. So just use TF Cloud in local mode, and it will work. Many folks have confirmed in the past, that it does work.

codeagencybe:
@mysticaltech 
ok that was the answer i was looking for. So minimum 1 Local key is always required. 
But I will try with null also this evening and will update you on that one. 

I know local mode will work for sure, but this means I'm missing out on some collab features from TF cloud that I need. 
Changing to local is not the solution for my use case specific. 
But if a local ssh key is required in all cases, I still need to find a workaround via TF cloud to somehow let them inject my SSH key. 

Do you think there is a method possible to get rid of that requirement for 1 local ssh key?

mysticaltech:
@codeagencybe Null will not work as the key is needed for remote exec. Maybe try to research more how remote exec is done on TF Cloud without local mode.

Issue #357, opened 2022-10-20
Outputted `kubeconfig`: Abide to the `cluster_name` module input var


dimisjim:
https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/350#issuecomment-1285340912

PR #359, opened 2022-10-20
Outputted `kubeconfig`: Abide to the `cluster_name` module input var

dimisjim:
closes: https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/357


mysticaltech:
There are some issues in this PR. Please review my previous comment, and do not change the host, we need that. Instead, add the cluster_name.

Also, kubeconfig_parsed must yamlencode kubeconfig_final, as all versions need to be consistent.

Last but not least, changing "default" that way needs to be tested, as I understand that this is the name of the kube context, and your changes might not work.

Please apply the needed changes, and test it well, and get back to me.

mysticaltech:
@dimisjim Thanks for your changes, that looks fine now. But didn't you want to have `cluster_name` in `kubeconfig_data`? Just add another line for it if you want.

dimisjim:
Indeed it would be nice to have that too in the outputs :+1:

mysticaltech (approved):
Looking good now, thanks!  🙏 

And I also tried it and checked the naming of the context, user, etc, and it works great. "Default" was just not optimal.
 
ksnip_20221020-151147

PR #370, opened 2022-10-25
v1.6

mysticaltech:
This is the final PR for v1.6, with all the latest changes.

### DIFF
--- a/kubeconfig.tf
+++ b/kubeconfig.tf
@@ -13,18 +13,20 @@ data "remote_file" "kubeconfig" {
 
 locals {
   kubeconfig_external = replace(data.remote_file.kubeconfig.content, "127.0.0.1", var.use_control_plane_lb ? hcloud_load_balancer.control_plane.*.ipv4[0] : module.control_planes[keys(module.control_planes)[0]].ipv4_address)
-  kubeconfig_parsed   = yamldecode(local.kubeconfig_external)
+  kubeconfig_final    = replace(local.kubeconfig_external, "default", var.cluster_name)
+  kubeconfig_parsed   = yamldecode(local.kubeconfig_final)
   kubeconfig_data = {
     host                   = local.kubeconfig_parsed["clusters"][0]["cluster"]["server"]
     client_certificate     = base64decode(local.kubeconfig_parsed["users"][0]["user"]["client-certificate-data"])
     client_key             = base64decode(local.kubeconfig_parsed["users"][0]["user"]["client-key-data"])
     cluster_ca_certificate = base64decode(local.kubeconfig_parsed["clusters"][0]["cluster"]["certificate-authority-data"])
+    cluster_name           = var.cluster_name
   }
 }
 
 resource "local_sensitive_file" "kubeconfig" {
   count           = var.create_kubeconfig ? 1 : 0
-  content         = local.kubeconfig_external
+  content         = local.kubeconfig_final
   filename        = "${var.cluster_name}_kubeconfig.yaml"
   file_permission = "600"
 }
--- a/kubeconfig.tf
+++ b/kubeconfig.tf
@@ -15,7 +15,7 @@ locals {
   kubeconfig_external = replace(data.remote_file.kubeconfig.content, "127.0.0.1", var.use_control_plane_lb ? hcloud_load_balancer.control_plane.*.ipv4[0] : module.control_planes[keys(module.control_planes)[0]].ipv4_address)
   kubeconfig_parsed   = yamldecode(local.kubeconfig_external)
   kubeconfig_data = {
-    host                   = var.use_control_plane_lb ? hcloud_load_balancer.control_plane.*.ipv4[0] : local.kubeconfig_parsed["clusters"][0]["cluster"]["server"]
+    host                   = local.kubeconfig_parsed["clusters"][0]["cluster"]["server"]
     client_certificate     = base64decode(local.kubeconfig_parsed["users"][0]["user"]["client-certificate-data"])
     client_key             = base64decode(local.kubeconfig_parsed["users"][0]["user"]["client-key-data"])
     cluster_ca_certificate = base64decode(local.kubeconfig_parsed["clusters"][0]["cluster"]["certificate-authority-data"])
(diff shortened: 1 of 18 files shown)


mysticaltech:
@ifeulner @WolfspiritM @captnCC @mnencia @phaer I did a final simple test, and somehow the LB does not get assigned an IP. I tried multiple times and the CCM shows this error. 

I have yet to find out why, any help would be appreciated to finally drop this release.

```
1025 01:59:20.205168       1 event.go:294] "Event occurred" object="kube-system/traefik" fieldPath="" kind="Service" apiVersion="v1" type="Normal" reason="EnsuringLoadBalancer" message="Ensuring load balancer"
I1025 01:59:20.683242       1 load_balancer.go:496] "attach to network" op="hcops/LoadBalancerOps.attachToNetwork" loadBalancerID=912181 networkID=2144043
E1025 01:59:21.436745       1 controller.go:310] error processing service kube-system/traefik (will retry): failed to ensure load balancer: hcloud/loadBalancers.EnsureLoadBalancer: hcops/LoadBalancerOps.ReconcileHCLB: hcops/LoadBalancerOps.attachToNetwork: no subnet or IP available (no_subnet_available)
I1025 01:59:21.436797       1 event.go:294] "Event occurred" object="kube-system/traefik" fieldPath="" kind="Service" apiVersion="v1" type="Warning" reason="SyncLoadBalancerFailed" message="Error syncing load balancer: failed to ensure load balancer: hcloud/loadBalancers.EnsureLoadBalancer: hcops/LoadBalancerOps.ReconcileHCLB: hcops/LoadBalancerOps.attachToNetwork: no subnet or IP available (no_subnet_available)"
```

kubeconfig.tf
```diff
@@ -13,17 +13,20 @@ data "remote_file" "kubeconfig" {
 
 locals {
   kubeconfig_external = replace(data.remote_file.kubeconfig.content, "127.0.0.1", var.use_control_plane_lb ? hcloud_load_balancer.control_plane.*.ipv4[0] : module.control_planes[keys(module.control_planes)[0]].ipv4_address)
-  kubeconfig_parsed   = yamldecode(local.kubeconfig_external)
+  kubeconfig_final    = replace(local.kubeconfig_external, "default", var.cluster_name)
+  kubeconfig_parsed   = yamldecode(local.kubeconfig_final)
   kubeconfig_data = {
-    host                   = var.use_control_plane_lb ? hcloud_load_balancer.control_plane.*.ipv4[0] : local.kubeconfig_parsed["clusters"][0]["cluster"]["server"]
```
mysticaltech, line 18: This shouldn't have been removed 🤔

mysticaltech:
Ah, the weird thing is that I also get that with v1.5.10. Either Hetzner has an issue tonight, or the latest CCM is causing problems.

My terraform config is as follows:

```
Terraform v1.2.8
on linux_amd64
+ provider registry.terraform.io/hashicorp/cloudinit v2.2.0
+ provider registry.terraform.io/hashicorp/local v2.2.3
+ provider registry.terraform.io/hashicorp/null v3.1.1
+ provider registry.terraform.io/hashicorp/random v3.4.3
+ provider registry.terraform.io/hetznercloud/hcloud v1.35.2
+ provider registry.terraform.io/integrations/github v5.6.0
+ provider registry.terraform.io/tenstad/remote v0.1.1

Your version of Terraform is out of date! The latest version
is 1.3.3. You can update by downloading from https://www.terraform.io/downloads.html
```

mysticaltech:
The error is probably because of on-going maintenance on LBs and the Cloud APIs.

kubeconfig.tf
WolfspiritM: This didn't work without that change (I forgot about that fix but it's actually a fix).

The reason is that two lines above it already gets replaced.
```  kubeconfig_external = replace(data.remote_file.kubeconfig.content, "127.0.0.1", var.use_control_plane_lb ? hcloud_load_balancer.control_plane.*.ipv4[0] : module.control_planes[keys(module.control_planes)[0]].ipv4_address)```
https://127.0.0.1:6443 -> https://lb-ip:6443


Then that is used to get the yaml parsed version
```kubeconfig_parsed   = yamldecode(local.kubeconfig_final)```

And from there we get the host correctly which is then e.g. ```https://lb-ip:6443```

In the previous state the host was the loadbalancer IP without any protocol or port which didn't work.

**Review by ifeulner — APPROVED** [review]

**Review by karbowiak — APPROVED** [review]
Looks good to me

.. Altho, don't take my approval as gospel - my terraform knowledge is hilariously basic 😅

File: templates/autoscaler.yaml.tpl [hunk 2]
```diff
@@ -0,0 +1,199 @@
...
+            requests:
+              cpu: 100m
+              memory: 300Mi
+          command:
+            - ./cluster-autoscaler
+            - --v=5
+            - --cloud-provider=hetzner
+            - --stderrthreshold=info
+            %{~ for pool in node_pools ~}
+            - --nodes=${pool.min_nodes}:${pool.max_nodes}:${pool.server_type}:${pool.location}:${pool.name}
```
captnCC, line 173: I think it would be a good idea to prefix the pool name with the cluster name to make the better identifiable.
Something like this:
```
- --nodes=${pool.min_nodes}:${pool.max_nodes}:${pool.server_type}:${pool.location}:${clustername}-${pool.name}
```

captnCC:
I tested the autoscaler and SSH labels, both working like a charm! 🤩
Besides the mentioned autoscaler node name, I don't see any hold backs!

templates/autoscaler.yaml.tpl
ifeulner: Good catch. Should we bring this into the release?

kubeconfig.tf
mysticaltech: Ok, thanks for the clarification @WolfspiritM 🙏🙏

templates/autoscaler.yaml.tpl
mysticaltech: @ifeulner I completely missed @captnCC's comment. Please PR. I will do a patch release right away.

PR #374, opened 2022-10-26
Helm - Replace local `value.yaml` files with variables/locals

nupplaphil:
I'm currently using this repository as a sub-module of my whole cluster-configuration and use the GitOp pattern for it.

Therefore I do think that we shouldn't use local *.yaml files at the root of this repo for configuration, because one must not commit it to this upstream (of course), but on the other hand, we cannot manage these values anywhere else ..

So I created a `*_values` variable for each Helm configuration, we do have so far and adapted the `locals` of them to either use the default setup, or use the content of each `*_values` variable.

For backward compatibility, one can "dump" the current, local file into the corresponding `*_value`, so it shouldn't make any difference.

ToDos:
- [x] Updating the `README`


variables.tf
```diff
@@ -250,6 +262,18 @@ variable "cni_plugin" {
   }
 }
 
+variable "cilium_values" {
+  type        = string
+  default     = ""
+  description = "Additional helm values file to pass to Cilium as 'valuesContent' at the HelmChart."
+}
+
+variable "calico_values" {
```
mysticaltech, line 271: @nupplaphil Currently, Calico does not use a helm chart cause I did not find a straightforward one; if you can replace the current manifest with a helm chart, it would be great!

mysticaltech:
@nupplaphil A lot cleaner, man; I love it! 🙏

mysticaltech:
@nupplaphil Is the read me the only thing left? Did you test everything?

nupplaphil:
> @nupplaphil Is the read me the only thing left? Did you test everything?

I successfully used it for creating my environment with a mix of customized values and standard values

mysticaltech:
@captnCC @WolfspiritM @ifeulner If you could take a quick look it would be awesome! Would love to hear what you think of that one. Personally, I like it, it cleaner.

If all good, I will just complete the readme part myself and we can merge!

ifeulner (approved):
sorry - was too fast.

locals.tf
```diff
@@ -337,10 +337,48 @@ controller:
       "load-balancer.hetzner.cloud/uses-proxyprotocol": "true"
   EOT
 
-  default_rancher_values = <<EOT
+  traefik_ingress_values = var.traefik_ingress_values != null ? var.traefik_ingress_values : <<EOT
+    globalArguments: []
+    service:
+      enabled: true
+      type: LoadBalancer
+%{if !local.using_klipper_lb~}
```
ifeulner, line 345: Ah I was to fast... this is not the same than the check previously `{ if using_hetzner_lb ~}`
at least for me the lb gets not the right name when using traefik.

ifeulner (changes requested):
see my comment

locals.tf
nupplaphil: Hmmm Strange I copied exactly the condition, are there currently any traefik issues with the current stable?

mysticaltech: @nupplaphil No there are not any known issues! Good catch @ifeulner.  Philipp, may you just need to pull the latest?!

mysticaltech:
@nupplaphil This is too good to stay on the shelf just for one line, but I would prefer it came from you, please! 🙏

locals.tf
nupplaphil: But it's effectly the same, see
https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/374/files#diff-1366caed88e51e22bcd5a911420820bee3495d781ecccd85f25e00170fc325b8L140
.. I just use the variable instead of the mapped variable ..

nupplaphil:
I updated the README including some minor fixings inside the README. I'm finished now, except the question about the variable :-/

locals.tf
mysticaltech: Ah yes, it is indeed the same! When you look at the link above @ifeulner, you'll understand. Thanks @nupplaphil!

mysticaltech:
Awesome @nupplaphil, thanks for that! Will merge and release tonight.

mysticaltech:
@nupplaphil Turns out  @ifeulner was right, traefik was not deploying, but it was fixed in #392. However, great PR man, thanks again! 🚀✨

mysticaltech:
Basically, there was two things, the test of equality to `!= null` was replaced by `!= ""`, and the spacing of the yaml object for traefik was wrong, some `~}` were also messing things up (and causing helm to fail), and some comments `# ...` (probably not important that last one).

PR #375, opened 2022-10-27
fix: support for labels, that are defined in `agent_nodepools`

boemjay:
Previously the labels defined in `agent_nodepools` were not applied when running `terraform apply`.
Now they are properly parsed and added as labels to the nodes. (either as "label=value" or "label").

```hcl
agent_nodepools = [
  {
    name        = "agent-small",
    server_type = "cpx41",
    location    = "fsn1",
    labels      = ["label_without_value", "label=with_value"], // labels control applied firewall rules
    taints      = [],
    count       = 1
  }
]
```

### DIFF
--- a/agents.tf
+++ b/agents.tf
@@ -24,7 +24,11 @@ module "agents" {
 
   private_ipv4 = cidrhost(hcloud_network_subnet.agent[[for i, v in var.agent_nodepools : i if v.name == each.value.nodepool_name][0]].ip_range, each.value.index + 101)
 
-  labels = merge(local.labels, local.labels_agent_node)
+  labels = merge(
+    local.labels,
+    local.labels_agent_node,
+    tomap({ for label in each.value.labels: split("=", label)[0] => length(split("=", label)) > 1 ? split("=", label)[1] : ""})
+  )
 
   automatically_upgrade_os = var.automatically_upgrade_os
 


boemjay:
Btw. according to the contributing guidelines, a pull request to `staging` should be opened, but there is no `staging` branch😅

mysticaltech:
This PR looks good @boemjay; thanks for finding out and fixing that!

Two things so we can move on, please, do test it! And also, run terraform validate, and fix the errors (or just use the terraform vscode extension that will do it all for you automatically)! 🙏

captnCC:
A question on the side about the label usage, are they applied to both the hcloud servers and the cluster nodes?

mysticaltech:
@captnCC Just the nodes! Now, maybe the Hetzner CCM can sync those, that I do not know. There is a way to configure it via the ccm_manifest template.

captnCC:
@mysticaltech As far as I'm aware, the ccm doesn't support the "back sync". I already looked into the overall label usage for resources created by the ccm, csi and autoscaler, but sadly none of these have support to add labels.

mysticaltech:
Ok, @captnCC, thanks for the info!

ifeulner:
Ah, this fix applies now the labels on the Hetzner servers, not to the kubernetes nodes (this worked already). Actually, is this wanted / needed?

mysticaltech:
@ifeulner No, it applies to the Kube labels if fact. According to @boemjay they did not work before. Which I must admit was a surprising statement to me too. 

What was the format you used @boemjay? Just paste in an example, please.

mysticaltech:
@captnCC I was wrong; @ifeulner is right. It applies the agent nodepools Kube labels to the Hetzner server labels, which is not really needed, indeed!

@boemjay Why do you need that? Actually, they do work but are added as labels WITHIN Kubernetes, not at the Hetzner level. There they are not needed!
 
To see them, just run (with the current vanilla release), `kubectl describe <node-name>`.

boemjay:
Thanks for the explanations, it makes a lot of sense that these are node labels on the kubernetes level.
I should have mentioned that I was talking about labels on the Hetzner level.

I needed the labels in order to apply custom firewall rules, e.g. to open MQTT ports only on the worker nodes where I schedule my MQTT Broker. 
But actually it doesn't quite work yet, since terraform throws an error when firewall rules are applied by Hetzner label policies.

I will investigate this further and set the PR to `Draft` until then.

mysticaltech:
Ok, thanks for clarifying the use case @boemjay! And if you will need this in the end, just create another attribute called "extra_hetzner_lables", for instance.

mysticaltech:
@boemjay Will close this for now as it's outdated, but please do not hesitate to submit another one in the future if needed. Thanks!

mysticaltech:
@maggie44 Please don't hesitate to open a new PR, you can inspire yourself from this one. If you need this, maybe other will need it too.

PR #398, opened 2022-11-04
fix: use same datatype for default tf example

quickstar:
fixes an issue where an initial create was not possible is the default kube.tf.example was just used

### DIFF
--- a/kube.tf.example
+++ b/kube.tf.example
@@ -107,7 +107,10 @@ module "kube-hetzner" {
       server_type = "cpx11",
       location    = "fsn1",
       labels      = [],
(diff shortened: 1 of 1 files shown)


mysticaltech:
@quickstar I agree it's better to have this taint to exclude it from longhorn, but it should not be necessary. What may have happened is that the nodes were not powerful enough and running out of memory. What do you think?

I will merge this as it's a good default!

quickstar:
> What may have happened is that the nodes were not powerful enough and running out of memory.

I don't think so because terraform failed during validation, hence there were no resource created.

PR #407, opened 2022-11-09
etcd snapshots backup to S3 storage

ifeulner:
Hi, this is an initial implementation to support snapshots to S3 storage. 
Done by passing simply the needed args to config.yaml. 
What do you think?


otavio:
The only missing bit would be the docs about the process to restore the cluster.

mysticaltech (approved):
@ifeulner LGTM, good work! I believe it's already tested. So good for merging! 🚀 Will do a minor release afterward.

mysticaltech:
@otavio @ifeulner We can just link to https://docs.k3s.io/backup-restore.

mysticaltech:
@ifeulner FYI I added the above link in kube.tf.example.

Issue #421, opened 2022-11-16
terraform output --raw kubeconfig_file > Error: Output "kubeconfig_file" not found


wombelix:
Hi,

I set `create_kubeconfig = false` and tried to get the kubeconfig file afterwards by running `terraform output --raw kubeconfig_file` as described in the README and the kube.tf.example, which throws the error `Error: Output "kubeconfig_file" not found`, the Output get registered as `kubeconfig` and not `kubeconfig_file`, so running `terraform output --raw kubeconfig` works as expected.

I think either the example file or the Readme should be adjusted? I can create a PR for this, just want to confirm that my assumption is correct.

Cheers
Dominik

mysticaltech:
@wombelix You are right! PR welcome, will merge ASAP.

mysticaltech:
@wombelix Your fix was good, thanks! But for backward compatibility decided just to change the docs, the new command is indeed `terraform output --raw kubeconfig`  as you found out above.

PR #422, opened 2022-11-16
fix: output kubeconfig renamed to kubeconfig_file

wombelix:
Align output name in kube.tf.example with existing Documentation

fixes issue https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/421

### DIFF
--- a/kube.tf.example
+++ b/kube.tf.example
@@ -437,7 +437,7 @@ terraform {
   }
 }
 
-output "kubeconfig" {
+output "kubeconfig_file" {
   value     = module.kube-hetzner.kubeconfig_file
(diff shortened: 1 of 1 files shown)


ifeulner:
Actually, this also does not work "out-of-the-box" if used as a module - as there are no outputs then defined. We need to adjust the documentation here.

mysticaltech:
Actually ingo it does, as the kube.tf comes with output definition.  I tested it. 

However, @wombelix, this PR is correct, but instead I will just change to doc to point to kubeconfig instead of kubeconfig_file, for backward compatibility. But thank you for this!

PR #426, opened 2022-11-19
Support custom configuration options for control planes

### LINKED ISSUE #408: [Enhacnement] Custom K3S config parameters
Would it be possible to add custom config parameters to the K3S config.yaml for the control planes? 
Much like the option to have additional Traefik parameters, I'm running this in production and I needed etcd/kube- controller, proxy and scheduler monitoring therefore I added these extra parameters to `/etc/rancher/k3s/config.yaml` for each control plane:
```
"etcd-expose-metrics": "true"
"kube-controller-manager-arg":
- "bind-address=0.0.0.0"
"kube-proxy-arg":
- "metrics-bind-address=0.0.0.0"
"kube-scheduler-arg":
- "bind-address=0.0.0.0"

```

This has the catch that I can't auto redeploy with the matrics exposed and it is a bit daunting needing to add those manually each time

Would that be a possibility or does this also fall into the  "We try not to support Third party X thing" category? If so then not a problem,  I will just stick with it :| or add it myself

11Tuvork28:
Closes #408 
Adds ability to have to custom config.yaml entries for the control planes, I'm not exactly sure if everything works.
`terraform validate` returns no errors, so please correct me if I made something wrong! I haven't actually tested it yet though

### DIFF
--- a/kube.tf.example
+++ b/kube.tf.example
@@ -144,9 +144,9 @@ module "kube-hetzner" {
   # kube-controller-manager-arg = "- bind-address=0.0.0.0",
   # kube-proxy-arg = "- metrics-bind-address=0.0.0.0",
   # kube-scheduler-arg = "- bind-address=0.0.0.0",
-  control_planes_custom_config = {
-
-  }
+  # control_planes_custom_config = {
+  #  
+  # }
 
   # * LB location and type, the latter will depend on how much load you want it to handle, see https://www.hetzner.com/cloud/load-balancer
   load_balancer_type     = "lb11"
--- a/kube.tf.example
+++ b/kube.tf.example
@@ -139,11 +139,11 @@ module "kube-hetzner" {
     }
   ]
   # Add custom control plane configuration options here.
-  # E.g:
-  # etcd-expose-metrics = true,
-  # kube-controller-manager-arg = "- bind-address=0.0.0.0",
-  # kube-proxy-arg = "- metrics-bind-address=0.0.0.0",
-  # kube-scheduler-arg = "- bind-address=0.0.0.0",
+  # E.g to enable monitoring for etcd, proxy etc:
+  # "etcd-expose-metrics": "true"
+  # "kube-controller-manager-arg": "bind-address=0.0.0.0"
+  # "kube-proxy-arg": "metrics-bind-address=0.0.0.0"
+  # "kube-scheduler-arg": "bind-address=0.0.0.0"
   # control_planes_custom_config = {
   #  
   # }
--- a/kube.tf.example
+++ b/kube.tf.example
@@ -138,6 +138,15 @@ module "kube-hetzner" {
       # longhorn_volume_size = 20
     }
   ]
+  # Add custom control plane configuration options here.
+  # E.g:
+  # etcd-expose-metrics = true,
+  # kube-controller-manager-arg = "- bind-address=0.0.0.0",
+  # kube-proxy-arg = "- metrics-bind-address=0.0.0.0",
+  # kube-scheduler-arg = "- bind-address=0.0.0.0",
+  control_planes_custom_config = {
+
+  }
 
   # * LB location and type, the latter will depend on how much load you want it to handle, see https://www.hetzner.com/cloud/load-balancer
   load_balancer_type     = "lb11"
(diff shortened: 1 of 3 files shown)


kube.tf.example
```diff
@@ -138,6 +138,15 @@ module "kube-hetzner" {
...
+  # E.g:
+  # etcd-expose-metrics = true,
+  # kube-controller-manager-arg = "- bind-address=0.0.0.0",
+  # kube-proxy-arg = "- metrics-bind-address=0.0.0.0",
+  # kube-scheduler-arg = "- bind-address=0.0.0.0",
+  control_planes_custom_config = {
+
+  }
 
   # * LB location and type, the latter will depend on how much load you want it to handle, see https://www.hetzner.com/cloud/load-balancer
```
mysticaltech, line 151: @11Tuvork28 Please comment this line and we're good! 🚀

mysticaltech (commented):
@11Tuvork28 You've tested it, right? Works well?

kube.tf.example
mysticaltech: Please also run `terraform format`, or install the terraform extension in vscode, save, and push again. To check, `terraform validate` must pass.

mysticaltech: Lines 147-149

11Tuvork28: About the testing, I can't since I have maxed my limits on Hetzner and I can't or better said don't want to scale down my production cluster. But it should work as long as its corrently formatted into yaml which I don't really know how to check

mysticaltech: Ok,  I will test it, thanks.

mysticaltech: @11Tuvork28 You could just open a support request to Hetzner and they will increase it, it's quite fast. I did it in the past. It's important that you test this code yourself, with your use case, meaning for the reason that you wanted to build this functionality. When that is done and you confirm that all works on your side, I will do one last test myself and we can merge.

11Tuvork28: Sorry for the delayed answer, took a bit to get my limits raised plus work. Anyways I tested it:
kube.tf
```yaml
   etcd-expose-metrics = true,
   kube-controller-manager-arg = "bind-address=0.0.0.0",
   kube-proxy-arg ="metrics-bind-address=0.0.0.0",
   kube-scheduler-arg = "bind-address=0.0.0.0",
  }
```
/etc/rancher/k3s/config.yaml
```yaml
"etcd-expose-metrics": "true"
"kube-controller-manager-arg": "bind-address=0.0.0.0"
"kube-proxy-arg": "metrics-bind-address=0.0.0.0"
"kube-scheduler-arg": "bind-address=0.0.0.0"
```

So it works as expected but its just for that usecase, I think array values would be a problem though but I'm not sure

mysticaltech: Thanks for pushing the envelope @11Tuvork28! Still, it's better to have this than not. @ifeulner What do you think, we merge huh? :)

kube.tf.example
```diff
@@ -139,11 +139,11 @@ module "kube-hetzner" {
     }
   ]
   # Add custom control plane configuration options here.
-  # E.g:
-  # etcd-expose-metrics = true,
-  # kube-controller-manager-arg = "- bind-address=0.0.0.0",
-  # kube-proxy-arg = "- metrics-bind-address=0.0.0.0",
-  # kube-scheduler-arg = "- bind-address=0.0.0.0",
+  # E.g to enable monitoring for etcd, proxy etc:
```
mysticaltech, line 142: @11Tuvork28 From my experience, those " " are added by yamlencode automatically. Are you sure that you used those in your testing?

mysticaltech: This was the exact format that you tested, correct? Just want to be 100% sure.

11Tuvork28:
What I exactly tested is what I posted in the above comment. I tested exactly this, I updated the comment in kub.tf.example in my latest commit
\`\`\`
etcd-expose-metrics = true, kube-controller-manager-arg = "bind-address=0.0.0.0",
kube-proxy-arg = "metrics-bind-address=0.0.0.0", kube-scheduler-arg = "bind-address=0.0.0.0",

\`\`\`

mysticaltech:
@11Tuvork28 Thanks for this, it will get released later tonight, you can subscribe to the releases to get a notification! 🙏 ✨

PR #432, opened 2022-11-21
Added option to configure private registries.

s3rius:
Hi and thanks for amazing project. 

I needed to connect my cluster to existing registries, but I couldn't find an option to do so. 
This is my attempt to create such option.

Because if you change registries.yaml contents later we need to reload all k3s services on all nodes.

> Why do new resources depend on kustomization resource?

Because otherwise, when we create cluster for the first time and you provide custom registry config, the error may occur.
Since we need to restart k3s services after updating the config, it may create a conflict where kustomization target and new targets try to interact with k3s service at the same time.

I would really appreciate a feedback for this PR.

Signed-off-by: Pavel Kirilin <win10@list.ru>

### DIFF
--- a/kube.tf.example
+++ b/kube.tf.example
@@ -266,6 +266,21 @@ module "kube-hetzner" {
   # Whether to use the cluster name in the node name, in the form of {cluster_name}-{nodepool_name}, the default is "true".
   # use_cluster_name_in_node_name = false
 
+  # Extra k3s registries. This is useful if you have private registries and you
+  # want to pull images without additional secrets. 
+  # registries.yaml file docs: https://docs.k3s.io/installation/private-registry
+  /* k3s_registries = <<-EOT
+    mirrors:
+      hub.my_registry.com:
+        endpoint:
+          - "hub.my_registry.com"
+    configs:
+      hub.my_registry.com:
+        auth:
+          username: username
+          password: password
+  EOT */
+
   # Adding extra firewall rules, like opening a port
   # More info on the format here https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs/resources/firewall
   # extra_firewall_rules = [
--- a/agents.tf
+++ b/agents.tf
@@ -21,6 +21,7 @@ module "agents" {
   ipv4_subnet_id             = hcloud_network_subnet.agent[[for i, v in var.agent_nodepools : i if v.name == each.value.nodepool_name][0]].id
   packages_to_install        = local.packages_to_install
   dns_servers                = var.dns_servers
+  k3s_registries             = var.k3s_registries
 
   private_ipv4 = cidrhost(hcloud_network_subnet.agent[[for i, v in var.agent_nodepools : i if v.name == each.value.nodepool_name][0]].ip_range, each.value.index + 101)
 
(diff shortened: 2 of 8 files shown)


kube.tf.example
```diff
@@ -257,6 +257,21 @@ module "kube-hetzner" {
   # Whether to use the cluster name in the node name, in the form of {cluster_name}-{nodepool_name}, the default is "true".
   # use_cluster_name_in_node_name = false
 
+  # Extra k3s registries. This is useful if you have private registries and you
+  # want to pull images without additional secrets. 
+  # registries.yaml file docs: https://docs.k3s.io/installation/private-registry
+  # k3s_registries = <<-EOT
```
mysticaltech, line 263: @s3rius Here, just to make it consistent with the helm values variables at the bottom of the file, and to remove potential indentation errors from users, can we change this to block comments (instead of line comments)?

mysticaltech (changes requested):
@s3rius Looking good to me (just one tiny change requested)! It's tried and tested, correct?

kube.tf.example
s3rius: Yeah. I tested it, it works as intended.

s3rius: Ok, I changed comment styles.

s3rius:
@mysticaltech, sorry, I requested rereview from you and it accidentally removed @ifeulner from reviewers.

mysticaltech (approved):
Awesome! 🚀

ifeulner:
Hi @s3rius thanks for contributing  - awesome! Just to get it right - the idea is to add/adjust the registry settings AFTER the cluster has been created, right? And you're restarting the node after applying the new setting. As there is no "drain" before, there will be at least possibilities to corrupt data, I think... 
Also the autoscaler nodes are not getting the new registry setting, this is another problem. 

So another alternative would be to set the registry settings when a node is created, and only then. This setting could also be used when the cloud-init for the autoscaler nodes is prepared.
But no change afterwards possible anymore...  but with the creation of new nodes a new registry setting could also be applied.

So I see some drawbacks with both ways ... what do you think @mysticaltech?

s3rius:
That's a good take. But reboot is necessary after updating registry.yaml. Also, I'm not sure if it's possible to update autoscaled nodes. Maybe adding it to cloudconfig would be a nice solution. But still, I'm not sure how to drain and reboot nodes after updates.

mysticaltech:
@ifeulner Right on spot, I completely missed the need for draining. 

But there's an easy way around that issue. @s3rius, just do the exact same thing for the registry file as is done for the k3s config file, which is first created in `/tmp/config.yaml` (see https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/b8dcab2bc7b5fc1a7b16b596364c9ecb4df76274/agents.tf#L52), and then moved to the right k3s folder in   common_commands_install_k3s (see https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/b8dcab2bc7b5fc1a7b16b596364c9ecb4df76274/locals.tf#L14).

That way, registries.yaml will follow the exact same path as config.yaml. And this will work as the [k3s docs](https://docs.k3s.io/installation/private-registry) says that _Upon startup, K3s will check to see if a registries.yaml file exists at /etc/rancher/k3s/ and instruct containerd to use any registries defined in the file._.

mysticaltech:
@s3rius If you need any help, please let me know, this is normally quite straightforward.

s3rius:
Thanks for offering help. I'm just chilling. I'll do it next week.

s3rius:
Hi, everyone. I rewrote this pull request using cloud-config as you told me. It works.

mysticaltech:
@s3rius Even better! Perfect. @ifeulner Any final comments before we merge?

ifeulner (approved):
LGTM for now!

mysticaltech:
Just released, thanks @s3rius!

Issue #433, opened 2022-11-22
Bare Metal Support


maaft:
Did anyone successfully integrated hetzner bare metal servers into their `kube-hetzner` cluster?

Bare-Metal servers are cheap and ideal for long running, resource-hungry tasks like databases.

Following steps are needed to make this work:
(`x` means that the step is absolutely possible currently) 
- [x] Install MicroOS on Bare-Metal server
- [x] configure hetzner VPN access through [vSwitch](https://docs.hetzner.com/cloud/networks/connect-dedi-vswitch/)
- [x] Install k3s
- [x] Join Cluster 
- [ ] Configure `kube-hetzners` CCM to retrieve IPs from bare-metal servers for LB configuration

The last point is currently missing. But I already successfully mixed Hcloud servers with Hetzner Bare Metal servers by using different Hetzner CCM than that from the official repo.

**Questions**:
- What CCM is this `kube-hetzner` using?
- Is it possible to use our own CCM (e.g. this  [one](https://github.com/syself/hetzner-cloud-controller-manager) here with hetzner bare metal support)

mysticaltech:
@maaft We use the official CCM https://github.com/hetznercloud/hcloud-cloud-controller-manager, the other is a fork of this one (maybe modified). 

May that previous discussion is related, please have a look https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/discussions/283.

Also, if you find a way, PRs are most welcome, as long as you test it well on your end.

@phaer You may have more ideas on this! 🙏

ifeulner:
@maaft I think that should work already, if attached via a `vswitch` to the private network of the cloud servers? Or do I miss something?

maaft:
@ifeulner the official CCM does not support bare metal servers (i.e. robot api). Which makes provisioning LBs with bare metal server IPs impossible currently.

mysticaltech:
@maaft In that case please do not hesitate to submit a PR that adds the ability to change the CCM, as long that everything keeps on working.

maaft:
@mysticaltech will do! Could you just kindly point me to the location where the current CCM is being configured/setup?

mysticaltech:
[...]

PR #440, opened 2022-12-02
Feature/ccm override

maaft:
This PR allows to configure a custom CCM.
Also, it allows to add custom secrets during init-phase that might be needed for the CCM to start working.

### DIFF
--- a/init.tf
+++ b/init.tf
@@ -224,8 +224,11 @@ resource "null_resource" "kustomization" {
   provisioner "remote-exec" {
     inline = [
       "set -ex",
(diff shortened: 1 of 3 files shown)


mysticaltech:
Thank you very much for your effort and research @maaft, I know this is just part of the logic, but for now, even though it is ok, it does not appear to be needed without the rest of the logic that would include the private network. Please see my comment on the corresponding issue here https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/433#issuecomment-1335518237.

mysticaltech:
@kube-hetzner/core Personally, I think it would add too much complexity to support metal hetzner servers at this point. But at least @maaft proved that this was doable.

Just FYI for now, am merging into its own branch just to back up the work done. 

@maaft If the future you feel inspired to continue working on this, please feel free to use new branch created for it. If it gets simple and maintainable enough, we could eventually merge it into master.

Issue #436, opened 2022-11-26
Close SSH and kube api to the world


danackerson:
Is it possible to lock down the SSH and kube api access to the world? Overriding files works until you run a `terraform init --upgrade` then all files under .terraform are discarded and recreated.

ifeulner:
What do you mean with "lock down"? What about just adjusting the firewall rules if you want to avoid access?

danackerson:
[`locals.tf`](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/2acaf6c259da6c2f863cedfda89b620aa4cbede4/locals.tf#L142) explicitly opens SSH to the world. Overriding `base_firewall_rules` works with an *_override.tf file parallel to it, but `terraform init --upgrade` discards and refetches all the files from this repo effectively eliminating the override.

Otherwise, I wasn't able to see how (in hcloud terraform) to explicitly *deny* ports? Any other suggestions would be much appreciated!

mysticaltech:
@danackerson My approach would be to, first deploy with the default firewall state in kube.tf (for instance). When your cluster is up, change those firewall rules in your kube.tf and apply! That would just update the Hetzner firewall rules and you would have reached the state you want. Please try it!

danackerson:
But those base firewall rules are in the [`locals.tf`](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/2acaf6c259da6c2f863cedfda89b620aa4cbede4/locals.tf#L142) file - not my kube.tf?

If I change them in locals.tf, they'll just be overridden by the next `init --upgrade` cmd.

andinger:
I think, what @mysticaltech means is, to set the `base_firewall_rules` from `locals.tf` explicitly in the `kube.tf` after you created your cluster initially. The ports are necessary for the initial boostrap and any changes to be applied in your cluster lifecycle. Nevertheless you can open the ports manually if you want to maintain your cluster.

mysticaltech:
@danackerson @andinger Nope! Sorry, I should have made it more obvious. You have an `extra_firewall_rules`, please see the `kube.tf.example` file. This is what you need to change after the cluster is up. 

And why not, also set the `ssh_port` from the get-go (before the first apply) to something other than 22, ideally > 1000. That way, you can later block that with the `extra_firewall_rules` variable, after the cluster has successfully deployed (so with a second terraform apply).

[...]

PR #443, opened 2022-12-05
WIP: Support RKE2

WolfspiritM:


xrow:
I would like to try this once it is merged....

mysticaltech:
@xrow This is in the pipes! Hopefully sooner rather than later.

control_planes.tf
```diff
@@ -71,6 +73,15 @@ resource "hcloud_load_balancer_service" "load_balancer_service" {
   listen_port      = "6443"
 }
 
+resource "hcloud_load_balancer_service" "load_balancer_service_rke2" {
```
mysticaltech, line 76: @WolfspiritM Do you remember what's the purpose of this?

mysticaltech:
@captnCC Hope all is good man, long time no see. If you have bandwidth to take this on, it would be wonderful, please don't hesitate. For the rest, all is being sorted out, see the other active PR.

control_planes.tf
WolfspiritM: RKE2 uses port 9345 to handle cluster joins. This adds the loadbalancer service for that port.

mysticaltech: Thanks, @WolfspiritM, if you find the courage to finish this PR one day, it would be awesome!

mysticaltech:
@kube-hetzner/core Who wants to take this on? This repo is just for inspiration, but @WolfspiritM gave us a beautiful POC.

Both k3s and rke2 were built and are maintained by the same team over at Rancher! They are extremely similar in how they work, lot of parallels, it should be quite easy. However it does require a good amount of work and testing.

My advice, if one of you is eager to take this on, create a new branch for it, and let's go. I will help you along the way and we can make this awesome!

Supporting docs:

- https://docs.k3s.io/installation/configuration
- https://docs.rke2.io/install/configuration

valkenburg-prevue-ch:
What are the pain points?

ifeulner:
I can work on this, but what happened to the better kubernetes integration @valkenburg-prevue-ch ? :)

valkenburg-prevue-ch:
I must have forgotten what I said weeks or months ago, because I'm not sure what you mean. Could you explain? 🧐

mysticaltech:
@valkenburg-prevue-ch No paint points for me personally, just that it seems rke2 has better security, seing they are very similar, it seems like a low hanging fruit to support deploying both in this project. I will bring us closer to having something extremely solid, as rke2 is government level secure and enforces best practices.

mysticaltech:
@ifeulner If you feel inspired please take this on!! 🙏 🚀

valkenburg-prevue-ch:
> @valkenburg-prevue-ch No paint points for me personally, ...

I meant more: what is needed, what needs to be fixed/checked/tested? I mean, I'm eager to adopt RKE2, but my current priorities are elsewhere.

zifeo:
I also would love to see rke2 support for Hertzner but do not have the bandwidth currently. Happy to help review though or to bring some inspirations over from the OpenStack [version](https://github.com/zifeo/terraform-openstack-rke2) I created aside. Let me know!

valkenburg-prevue-ch:
I've tangentially been looking into this (not putting in serious work), and I think that the selinux problem which we had with k3s which was later resolved by an update on k3s-selinux, has not yet been fixed for rke2-selinux. See https://github.com/rancher/rke2-selinux (no updates since november) and https://bugzilla.opensuse.org/show_bug.cgi?id=1207836 (issue is the same for k3s and rke2).

So, at this moment, rke2 on microos would only work without full selinux setup.

valkenburg-prevue-ch:
And, by the way, the minimum hardware for a rke2 node is considerably larger than k3s:

## k3s 
Recommended 2 CPUs and 1GB RAM (https://docs.k3s.io/installation/requirements#hardware), that's a CPX11 Hetzner server.

## rke2
Recommended 4CPUs and 8GB RAM (https://docs.rke2.io/install/requirements#hardware), that's a CPX31 Hetzner server, a bit over 3x the price of a CPX11.

mysticaltech:
@valkenburg-prevue-ch Very good points! I had no idea they were behind on the SELinux issue, that is a bad sign, it means it's used a lot less.

About the node requirements, I was aware it was more, but not that much more.

In that new light, I could live without rke2, @ifeulner, the ball is in your camp. If you still want to make it work, we could probably fix the `rke2_selinux` package ourselves based on the changes done on the k3s one, I already compiled the latter RPM once and it's not sorcery.

ifeulner:
@valkenburg-prevue-ch good catch with the selinux topic... and a pity that it not has been fixed yet.
Regarding node requirements: Yes, was aware of that, but to be honest, none of our cluster has nodes with less than 16GiB RAM (at least for the agents), so this is not a topic for me (and most who want to have a production-grade cluster, I think). 
RKE2 has a close alignment to upstream kubernetes and is more secure including FIPS140-2 if setup correspondingly. that supports a lot of governmental use cases, that's why I think it's attractive to support it.

valkenburg-prevue-ch:
Yes, I agree, @ifeulner . 

I've been wondering if, since the motivation is security, it would make sense to go for an architecture where none of the cluster nodes have a public ip, and all maintenance and node-level egress pass through a bastion. I think `squid` is a good candidate for running a proxy in the bastion, listening only on private network.

mysticaltech:
@valkenburg-prevue-ch Basically in the current setup, [you can close the SSH port](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/436#issuecomment-1379241029) after the cluster is up and that basically isolates it, and same for the other in ports like the Kubernetes API one.

mysticaltech:
FYI I went ahead and let the admins of `rke2_selinux` know of our intention to use it and our need for the fix. https://github.com/rancher/rke2-selinux/issues/33

ifeulner:
> I've been wondering if, since the motivation is security, it would make sense to go for an architecture where none of the cluster nodes have a public ip, and all maintenance and node-level egress pass through a bastion. I think `squid` is a good candidate for running a proxy in the bastion, listening only on private network.

I am not sure how much more security that gives you compared to block the port via a firewall. 
But it introduces a lot of new challenges... without a public interface the host has no internet connectivity. So you have either to use a proxy for every outside connection or a router. These resources need to be managed / maintained and introduce new bottlenecks / overhead.

valkenburg-prevue-ch:
That's true. GKE has a NAT router for their solution, https://cloud.google.com/kubernetes-engine/docs/concepts/private-cluster-concept . The fact that it exists means there are customers who appreciate this design. It's how you do it when you setup a classical hadoop cluster for banks. But yes, as Hetzner does not have NAT routers in their cloud offering (or am I wrong?) you have to have a proxy on the bastion (or another) host, like I said already. 

I understand this might be too heavy on maintenance on this public open source project.

valkenburg-prevue-ch:
About the security:
- it's simply a different attack surface compared the firewall-only solution,
- with a proxy like squid, you can monitor (and filter) all outgoing traffic.

jspizziri:
We're currently running rke2 on AWS and have been playing around with the idea of migrating to Hetzner. I can't guarantee anything but we might be able to justify pushing this PR forward to explore that further. I'd be interested in what the maintainers Definition of Done is for merge. I haven't given the TF a thorough look yet but as I understand it from the above:

- [ ] Fix the `rke2-selinux` bug
- [ ] ...?

Is it really just a matter of taking the PR and trying to get something deployed on Hetzner with it? Thanks in advance!

mysticaltech:
@jspizziri Thanks for looking into this. Basically, this PR will be closed as no longer up to date with the changes. However, we will try to support RKE2. And that should be straightforward because we use k3s, and RKE2 is done by the same team using the the same config format and flow.

Done will be reached when a future PR (that you could start if you want) has both k3s and RKE2 well modularized and working with most features that this project offers.

The rke2-selinux is not something I worry a lot, because it can be fixed in a few hours of testing, it's just selinux config files, and a similar fix has already been done on the rancher k3s-selinux repo.

PR #450, opened 2022-12-08
Adds option to modify longhorn namespace upon deployment

morphine1:
Hi and thanks for the great project.
Due to some tooling constraints I had to modify the namespace longhorn is deployed into.
Initially, this namespace is hardcoded in the deployment.yaml, so this is my attempt to make it configurable via kube.tf
I thought I ask if you might find this useful, but feel free to dismiss if such additions are not considered relevant.
Thanks,
morphine


mysticaltech:
Thank you @morphine1! If it helped you then it will help others. Appreciate the contribution.

mysticaltech:
LGTM, if you could just run `terraform format`, or just install the Terraform extension in VSCode and save, it would pass the validation. You can double-check with `terraform validate`.

morphine1:
Thanks for the reply. Fixed the formatting, sorry!

mysticaltech:
Thanks, @morphine1, it will be merged into master and released on the weekend, don't hesitate to watch for releases to get a notification when that happens.

PR #454, opened 2022-12-14
fix: control_planes_custom_config must be applied to all control-plan…

admin-prevue:
The user variable `control_planes_custom_config` was only applied on the first control plane node, but with this fix it gets applied on all control plane nodes. Without this fix, a config like this:
```
    "secrets-encryption" : "true",
  }
```

will hang, as the additional control-plane nodes will never come online. With this fix, they do come online without problems.

Is secrets-encryption=true perhaps a good default to consider?


mysticaltech:
@admin-prevue Thank you!

PR #455, opened 2022-12-15
fix: destroy on empty state should work ok.

admin-prevue:
When calling `terraform destroy` without an actual cluster, an error is raised because entry [0] is accessed on an empty list `module.control_planes`. This commit catches that.


mysticaltech:
@admin-prevue Awesome catch, thanks! Will be released on the week-end (don't hesitate to watch for the releases to get a notification).

PR #458, opened 2022-12-15
fix: avoid querying repos for latest version if versions are fixed.

admin-prevue:
A minor modification, but it seems that github can have issues which make this change useful: if versions of packages are fixed in user input (hetzner_ccm, hetzner_csi and kured), there is no need to ping github to retrieve the latest release numbers. This pull request implements the conditionality of the github_release calls.


mysticaltech:
@admin-prevue Absolutely!

PR #462, opened 2022-12-19
Fix/destroy on correct state of failed build.

admin-prevue:
Apparently it can happen that Hetzner Cloud takes too long to setup on of the servers. In this case, terraform times out, and there is no ip in the state, even though the server exists. When terraform-destroying on that state, the ipv4 field may not exist and the destroy could fail. With this fix (building on a previous fix), that scenario now works.

### DIFF
--- a/kubeconfig.tf
+++ b/kubeconfig.tf
@@ -12,7 +12,7 @@ data "remote_file" "kubeconfig" {
 }
 
 locals {
-  kubeconfig_external = replace(replace(data.remote_file.kubeconfig.content, "127.0.0.1", var.use_control_plane_lb ? hcloud_load_balancer.control_plane.*.ipv4[0] : (length(keys(module.control_planes)) > 0 ? module.control_planes[keys(module.control_planes)[0]].ipv4_address : "unknown")), "default", var.cluster_name)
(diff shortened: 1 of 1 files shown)


mysticaltech:
Thanks, @admin-prevue!

admin-prevue:
The pleasure is all mine, happy to contribute tiny bits to an excellent open source project. Thank you!

mysticaltech:
@admin-prevue Just released in v1.6.9. Really appreciate this, great contributions! 🙏 🥳

If you'd like to become a member of the kube-hetzner org, you just let me know it would be an honor. We have a team-only discussion section, and it would also give you write priviledge, and the ability to create branches and stuff. We're about eight devs now but not all are active. It's really a best-effort thing!

ksnip_20221220-121007

mysticaltech:
@admin-prevue I invited you, no worries, if you can't, your PRs and contributions are always welcome! Keep up the good work 🙏

admin-prevue:
Thanks for the invite. Not sure how much I can contribute, but we'll see!

-------- Oorspronkelijk bericht --------
Op 21 dec. 2022 09:27, schreef Karim Naufal :

mysticaltech:
Good to hear! Welcome and happy holidays.

PR #470, opened 2022-12-27
fix typo in readme

peterbecich:


mysticaltech:
Ah, good catch, thanks @peterbecich!

PR #477, opened 2023-01-02
Output control plane ip as ingress

captnCC:
This feature existed but somehow got lost in refactoring


mysticaltech:
Yes! Thanks, @captnCC.

Issue #460, opened 2022-12-16
Apply changes to k3s_registries without needing to destroy cluster


timopulkkinen:
Hello, 

thanks for the project!

I made changes to k3s_registries variable in kube.tf, but these do not get applied to the cluster. 

Expected result: 
Terraform applies changes and cluster nodes get rebooted

Result:
`No changes. Your infrastructure matches the configuration.

Terraform has compared your real infrastructure against your configuration and found no differences, so no changes are needed.
`

Is there some workaround to apply these settings?

mysticaltech:
@timopulkkinen This is something that needs to be done on first start. However, what would definitely work is the following. (@s3rius FYI if you have more ideas, please drop them here too).

- Create new nodepools for both control plane and agents, with the same counts as the others. Apply.
- Those new nodepools with get created with the new settings, so with the k3s_registries.
- Drain each node in the old nodepools, for each nodepool, proceed node by node, and nodepool by nodepool, set the count to 0 (don't remove), apply.
- When all is done, you will have a cluster with the same number of nodes but with the new settings.

OR

Backup your cluster with a tool like Velero from VMWare (this repo for example) and create a new cluster with the new settings, and restore everything there.


I know it's not ideal, but for now, the above is the only thing that would work IMHO.

s3rius:
The easiest way to update is to do it manually. But this would make your terraform state out of sync with the real infra. 

@mysticaltech, I guess we can check if this variable were updated, drain all affected nodes, update the file and reboot them.

mysticaltech:
Yep, @s3rius. something along these lines. We could use the `lifecycle` block with `create_before_destroy` to do that. Also, we can check if it exists in the state beforehand with the `terraform.state.exists(<ressource-name>)`  function.

In which case, we need to decouple the generation and copying of the registries yaml file from the rest.

s3rius:
@mysticaltech, thanks for advice. I'll try to do it later.

ifeulner:
All these changes lead us to *operate* the cluster... the current terraform approach is mainly for initial provisioning. Changing configurations after cluster creation needs a proper draining and deletion of the nodes, for example. If we want to support the operation of the cluster, the next question is if terraform is the right tool for it, from my perspective ? What do you thing @mysticaltech ?

mysticaltech:
[...]

PR #486, opened 2023-01-04
Update `registries.yaml` on existing nodes

captnCC:
Issue: #460 

- [x] Deploy `registries.yaml` as `null_resource`
- [x] Add `registries.yaml` to cloud-init for new autoscaled instances
- [x] Update autoscaler on update
- [x] Prevent reboot on first creation
- [x] Find a way to handle autoscaled nodes

### DIFF
--- a/autoscaler-agents.tf
+++ b/autoscaler-agents.tf
@@ -1,4 +1,5 @@
 locals {
+  cluster_prefix = var.use_cluster_name_in_node_name ? "${var.cluster_name}-" : ""
   autoscaler_yaml = length(var.autoscaler_nodepools) == 0 ? "" : templatefile(
     "${path.module}/templates/autoscaler.yaml.tpl",
     {
@@ -6,12 +7,20 @@ locals {
       ca_image         = var.cluster_autoscaler_image
       ca_version       = var.cluster_autoscaler_version
       ssh_key          = local.hcloud_ssh_key_id
-      ipv4_subnet_id   = hcloud_network.k3s.id # for now we use the k3s network, as we cannot reference subnet-ids in autoscaler
-      snapshot_id      = hcloud_snapshot.autoscaler_image[0].id
-      firewall_id      = hcloud_firewall.k3s.id
-      cluster_name     = var.use_cluster_name_in_node_name ? "${var.cluster_name}-" : ""
-      node_pools       = var.autoscaler_nodepools
+      ipv4_subnet_id   = hcloud_network.k3s.id
+      # for now we use the k3s network, as we cannot reference subnet-ids in autoscaler
+      snapshot_id  = hcloud_snapshot.autoscaler_image[0].id
+      firewall_id  = hcloud_firewall.k3s.id
+      cluster_name = local.cluster_prefix
+      node_pools   = var.autoscaler_nodepools
   })
+  # A concatenated list of all autoscaled nodes
+  autoscaled_nodes = {
+    for v in concat([
+      for k, v in data.
+      hcloud_servers.autoscaled_nodes : [for v in v.servers : v]
+    ]...) : v.name => v
+  }
 }
 
 resource "hcloud_snapshot" "autoscaler_image" {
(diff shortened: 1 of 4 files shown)


captnCC:
I'm still searching for a way to select the autoscaled nodes, so far I came up with:
- `hcloud_servers` data source in terraform but it would require labels on the servers
- selecting nodes via `kubectl` but as far as I'm aware, autoscaled nodes don't have any labels either to identify them

If anybody has an idea, please share!

mysticaltech:
@captnCC Definitely we should label the autoscaled nodes. @ifeulner any ideas on how to do that? You might know this better.

mysticaltech:
@captnCC If you choose to tag the Hetzner nodes themselves, users that want to force backward compatibility could for instance add the required labels to autoscaled nodes manually. Just an example.

ksnip_20230104-182723

ifeulner:
hi, the implementation for the Hetzner provider already seems to use labels (on Hetzner server level, not on Kubernetes node level!), see the implementation here: https://github.com/kubernetes/autoscaler/blob/1a459646fc795825e4df8285e61151bcf9ab22dc/cluster-autoscaler/cloudprovider/hetzner/hetzner_node_group.go#L355
So these label could be used for selecting the server.

captnCC:
Thanks @ifeulner, I oversaw that the label was set! This solves the problem for existing autoscaled nodes 👌🏻

This narrows the problems down to one: Is there a clean solution to limit the execution of the null_resource provisioners to changes of `var.k3s_registries` and be ignored on the initial server creation?

mysticaltech:
@captnCC I used OpenAI's ChatGPT to help me reason about the problem, here's the final solution it says could work, now I am not sure personally, but you could give it a try.

```
```

ifeulner:
> Thanks @ifeulner, I oversaw that the label was set! This solves the problem for existing autoscaled nodes 👌🏻
> 
> This narrows the problems down to one: Is there a clean solution to limit the execution of the null_resource provisioners to changes of `var.k3s_registries` and be ignored on the initial server creation?

@captnCC what do you mean exactly? If the registry config is included in the cloud-init, then it will change whenever the registry settings change? Currently the registry settings are not considered... 

I also have doubts if we should change running autoscaler nodes as they are managed by the cluster-autoscaler...

captnCC:
@mysticaltech Thanks, I will give it a try! 

@ifeulner What I did so far: The `registries.yaml` will be included in the cloud-init for the autoscaler. This insures that new nodes have the up to date values. Additionally all nodes (static and scaled) will receive the updated registries and instruct kured to reboot the nodes.

In an ideal setup the autoscaler would take care of recreating the nodes after the cloud-init changes, but I can't find any documentation if there is any handling for this

mysticaltech:
@captnCC From my limited standpoint. It seems impossible to access autoscaled nodes, as we do not have control over them. So I would suggest we let those be. And it will be up to the user to scale them down manually, temporarily, to ensure that new ones are created with the correct registries.yaml. But up to you!

captnCC:
@mysticaltech I have to correct you there! I'm now able to query the servers via the hcloud provider and then access them via `remote-exec. So technically it is possible, open is the question of "should we"

mysticaltech:
Oh wow, that is what you were asking for indeed, via the labels. Really, if you ask me, it seems clean enough, why not!

captnCC:
> @captnCC It's this kind of special magic trick that makes our project special and awesome! Go for it.
Thanks!

So the last problem is fixed, a bit of bash magic solved it easily!
@mysticaltech Its interesting what ChatGPT can spit out, but I don't think it could work. I didn't even find any docs about accessing `terraform.state`

mysticaltech (approved):
@captnCC Looks good to me! Love the bash magic, simple and efficient. 

Man, now I am afraid of ChatGPT, I too cannot find any docs on terraform.state.exists(), it "imagined" it lol.

mysticaltech:
@ifeulner Please have a look if you can, on my end, it looks ready to merge!

autoscaler-agents.tf
```diff
@@ -85,7 +94,47 @@ data "cloudinit_config" "autoscaler-config" {
...
         })
+        k3s_registries = var.k3s_registries
       }
     )
   }
 }
+
+data "hcloud_servers" "autoscaled_nodes" {
+  for_each      = toset(var.autoscaler_nodepools[*].name)
+  with_selector = "hcloud/node-group=${local.cluster_prefix}${each.value}"
```
ifeulner, line 105: @captnCC Are you sure this will work? I haven't tried it yet but I thought this is the group name you set when defining the autoscaler groups in `autoscaler_nodepools` variable? Don't forget it's possible to define multiple groups there.

captnCC: Exactly, the splat expression takes the names of all nodes pools and puts them in an array

ifeulner: Ah, right, sorry, did miss that. Looks good then, thanks a lot @captnCC

mysticaltech:
@captnCC FYI. Before doing a release I always do a test deploy. And with the following kube.tf, it did not work. Screenshot of error below.

```
locals {
  hcloud_token = "xxxxx"
}

module "kube-hetzner" {
  providers = {
    hcloud = hcloud
  }

  source = "../kube-hetzner"

  cluster_name = "test12"

  network_region = "eu-central"

  control_plane_nodepools = [
    {
      server_type = "cpx21",
      location    = "fsn1",
      labels      = [],
      taints      = [],
      count       = 1
    },
  ]

  agent_nodepools = [
    {
      name        = "agent-small",
      server_type = "cpx21",
      location    = "fsn1",
      labels      = [],
      taints      = [],
      count       = 1
    },
  ]

}

provider "hcloud" {
  token = local.hcloud_token
}

terraform {
  required_version = ">= 1.2.0"
  required_providers {
    hcloud = {
      version = ">= 1.35.1"
    }
  }
}

output "kubeconfig" {
  value     = module.kube-hetzner.kubeconfig_file
  sensitive = true
}
```

ksnip_20230110-211435


Please try and issue a fix PR when you can 🙏

mysticaltech:
@captnCC I forced pushed the master branch to a previous commit so that it is now in a working state. The changes are all in the staging branch. If you open a new PR, you can point to it directly 🚀

captnCC:
@mysticaltech Good catch! This should be an easy fix, should be ready quickly!

Issue #489, opened 2023-01-07
Upgrading Module from v1.4.x to >=v1.5.x fails


me-marcel:
**Description**
Upgrading the Module from v1.4.x to >=v1.5.x fails. 

**Error**
```
Error: no Load Balancer found with name kube1
│
│   with module.kube-hetzner.data.hcloud_load_balancer.cluster[0],
│   on .terraform/modules/kube-hetzner/data.tf line 20, in data "hcloud_load_balancer" "cluster":
│   20: data "hcloud_load_balancer" "cluster" {
```

**Environment**
```
locals {
  # Fill first and foremost your Hetzner API token, found in your project, Security, API Token, of type Read & Write.
  hcloud_token = "XXXXXXXXXXxxxxxxxxxxXXXXXXXXXXxxxxxxxxxxXXXXXXXXXXxxxxxxxxxxXXXX"
}

module "kube-hetzner" {
  providers = {
    hcloud = hcloud
  }
  hcloud_token    = local.hcloud_token
  source          = "kube-hetzner/kube-hetzner/hcloud"
  version         = "1.4.8"
  ssh_public_key  = file("./xxx")
  ssh_private_key = file("./yyy")
  network_region  = "eu-central"
  control_plane_nodepools = [
    {
      name        = "control-plane-fsn1",
      server_type = "cpx11",
      location    = "fsn1",
      labels      = [],
      taints      = [],
      count       = 3
    }
  ]
  agent_nodepools = [
    {
      name        = "agent-fsn1",
      server_type = "cpx21",
      location    = "fsn1",
      labels      = [],
      taints      = [],
      count       = 3
    }
  ]
  load_balancer_type                = "lb11"
  load_balancer_location            = "fsn1"
  base_domain                       = "xxx.dev"
  traefik_acme_tls                  = true
  traefik_acme_email                = "aaa@bbb.ccc"
  traefik_enabled                   = true
  metrics_server_enabled            = true
  allow_scheduling_on_control_plane = false
  automatically_upgrade_k3s         = true
  cluster_name                      = "kube1"
  use_cluster_name_in_node_name     = true
  placement_group_disable           = true
  enable_cert_manager               = true
  use_control_plane_lb              = true
}

provider "hcloud" {
  token = local.hcloud_token
}

terraform {
  required_version = ">= 1.2.0"
  required_providers {
    hcloud = {
      source  = "hetznercloud/hcloud"
[...]

PR #490, opened 2023-01-07
fix: traefik loadbalancer resource name mismatch

me-marcel:
a name mismatch in accordance to the traefik loadbalancer resource name made upgrades from <=v1.4.x to >=v1.5.x impossible

part of: Issue #489

### DIFF
--- a/data.tf
+++ b/data.tf
@@ -22,7 +22,7 @@ data "github_release" "kured" {
 
 data "hcloud_load_balancer" "cluster" {
   count = local.has_external_load_balancer ? 0 : 1
-  name  = var.cluster_name
+  name  = local.ingress_controller == "traefik" ? "${var.cluster_name}-traefik" : var.cluster_name
 
   depends_on = [null_resource.kustomization]
 }


mysticaltech:
@me-marcel Thank you for this! I understand your concern, but it would break deploys that happened after that change was done. What I suggest it just to point the module in kube.tf, to a local GitHub clone, and do your change manually to be able to upgrade.

Like on like 16 of the screenshot below.

ksnip_20230109-230811

me-marcel:
@mysticxaltech This has already been solved exactly the way you suggested as you have have seen. For reference please take a look at the corresponding issue linked to the PR at https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/489. 

@captnCC actually suggested to allow a full customization of the resource name by variable within the issue. I just didn't get to modifying the PR just yet. That would provide us with a way to tackle the problem without forcing the users to always keeping a local copy of the module.

Bear in mind that one way or another we will have to implement a permanent solution sooner or later. At least I'd suggest to do since implementing a permanent fix isn't too complicated after all.

mysticaltech:
Great to hear... I also think that allowing for the customization of the LB name is an elegant solution!

mysticaltech:
@me-marcel I am closing this for now, but do not hesitate to push a new PR that adds the ability to set a custom name for the LB, ideally for both Traefik and Ingress-Nginx. Such a feature would be very much welcome! Thanks.

Issue #469, opened 2022-12-26
Image pull error on a single node on fresh install


pbnsilva:
Hi,

I've made a fresh install, where my terraform config is the kube.tf example with some extra nodes and longhorn enabled:
```
  control_plane_nodepools = [
    {
      name        = "control-plane-fsn1",
      server_type = "cpx11",
      location    = "fsn1",
      labels      = [],
      taints      = [],
      count       = 1
    },
    {
      name        = "control-plane-nbg1",
      server_type = "cpx11",
      location    = "nbg1",
      labels      = [],
      taints      = [],
      count       = 1
    },
    {
      name        = "control-plane-hel1",
      server_type = "cpx11",
      location    = "hel1",
      labels      = [],
      taints      = [],
      count       = 1
    }
  ]

  agent_nodepools = [
    {
      name        = "agent-small",
      server_type = "cpx11",
      location    = "fsn1",
      labels = [
        "node.kubernetes.io/server-usage=storage"
      ],
      taints = [],
      count       = 1
    },
    {
      name        = "agent-large",
      server_type = "cpx21",
      location    = "nbg1",
      labels      = [],
      taints      = [],
      count = 0
    },
    {
      name        = "agent-xlarge",
      server_type = "cpx41",
      location    = "nbg1",
      labels = [
        "node.kubernetes.io/server-usage=storage"
      ],
      taints      = [],
      count = 0
    },
    {
      name        = "agent-xxlarge",
      server_type = "cpx51",
      location    = "nbg1",
      labels = [
        "node.kubernetes.io/server-usage=storage"
      ],
      taints      = [],
      count = 3
    }
  ]
```

After it is installed, the `hcloud-csi-node` pod on a single node keeps crashing due to:
```
Events:
  Type     Reason     Age                    From               Message
  ----     ------     ----                   ----               -------
  Normal   Scheduled  2m42s                  default-scheduler  Successfully assigned kube-system/hcloud-csi-node-swkbt to k3s-agent-xxlarge-iog
  Normal   Pulling    2m38s                  kubelet            Pulling image "hetznercloud/hcloud-csi-driver:2.1.0"
  Normal   Pulled     2m34s                  kubelet            Successfully pulled image "hetznercloud/hcloud-csi-driver:2.1.0" in 4.139931398s
  Normal   Created    2m33s                  kubelet            Created container hcloud-csi-driver
[...]

Issue #467, opened 2022-12-20
Custom TLS-SAN array


maaft:
It would be great to be able to add custom tls-san entries.

Use-case:
- if primary cp node goes down, you don't want to edit any IPs in your kubeconfig

Solution:
- have multiple A records `cp.cluster.my.org` pointing to all cp nodes
- add `cp.cluster.my.org` to tls-san in `/etc/rancher/k3s/config.yaml`

mysticaltech:
This is quite easy, will do ASAP.

Issue #461, opened 2022-12-17
Can't create single node cluster in us-west (hil)


xrendan:
I get the following error about `qemu-img` not being found.

```
module.kube-hetzner.module.control_planes["0-0-control-plane"].hcloud_server.server (remote-exec): MicroOS.x86  99% 398.57M  18.2MB/s    eta 0s
module.kube-hetzner.module.control_planes["0-0-control-plane"].hcloud_server.server (remote-exec): openSUSE-Mi 100% 401.12M  18.8MB/s    in 23s

module.kube-hetzner.module.control_planes["0-0-control-plane"].hcloud_server.server (remote-exec): 2022-12-16 23:26:16 (17.5 MB/s) - ‘openSUSE-MicroOS.x86_64-OpenStack-Cloud.qcow2’ saved [420610048/420610048]

module.kube-hetzner.module.control_planes["0-0-control-plane"].hcloud_server.server (remote-exec): + ls -a
module.kube-hetzner.module.control_planes["0-0-control-plane"].hcloud_server.server (remote-exec): + grep -ie ^opensuse.*microos.*qcow2$
module.kube-hetzner.module.control_planes["0-0-control-plane"].hcloud_server.server (remote-exec): + qemu-img convert -p -f qcow2 -O host_device openSUSE-MicroOS.x86_64-OpenStack-Cloud.qcow2 /dev/sda
module.kube-hetzner.module.control_planes["0-0-control-plane"].hcloud_server.server (remote-exec): /tmp/terraform_82492173.sh: 4: qemu-img: not found
╷
│ Error: remote-exec provisioner error
│ 
│   with module.kube-hetzner.module.control_planes["0-0-control-plane"].hcloud_server.server,
│   on .terraform/modules/kube-hetzner/modules/host/main.tf line 64, in resource "hcloud_server" "server":
│   64:   provisioner "remote-exec" {
│ 
│ error executing "/tmp/terraform_82492173.sh": Process exited with status 127
```
Here's my kube.tf
```
variable "HCLOUD_TOKEN" {
  type = string
}

variable "REGION" {
  type = string
  default = "us-east"
}

variable "region_data" {
  default = {
    "us-east" = {
      "code" = "ash"
    },
    "us-west" = {
      "code" = "hil"
    },
  }
}

locals {
  # Fill first and foremost your Hetzner API token, found in your project, Security, API Token, of type Read & Write.
[...]

Issue #447, opened 2022-12-07
Default loadbalancer for nginx ingress also targets control plane nodes


mysticaltech:
### Discussed in https://github.com/orgs/kube-hetzner/discussions/446

<div type='discussions-op-text'>

<sup>Originally posted by **MariusSchiffer** December  7, 2022</sup>
I have created a fresh cluster with 5 agent nodes and 3 control plane nodes.  
Unfortunately, the ingress loadbalancer not only targets the agent nodes, but also the control plane nodes.

Do you have any idea what influences this?</div>

KaiGrassnick:
This Comment might help with this issue: https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/331#issuecomment-1344451969

TL;DR; - Set kubernetes label on Control-Plane Nodes:
```
labels:
  node.kubernetes.io/exclude-from-external-load-balancers: "true"
```

mysticaltech:
@KaiGrassnick Thanks a lot for sharing this, appreciate it!

captnCC:
@mysticaltech Should this be set by default for control planes? 🤔

mysticaltech:
@captnCC It should be disabled for the control plane by default unless we enable scheduling on it. We need to open a PR, will do it ASAP.

PR #504, opened 2023-01-11
Multiple fixes

mysticaltech:
Closes:
- #469, fixes container pull errors.
- #467, brings multiple TLS-SANs.
- #461, brings the ability to set a custom mirror link for MicroOS.
- #447, fixes control plane nodes showing up in LB as targets when scheduling is not allowed.
- #436, fixes SSH extra rules merging, now done correctly.

### DIFF
--- a/modules/host/main.tf
+++ b/modules/host/main.tf
@@ -75,7 +75,7 @@ resource "hcloud_server" "server" {
 
     inline = [
       "set -ex",
-      "wget --timeout=5 --waitretry=5 --tries=5 --retry-connrefused --inet4-only https://ftp.gwdg.de/pub/opensuse/repositories/devel:/kubic:/images/openSUSE_Tumbleweed/openSUSE-MicroOS.x86_64-OpenStack-Cloud.qcow2",
+      "wget --timeout=5 --waitretry=5 --tries=5 --retry-connrefused --inet4-only ${var.opensuse_microos_mirror_link}",
       "qemu-img convert -p -f qcow2 -O host_device $(ls -a | grep -ie '^opensuse.*microos.*qcow2$') /dev/sda",
     ]
   }
--- a/modules/host/main.tf
+++ b/modules/host/main.tf
@@ -193,6 +193,7 @@ resource "null_resource" "registries" {
       echo "No reboot required"
     else
       echo "Update registries.yaml, reboot required"
+      mkdir -p /etc/rancher/k3s
       cp /tmp/registries.yaml /etc/rancher/k3s/registries.yaml
       touch /var/run/reboot-required
     fi
--- a/agents.tf
+++ b/agents.tf
@@ -7,21 +7,22 @@ module "agents" {
 
   for_each = local.agent_nodes
 
-  name                       = "${var.use_cluster_name_in_node_name ? "${var.cluster_name}-" : ""}${each.value.nodepool_name}"
-  base_domain                = var.base_domain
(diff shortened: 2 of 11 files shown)


mysticaltech:
@kube-hetzner/core This was tested and seems to work fine, it fixes many small issues.

maaft:
@mysticaltech had a look at lb_hostname, tls-san, k3s upgrade mutex.

 looks good - didn't test though.

Many thanks for your work <3

modules/host/main.tf
```diff
@@ -193,6 +193,7 @@ resource "null_resource" "registries" {
       echo "No reboot required"
     else
       echo "Update registries.yaml, reboot required"
+      mkdir -p /etc/rancher/k3s
```
captnCC, line 196: Just a question, did you run in a problem where the directory didn't exists or is it just a precaution?

captnCC:
From my side, all looks good! Great work!

ifeulner (approved):
LGTM, but weren't able to test yet. Great work!

modules/host/main.tf
mysticaltech: @captnCC Ahh. I've missed the if-else logic in bash. Maybe that is the only thing that needs to be tweaked, the rest is fine if that works to not run on the first deployment!

mysticaltech: "cmp -s" probably doesn't work because the folder is not even there in the first place.

mysticaltech: @captnCC Thanks! It seems the code in the cloud-init itself fails to run. As the folder would not be present too. I wonder how this worked for some people that ended up using the k3s_registries variable. Am a bit confused lol.

captnCC: @mysticaltech I have that feature already in production and can confirm that it is working.
How does your config for the test look?

mysticaltech: @captnCC I trust you, and many users have reported it working too. That means that the cloud-init write-files directive does create the path even if it is missing, I do not see any other explanation.

Here's the test kube.tf I use:

```
locals {
  hcloud_token = "xxxx"
}

module "kube-hetzner" {
  providers = {
    hcloud = hcloud
  }

  source = "../kube-hetzner"

  cluster_name = "test12"

  network_region = "eu-central"

  control_plane_nodepools = [
    {
      server_type = "cpx21",
      location    = "fsn1",
      labels      = [],
      taints      = [],
      count       = 1
    },
  ]

  agent_nodepools = [
    {
      name        = "agent-small",
      server_type = "cpx21",
      location    = "fsn1",
      labels      = [],
      taints      = [],
      count       = 1
    },
  ]

}

provider "hcloud" {
  token = local.hcloud_token
}

terraform {
  required_version = ">= 1.2.0"
  required_providers {
    hcloud = {
      version = ">= 1.35.1"
    }
  }
}

output "kubeconfig" {
  value     = module.kube-hetzner.kubeconfig_file
  sensitive = true
}
```

mysticaltech: @captnCC My only concern at this moment is that it seems the null_ressource.registries does run on first deploy. In this case, it needs to be fixed. 

For now, I will merge this, and you can push an eventual fix in another PR. So that we compartmentalize things better. I will wait for the fix to do a new release.

mysticaltech: @captnCC I understand why it is not creating the file for me, because I do not have the k3s_registries variable defined.

There is this `%{ if k3sRegistries != "" }` check. So the "if cmp" needs to be adapted a little it. I will attempt it quickly.

mysticaltech: Ok, found a way, by tweaking the if, as such `if cmp -s /tmp/registries.yaml /etc/rancher/k3s/registries.yaml || [ ! -f /etc/rancher/k3s/registries.yaml ]; then`. Will use that for now!

mysticaltech: Perfect, sounds good @captnCC ! 🙏

PR #506, opened 2023-01-12
fix spelling

digitalbuddha:
fix spelling


mysticaltech:
Thanks, @digitalbuddha!

PR #508, opened 2023-01-13
fix: validation regex for var.opensuse_microos_mirror_link , because …

valkenburg-prevue-ch:
…the urls listed at the mirror list now include version info.


mysticaltech:
Thanks, @valkenburg-prevue-ch!

PR #510, opened 2023-01-13
feat: added variable for choosing repository for longhorn helm chart.

valkenburg-prevue-ch:
The default repository from which longhorn is pullen, is a mirror from rancher. This assures compatibility with rancher. However, as it is a modified mirror, it can be behind in version compared to the upstream source longhorn.io. This PR adds an optional variable so you can point to longhorn.io.

Use case today: k3s v1.25 with longhorn 1.4 run perfectly fine together when you use the longhorn.io helm repository.

### DIFF
--- a/init.tf
+++ b/init.tf
@@ -194,8 +194,9 @@ resource "null_resource" "kustomization" {
     content = templatefile(
       "${path.module}/templates/longhorn.yaml.tpl",
       {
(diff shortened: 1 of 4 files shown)


mysticaltech:
@valkenburg-prevue-ch Good one! 🙏

PR #511, opened 2023-01-13
README: correct distro name

rombert:


mysticaltech:
@rombert Good catch, thanks!

PR #519, opened 2023-01-16
Add options kured

captnCC:
This MR adds the ability to configure kured with additional options (and override defaults)

### DIFF
--- a/templates/kured.yaml.tpl
+++ b/templates/kured.yaml.tpl
@@ -21,3 +21,7 @@ spec:
             - --pre-reboot-node-labels=kured=rebooting
             - --post-reboot-node-labels=kured=done
             - --period=5m
+            %{~ for key, value in options ~}
+            - --${key}=${value}
+            %{~ endfor ~}
+
--- a/templates/kured.yaml.tpl
+++ b/templates/kured.yaml.tpl
@@ -17,10 +17,6 @@ spec:
         - name: kured
           command:
             - /usr/bin/kured
-            - --reboot-command=/usr/bin/systemctl reboot
-            - --pre-reboot-node-labels=kured=rebooting
-            - --post-reboot-node-labels=kured=done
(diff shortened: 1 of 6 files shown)


templates/kured.yaml.tpl
```diff
@@ -21,3 +21,7 @@ spec:
             - --pre-reboot-node-labels=kured=rebooting
```
mysticaltech, line 21: @captnCC Those actually get re-included below with your new logic. So they need to get deleted, no?

mysticaltech (changes requested):
Looks great, just see the tiny comment about previous kured arguments, those may need to be removed.

templates/kured.yaml.tpl
captnCC: Yes that was an oversight, they are gone now

variables.tf
```diff
@@ -390,6 +390,11 @@ variable "rancher_values" {
   description = "Additional helm values file to pass to Rancher as 'valuesContent' at the HelmChart."
 }
 
+variable "kured_options" {
+  type    = map(string)
+  default = {}
```
mysticaltech, line 395: @captnCC Now we need to re-add the defaults in here.

captnCC: @mysticaltech They will be actually merged with the defaults in the `locals.tf`

mysticaltech: My bad, indeed! :)

mysticaltech (approved):

mysticaltech:
@captnCC If you've done a test deployment, feel free to merge to master, and will do a quick release! As this adds a small new feature, I believe we can move to 1.8, it's about time lol.

Otherwise, if you want me to try it too, you can merge it to staging, and will take care of it ASAP.

mysticaltech:
@captnCC Released, thanks! 🙏🚀

PR #523, opened 2023-01-18
A few changes and fixed check for IP when using lb_hostname

aleksasiriski:
1. **FIXED CHECK FOR `IP` WHEN USING LB_HOSTNAME**
1. Switched to a faster mirror for MicroOS
1. Added option to disable restrictive outbound rules which can cause issues for some images.
1. Switched to official helm chart for Longhorn by default, which doesn't have CRD.
1. Changed the default k8s ver from `v1.24` to `v1.25`.
1. Added `v1.26` k8s to validation list.
1. Replaced K3s integrated Traefik with the official helm chart, which allows for replication.
1. Removed traefik integrated acme since it can't be replicated and no one should be using it even if they can.
1. Enabled cert-manager by default, since the above removes integrated acme anyway, it will be used by both traefik and nginx.
1. Moved both nginx and traefik to their namespaces respectively.
1. Added `traefik_redirect_to_https`.
1. Moved Traefik's `TrustedIPs` to Helm values instead of cli arguments.
1. Added `var.ingress_replica_count` for both nginx and traefik which is by default `0` and then checked with:
```
local.ingress_replica_count = (var.ingress_replica_count > 0) ? var.ingress_replica_count : (local.agent_count > 2) ? 3 : (local.agent_count == 2) ? 2 : 1
```


main.tf
```diff
@@ -73,13 +73,6 @@ resource "null_resource" "destroy_cluster_loadbalancer" {
     cluster_name     = var.cluster_name
   }
 
-  # Important when issuing terraform destroy, otherwise the LB will not let the network get deleted
```
mysticaltech, line 76: You are right this was useless.

mysticaltech (approved):

mysticaltech:
@aleksasiriski Pretty good work there! Thanks.

@ifeulner @captnCC, can you have a look folks, he cleaned up a good amount of obsolete logic.

main.tf
aleksasiriski: Actually I think it was necessary when using k3s integrated Traefik, but since I switched to helm chart for traefik it gets deleted with the code above which deletes the kustomize (all helm charts)

aleksasiriski:
Why is IPv6 failing? I just tried it on my cluster and it changed the firewall correctly. Also it was literally copied from the example config.

mysticaltech:
@aleksasiriski We never had support for IPv6. It's just supported to reach the LB because Hetzner also gives IPv6 addresses.

mysticaltech:
And for the failing checks, please use `terraform format` and `terraform validate` to make sure it passes.

captnCC (changes requested):
All in all great work! I have left a few minor notes to be addressed.
Since this PR touches a lot of logic, we should test it especially for existing clusters

variables.tf
```diff
@@ -191,7 +197,7 @@ variable "enable_metrics_server" {
 
 variable "initial_k3s_channel" {
   type        = string
-  default     = "v1.24"
+  default     = "v1.25"
```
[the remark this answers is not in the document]
aleksasiriski: Even 1.26 in using the official chart, idk about rancher's fork:
https://longhorn.io/docs/1.4.0/deploy/important-notes/#supported-kubernetes-versions

variables.tf
```diff
@@ -153,16 +159,16 @@ variable "enable_traefik" {
   description = "Whether to enable or disable the installation of the Traefik Ingress Controller."
 }
 
-variable "traefik_acme_tls" {
-  type        = bool
-  default     = false
-  description = "Whether to include the TLS configuration with the Traefik configuration."
+variable "traefik_replica_count" {
+  type        = number
+  default     = 3
```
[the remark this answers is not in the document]
aleksasiriski: Why would that make any difference for stability or compatibility? Compatibility is already gone since I switched from K3s' integrated Traefik to official Helm Chart, and keeping this below 2 results in dropped packets when Kubernetes is moving the pod.

templates/traefik_ingress.yaml.tpl
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: traefik
```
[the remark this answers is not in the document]
aleksasiriski: I agree, will make the change. Also I would like to rename "ngx" and "nginx-ingress" occurences to "ingress-nginx" since that's the correct name for the Go implementation of nginx from Kubernetes, whereas "nginx-ingress" is the less popular controller from nginx directly.

main.tf
```diff
@@ -73,13 +73,6 @@ resource "null_resource" "destroy_cluster_loadbalancer" {
...
-  provisioner "local-exec" {
```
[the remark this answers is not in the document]
aleksasiriski: But it is deleted? If you mean that part with traefik, I didn't even think about deleting other parts.

docs/terraform.md
```diff
@@ -110,6 +110,7 @@
 | <a name="input_hetzner_csi_version"></a> [hetzner\_csi\_version](#input\_hetzner\_csi\_version) | Version of Container Storage Interface driver for Hetzner Cloud. | `string` | `null` | no |
 | <a name="input_initial_k3s_channel"></a> [initial\_k3s\_channel](#input\_initial\_k3s\_channel) | Allows you to specify an initial k3s channel. | `string` | `"v1.24"` | no |
 | <a name="input_k3s_registries"></a> [k3s\_registries](#input\_k3s\_registries) | K3S registries.yml contents. It used to access private docker registries. | `string` | `""` | no |
+| <a name="input_kured_options"></a> [kured\_options](#input\_kured\_options) | n/a | `map(string)` | `{}` | no |
```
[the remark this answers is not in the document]
aleksasiriski: Will do

variables.tf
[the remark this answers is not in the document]
ifeulner: Yes, meanwhile longhorn 1.4.0 is release, and this supports K8s 1.25 finally.

templates/traefik_ingress.yaml.tpl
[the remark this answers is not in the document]
mysticaltech: @aleksasiriski Try it, you will see that "ingress-nginx" already shows up in the service name, so it becomes something along the lines of "ngx-ingress-nginx" less repetition than "ingress-nginx-ingress-nginx" (the name of the helm chart is used in there).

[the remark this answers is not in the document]
aleksasiriski: > @aleksasiriski Try it, you will see that "*ingress-nginx" already shows up in the service name, so it becomes something along the lines of "ngx-ingress-nginx" less repetition than "ingress-nginx-ingress-nginx" (the name of the helm chart is used in there).

Well then it's OK for it to stay ngx, but I would like to change every nginx-ingress to ingress-nginx since it may be confusing as it currently is.

ifeulner (commented):
good stuff @aleksasiriski but some comments from me

main.tf
```diff
@@ -73,13 +73,6 @@ resource "null_resource" "destroy_cluster_loadbalancer" {
...
-    when       = destroy
-    command    = "kubectl -n kube-system delete service traefik --kubeconfig kubeconfig.yaml"
-    on_failure = continue
-  }
-
   depends_on = [
```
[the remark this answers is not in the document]
aleksasiriski: Because I switched to Helm Traefik instead of the K3s integrated one and because it's a external Helm chart like ingress-nginx and Longhorn it gets properly removed with Kustomization file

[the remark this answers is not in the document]
aleksasiriski: > Why to have this at all? Maybe we should just support one ingress implementation - nginx. Too many variables otherwise. @mysticaltech what do you think?

I can't comment below for some reason...

I personally prefer Traefik since it's easier to use global middlewares, but for replication to work you can't use the integrated K3s Traefik but the official Helm Chart instead. Also, why would we remove support for it, it's not hard to support both and the code is already here and working..

locals.tf
```diff
@@ -129,121 +127,51 @@ locals {
...
-      #NTP
-      {
-        direction = "out"
-        protocol  = "udp"
-        port      = "123"
-        destination_ips = [
-          "0.0.0.0/0"
-        ]
-      }
-      ], !local.using_klipper_lb ? [] : [
```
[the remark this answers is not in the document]
aleksasiriski: I'll add an option for people to choose if they want to restrict or allow all outbound traffic. My main problem was that this was hardcoded and I had to manually add rules that allow all traffic which resulted in having 6 outbound rules that essentially just cancel each other out for no reason.

locals.tf
```diff
@@ -92,7 +90,7 @@ locals {
   default_agent_taints         = concat([], var.cni_plugin == "cilium" ? ["node.cilium.io/agent-not-ready:NoExecute"] : [])
 
 
-  packages_to_install = concat(var.enable_longhorn ? ["open-iscsi", "nfs-client", "xfsprogs", "cryptsetup"] : [], var.extra_packages_to_install)
+  packages_to_install = concat(["open-iscsi", "nfs-client", "xfsprogs", "cryptsetup", "cifs-utils"], var.extra_packages_to_install)
```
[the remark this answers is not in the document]
aleksasiriski: I'll add an option for other packages as well, but force them of Longhorn is enabled.

variables.tf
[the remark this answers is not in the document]
aleksasiriski: I can revert the check that was used for ingress-nginx replicas and apply if here (if there are a certain number of nodes, use 2 replicas, etc.)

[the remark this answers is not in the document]
aleksasiriski: > Why to have this at all? Maybe we should just support one ingress implementation - nginx. Too many variables otherwise. @mysticaltech what do you think?

I personally prefer Traefik since it's easier to use global middlewares, but for replication to work you can't use the integrated K3s Traefik but the official Helm Chart instead. Also, why would we remove support for it, it's not hard to support both and the code is already here and working..

docs/terraform.md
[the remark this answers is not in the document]
aleksasiriski: I'll leave this to the bot if/when this gets merged...

templates/traefik_ingress.yaml.tpl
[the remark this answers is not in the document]
aleksasiriski: I believe I found a good middleground that isn't confusing and is consistent. It's commited.

aleksasiriski:
I revamped the first comment mentioning all the changes, making it more readable as well as updating what has changed and sorting it.

main.tf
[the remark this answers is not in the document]
otavio: We also do this here. However, in our case, we do that externally by installing it from `fluxcd.`

otavio (approved):

ifeulner (approved):

main.tf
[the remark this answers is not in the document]
captnCC: @ifeulner This line didn't work for a long time since the kubeconfig name changed anyway. Additionally it would only work under a set of circumstances anyway: `kubectl` must be installed, the kubeconfig must exist from a before apply (which can be disabled and doesn't work in CI systems)

aleksasiriski:
Also, if longhorn has the default number of replicas 3, why wouldn't that be default for traefik and nginx as well?

locals.tf
```diff
@@ -129,85 +126,66 @@ locals {
...
+    ], !var.restrict_outbound_traffic ? [] : [
       # Allow basic out traffic
       # ICMP to ping outside services
       {
         direction = "out"
         protocol  = "icmp"
-        destination_ips = [
-          "0.0.0.0/0"
-        ]
+        destination_ips = ["0.0.0.0/0", "::/0"]
```
captnCC, line 152: It's a weirdness in terraform but no the `icmp` fails the validation:
```
╷
│ Error: Inconsistent conditional result types
│ 
│   on locals.tf line 146, in locals:
│  146:     ], !var.restrict_outbound_traffic ? [] : [
│  147:     # Allow basic out traffic
│  148:     # ICMP to ping outside services
│  149:     {
│  150:       direction       = "out"
│  151:       protocol        = "icmp"
│  152:       destination_ips = ["0.0.0.0/0", "::/0"]
│  153:     },
│  154:     # DNS
│  155:     {
│  156:       direction       = "out"
│  157:       protocol        = "tcp"
│  158:       port            = "53"
│  159:       destination_ips = ["0.0.0.0/0", "::/0"]
│  160:     },
│  161:     {
│  162:       direction       = "out"
│  163:       protocol        = "udp"
│  164:       port            = "53"
│  165:       destination_ips = ["0.0.0.0/0", "::/0"]
│  166:     },
│  167:     # HTTP(s)
│  168:     {
│  169:       direction       = "out"
│  170:       protocol        = "tcp"
│  171:       port            = "80"
│  172:       destination_ips = ["0.0.0.0/0", "::/0"]
│  173:     },
│  174:     {
│  175:       direction       = "out"
│  176:       protocol        = "tcp"
│  177:       port            = "443"
│  178:       destination_ips = ["0.0.0.0/0", "::/0"]
│  179:     },
│  180:     #NTP
│  181:     {
│  182:       direction       = "out"
│  183:       protocol        = "udp"
│  184:       port            = "123"
│  185:       destination_ips = ["0.0.0.0/0", "::/0"]
│  186:     }
│  187:     ], !local.using_klipper_lb ? [] : [
│     ├────────────────
│     │ var.restrict_outbound_traffic is a bool
│ 
│ The true and false result expressions must have consistent types. The 'true' tuple has length 0, but the 'false' tuple has length 6.
```
The ports needs to be defined anyway for terraform

variables.tf
```diff
@@ -153,16 +164,21 @@ variable "enable_traefik" {
...
-  description = "Whether to include the TLS configuration with the Traefik configuration."
+variable "traefik_replica_count" {
+  type        = number
+  default     = null
+  description = "Number of replicas per traefik controller."
+
+  validation {
+    condition     = var.traefik_replica_count == null || var.traefik_replica_count > 0
+    error_message = "Number of traefik replicas can't be below 1."
+  }
```
captnCC, line 175: The default should be `1` instead of `null` (as the variable isn't flagged as nullable). Then the validation can be shortened to just `var.traefik_replica_count > 0`

variables.tf
```diff
@@ -116,19 +116,30 @@ variable "hetzner_csi_version" {
...
-variable "nginx_ingress_values" {
+variable "nginx_replica_count" {
+  type        = number
+  default     = null
+  description = "Number of replicas per nginx controller."
+
+  validation {
+    condition     = var.nginx_replica_count == null || var.nginx_replica_count > 0
+    error_message = "Number of nginx replicas can't be below 1."
+  }
```
captnCC, line 139: The default should be `1` instead of `null` (as the variable isn't flagged as nullable). Then the validation can be shortened to just `var.nginx_replica_count > 0`

captnCC (changes requested):
I've done a test to upgrade from the current `master`-branch to the PR branch. Seems to far to go without breaking changes. (I didn't test for longhorn itself).

Yet none of the changes to ingress or anything will be applied to the cluster. 
Should we just go with it letting the existing config be untouched or should we make sure that the changes from the module upgrade will we applied? @mysticaltech @ifeulner

variables.tf
aleksasiriski: Then the code for node count won't ever be activated. Why do you insist so much that is should be 1 and not 3 or 2? Longhorn has 3 for default...

locals.tf
aleksasiriski: Why? It's identical to ```], !local.using_klipper_lb ? [] : [```

variables.tf
captnCC: It's not about the number itself, just that with default = null the config is broken. So either the variable should be flagged as nullable or the default needs to be changed

variables.tf
aleksasiriski: I just removed the replica count and set it to autodetect based on number of agents. It doesn't make sense to be 1 by default and 3 can be overkill for smaller setups.

locals.tf
captnCC: It seems to be a quirk of terraform and the short hand if. So all objects of the array need to have the same form. It played around and found that adding the port fixed it magically

variables.tf
aleksasiriski: Fixed.

variables.tf
aleksasiriski: Fixed.

locals.tf
aleksasiriski: Adding ```port            = ""``` fixed terraform warnings?

mysticaltech: @captnCC Something seems indeed wrong with `var.restrict_outbound_traffic`, as the error message about says "The 'true' tuple has length 0, but the 'false' tuple has length 6".

variables.tf
mysticaltech: @aleksasiriski @captnCC If there is one node in the cluster, it cannot be 3 anyways. So we need to use smart logic here. 1 is total agent counts < 3, and 3 if total agent counts >=3. The nodes counts are already computed in local.tf, so that can be reused.

mysticaltech: > I've done a test to upgrade from the current `master`-branch to the PR branch. Seems to far to go without breaking changes. (I didn't test for longhorn itself).
> 
> Yet none of the changes to ingress or anything will be applied to the cluster. Should we just go with it letting the existing config be untouched or should we make sure that the changes from the module upgrade will we applied? @mysticaltech @ifeulner

@captnCC I need we can just release it under a minor next version so 1.8, it's not big breaking changes, and as you said, the main changes will not get applied anyways.

locals.tf
mysticaltech: @aleksasiriski Yes it needs to be the same format, as the inside logic relies on it for merging.

variables.tf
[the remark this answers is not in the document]
mysticaltech: @ifeulner IMHO we can keep both, as they will work the same, and the logic is there already. They are the two most common ingress controllers for Kube, and supporting both insures we appeal to the max audience!

locals.tf
aleksasiriski: Ok, it's fixed now. I also took the liberty of correcting grammar and updating info in the example file.

locals.tf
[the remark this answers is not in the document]
mysticaltech: We already have an option for extra packages, so @aleksasiriski if cifs is not often used like @ifeulner says, we can just remove it, and people that needs it can use the `extra_packages_to_install` variable.

captnCC (approved):
From my perspective it look good now!

aleksasiriski:
Also switched to markdown's auto incrementing when using an ordered list.

README.md
```diff
@@ -81,10 +81,10 @@ brew install hcloud
 ### 💡 [Do not skip] Creating your kube.tf file
 
 1. Create a project in your [Hetzner Cloud Console](https://console.hetzner.cloud/), and go to **Security > API Tokens** of that project to grab the API key. Take note of the key! ✅
-2. Generate a passphrase-less ed25519 SSH key pair for your cluster; take note of the respective paths of your private and public keys. Or, see our detailed [SSH options](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/master/docs/ssh.md). ✅
-3. Prepare the module by copying `kube.tf.example` to `kube.tf` **in a new folder** which you cd into, then replace the values from steps 1 and 2. ✅
-4. (Optional) Many variables in `kube.tf` can be customized to suit your needs, you can do so if you want. ✅
-5. At this stage you should be in your new folder, with a fresh `kube.tf` file, if it is so, you can proceed forward! ✅
+1. Generate a passphrase-less ed25519 SSH key pair for your cluster; take note of the respective paths of your private and public keys. Or, see our detailed [SSH options](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/master/docs/ssh.md). ✅
```
mysticaltech, line 84: @aleksasiriski FYI, look at the numbers.

README.md
```diff
@@ -462,10 +426,10 @@ There is also a branch where openSUSE MicroOS came preinstalled with the k3s RPM
 Code contributions are very much **welcome**.
 
 1. Fork the Project
-2. Create your Branch (`git checkout -b AmazingFeature`)
-3. Commit your Changes (`git commit -m 'Add some AmazingFeature')
-4. Push to the Branch (`git push origin AmazingFeature`)
-5. Open a Pull Request targeting the `staging` branch.
+1. Create your Branch (`git checkout -b AmazingFeature`)
```
mysticaltech, line 429: @aleksasiriski Same here for the numbered list.

locals.tf
```diff
@@ -415,5 +410,8 @@ installCRDs: true
     "pre-reboot-node-labels" : "kured=rebooting",
     "post-reboot-node-labels" : "kured=done",
     "period" : "5m",
+    "start-time": "${var.kured_start_time}",
```
mysticaltech, line 413: @aleksasiriski Please remove these time constraints for the default case. The reason for that is that for a huge cluster, the way it is configured now, it will automatically reboot 1 node at a time, and will spread it in time as long as needed.

Also, most cluster operations are not contained to any local times, so just best to remove that constraint by default IMHO.

variables.tf
```diff
@@ -395,6 +404,30 @@ variable "rancher_values" {
   description = "Additional helm values file to pass to Rancher as 'valuesContent' at the HelmChart."
 }
 
+variable "kured_version" {
+  type        = string
+  default     = null
+  description = "Version of Kured."
+}
+
+variable "kured_start_time" {
```
mysticaltech, line 413: @aleksasiriski See the comment above, about the default.

README.md
```diff
@@ -177,22 +177,21 @@ You can copy and modify the [one in the templates](https://github.com/kube-hetzn
 
 ### Configuring update timeframes
 
-Per default a node that installed updates will reboot within the next few minutes and updates are installed roughly every 24 hours.
+Per default a node that installed updates will reboot between 3AM and 8AM Local time, and updates are installed roughly every 24 hours.
```
mysticaltech, line 180: @aleksasiriski See my comment below, I would really prefer that to stay the same as before by default, though adding the possibility to customize is great.

mysticaltech (changes requested):

mysticaltech:
@aleksasiriski Keep up the good work! After the above comments are handled, I'll give it one last check and if all good, will do a test deploy, and when that passes, we will merge and release 1.8 🙏

README.md
mysticaltech: And about the detailed options below, best to include them in kube.tf instead, and commented, so folks can just uncomment them if they want to use it. And it will also make the Readme clearer by removing everything that can live in the Kube.tf file instead.

README.md
aleksasiriski: It's autoincrementing when everything is 1. It's a markdown feature!

README.md
aleksasiriski: It's autoincrementing when everything is 1. It's a markdown feature!

locals.tf
aleksasiriski: Will do!

README.md
aleksasiriski: Alright.

README.md
aleksasiriski: You can see [here](https://github.com/aleksasiriski/k3s-on-hetzner/blob/master/README.md) that it's working properly.

README.md
aleksasiriski: Above comment.

aleksasiriski:
I suggest reading the first comment once more, I've updated it with all the changes.

aleksasiriski:
I just made a new cluster and it works flawlessly, here's the config:

```terraform
locals {
  hcloud_token = "woohoo"
}

module "kube-hetzner" {
  providers = {
    hcloud = hcloud
  }

  source = "github.com/aleksasiriski/k3s-on-hetzner"
  ssh_public_key = file("~/.ssh/id_rsa.pub")
  ssh_private_key = file("~/.ssh/id_rsa")

  control_plane_nodepools = [
    {
      server_type = "cpx11",
      location    = "nbg1",
      labels      = [
        "node.kubernetes.io/server-usage=master",
        "node.kubernetes.io/server-location=nbg1",
        "node.kubernetes.io/server-country=ger"
      ],
      taints      = [],
      count       = 1
    },
    {
      server_type = "cpx11",
      location    = "fsn1",
      labels      = [
        "node.kubernetes.io/server-usage=master",
        "node.kubernetes.io/server-location=fsn1",
        "node.kubernetes.io/server-country=ger"
      ],
      taints      = [],
      count       = 1
    },
    {
      name        = "control-plane-hel1",
      server_type = "cpx11",
      location    = "hel1",
      labels      = [
        "node.kubernetes.io/server-usage=master",
        "node.kubernetes.io/server-location=hel1",
        "node.kubernetes.io/server-country=fin"
      ],
      taints      = [],
      count       = 1
    }
  ]

  agent_nodepools = [
    {
      name        = "agent-main-nbg1",
      server_type = "cpx21",
      location    = "nbg1",
      labels      = [
        "node.kubernetes.io/server-usage=main",
        "node.kubernetes.io/server-location=nbg1",
        "node.kubernetes.io/server-country=ger"
      ],
      taints      = [],
      count       = 1
    },
    {
      name        = "agent-main-fsn1",
      server_type = "cpx21",
      location    = "fsn1",
      labels      = [
        "node.kubernetes.io/server-usage=main",
        "node.kubernetes.io/server-location=fsn1",
        "node.kubernetes.io/server-country=ger"
      ],
      taints      = [],
      count       = 1
    }
  ]

  cluster_name = "example"
  base_domain = "cluster.example.com"
  lb_hostname = "ingress.example.com"
  extra_packages_to_install = ["nfs-client", "cifs-utils"]
  restrict_outbound_traffic = false
  ingress_replica_count = 3

  kured_options = {
    "start-time": "03:00"
    "end-time": "08:00"
    "time-zone": "Local"
  }

}

provider "hcloud" {
  token = local.hcloud_token
}

terraform {
  required_version = ">= 1.3.3"
  required_providers {
    hcloud = {
      version = ">= 1.35.2"
    }
  }
}

output "kubeconfig" {
  sensitive = true
}
```

captnCC:
@aleksasiriski Now just the formatting needs to be fixed with `terraform fmt -recursive`, then this should be ready to merge! 🎉

variables.tf
```diff
@@ -456,7 +471,7 @@ variable "k3s_registries" {
 
 variable "opensuse_microos_mirror_link" {
   description = "The mirror link to use for the opensuse microos image."
-  default     = "https://ftp.gwdg.de/pub/opensuse/repositories/devel:/kubic:/images/openSUSE_Tumbleweed/openSUSE-MicroOS.x86_64-OpenStack-Cloud.qcow2"
+  default     = "https://download.opensuse.org/tumbleweed/appliances/openSUSE-MicroOS.x86_64-OpenStack-Cloud.qcow2"
```
mysticaltech, line 474: @aleksasiriski Please revert that. We have chosen that mirror for because the main link gave mirrors that had blacklisted hetzner servers. But the default one we have is working great!

mysticaltech (changes requested):
@aleksasiriski Almost perfect. Please just address the point above and we're good to go!

variables.tf
aleksasiriski: The default one downloads at 300Kbps for me, but OK I'll revert it.. There are also other mirrors available that are much faster

mysticaltech: Yes, but it depends on people's locations. At least this one works well for most people. It's been several months now with no issues. But anyways, you can customize the link by setting the variable.

aleksasiriski: It only depends on Hetzner VPS location, and this mirror is fixated in DE which in turn means across the seas connection for VPSs in USA, and even though it's in DE a DE VPS gets only 300KB/s whereas the official mirror or for example [this one](https://mirror.dogado.de/opensuse/tumbleweed/appliances/openSUSE-MicroOS.x86_64-OpenStack-Cloud.qcow2) gets 1MB/s.

Testing on my own home connection, the old mirror gives around 40MB/s and both the official and dogado DE max out my connection at 60MB/s.

mysticaltech:
@aleksasiriski Thank you for this, it's a great contribution. I will give it a quick deploy, and if all good. It will be released tonight CET under 1.8.0 🚀

mysticaltech (approved):

mysticaltech:
Deployed! 🙏 🚀

Issue #527, opened 2023-01-23
Hetzner Cloud Networks


shyblower:
You write in your feature list:
"Proper use of the Hetzner private network to minimize latency and remove the need for encryption."

But the Hetzner FAQ (https://docs.hetzner.com/cloud/networks/faq) state this:
Q: "Is traffic inside Hetzner Cloud Networks encrypted?"
A: "Traffic between cloud servers inside a Network is private and isolated, but not automatically encrypted. We recommend you use TLS or similar protocols to encrypt sensitive traffic."

So Hetzner Cloud private networks are assumably isolated from from other Hetzner networks but obviously not encrypted, which might pose a security risk, particularly when nodes are connected over the "private network" between different Hetzner data centers which is probably just tunneled over the internet, but as they say, unencrypted.

It should be possible to configure the CNI to use e.g. Flannel with the Wireguard backend for encryption as they do here:
https://github.com/StarpTech/k-andy/pull/15

mysticaltech:
@shyblower You are right! Thanks for bringing this to our attention. @kube-hetzner/core FYI.

Please feel free to submit a PR adding support for wireguard with flannel (the k3s config happens in control_planes.tf, agents.tf, and init.tf), otherwise, we will do it ASAP.

ifeulner:
Hi @shyblower I didn't test it yet but it should work with some additional settings "out-of-the-box" I think:

- Add `wireguard-tools` to `extra_packages_to_install ` variable
- add  `flannel-backend=wireguard-native` to `control_planes_custom_config` variable (also see [k3s docs](https://docs.k3s.io/reference/server-config#networking) )
- create the cluster

That *should* work, but as said, not tested yet. Maybe you can give it a try?

shyblower:
Thank you, I will try and report back, but it will probably take some time. I'm not sure that I can do it today.

valkenburg-prevue-ch:
> * Add `wireguard-tools` to `extra_packages_to_install ` variable
> * add  `flannel-backend=wireguard-native` to `control_planes_custom_config` variable (also see [k3s docs](https://docs.k3s.io/reference/server-config#networking) )
> * create the cluster

I tested it now and it seems to work exactly this way. For full transparency though, I am right now running with a modified locals.tf as per [this issue](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/528) . I hope this has no relevance.

ifeulner:
[...]

PR #529, opened 2023-01-24
feat: add wireguard option

valkenburg-prevue-ch:
I chose not to make this an optional thing, avoiding adding variables. Only merge this branch if tests show that it is a good idea. Otherwise modify the code to make wireguard optional. See https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/527

I do not know all the arguments for and against making this standard, so I am marking this MR as draft and hope to see some discussion about it.

### DIFF
--- a/autoscaler-agents.tf
+++ b/autoscaler-agents.tf
@@ -127,7 +127,7 @@ resource "null_resource" "autoscaled_nodes_registries" {
 
   provisioner "remote-exec" {
     inline = [<<-EOT
-    if cmp -s /tmp/registries.yaml /etc/rancher/k3s/registries.yaml; then
+    if cmp -s /tmp/registries.yaml /etc/rancher/k3s/registries.yaml || [ ! -f /etc/rancher/k3s/registries.yaml ]; then
       echo "No reboot required"
     else
       echo "Update registries.yaml, reboot required"
--- a/control_planes.tf
+++ b/control_planes.tf
@@ -104,6 +104,7 @@ resource "null_resource" "control_planes" {
           kubelet-arg                 = local.kubelet_arg
           kube-controller-manager-arg = local.kube_controller_manager_arg
           flannel-iface               = local.flannel_iface
+          flannel-backend             = "wireguard-native"
           node-ip                     = module.control_planes[each.key].private_ipv4_address
           advertise-address           = module.control_planes[each.key].private_ipv4_address
           node-label                  = each.value.labels
--- a/init.tf
+++ b/init.tf
@@ -20,6 +20,7 @@ resource "null_resource" "first_control_plane" {
           kubelet-arg                 = local.kubelet_arg
           kube-controller-manager-arg = local.kube_controller_manager_arg
           flannel-iface               = local.flannel_iface
+          flannel-backend             = "wireguard-native"
           node-ip                     = module.control_planes[keys(module.control_planes)[0]].private_ipv4_address
           advertise-address           = module.control_planes[keys(module.control_planes)[0]].private_ipv4_address
           node-taint                  = local.control_plane_nodes[keys(module.control_planes)[0]].taints
--- a/locals.tf
+++ b/locals.tf
@@ -91,7 +91,11 @@ locals {
   default_control_plane_taints = concat([], local.allow_scheduling_on_control_plane ? [] : ["node-role.kubernetes.io/control-plane:NoSchedule"])
   default_agent_taints         = concat([], var.cni_plugin == "cilium" ? ["node.cilium.io/agent-not-ready:NoExecute"] : [])
 
-  packages_to_install = concat(var.enable_longhorn ? ["open-iscsi", "nfs-client", "xfsprogs", "cryptsetup"] : [], var.extra_packages_to_install)
+  packages_to_install = concat(
+    ["wireguard-tools"],
+    var.enable_longhorn ? ["open-iscsi", "nfs-client", "xfsprogs", "cryptsetup"] : [],
+    var.extra_packages_to_install,
+  )
 
   # The following IPs are important to be whitelisted because they communicate with Hetzner services and enable the CCM and CSI to work properly.
   # Source https://github.com/hetznercloud/csi-driver/issues/204#issuecomment-848625566
--- a/modules/host/main.tf
+++ b/modules/host/main.tf
@@ -189,7 +189,7 @@ resource "null_resource" "registries" {
 
   provisioner "remote-exec" {
     inline = [<<-EOT
-    if cmp -s /tmp/registries.yaml /etc/rancher/k3s/registries.yaml; then
+    if cmp -s /tmp/registries.yaml /etc/rancher/k3s/registries.yaml || [ ! -f /etc/rancher/k3s/registries.yaml ]; then
       echo "No reboot required"
     else
       echo "Update registries.yaml, reboot required"


aleksasiriski:
This should be optional, Hetzner networks are isolated and if they're used only within the same location (eu-cental for example) I don't think there would be any security risks, but maybe I'm wrong about that? I'll send an email to Hetzner..

aleksasiriski:
> I wonder if there is really any noticeable performance loss.

It can be the default but should be optional, for some edge cases..

valkenburg-prevue-ch:
Yes, I think your absolutely right. Let's make it optional with a good default. Perhaps default false for a little while and then switch to true after we find no issues for a while?

valkenburg-prevue-ch:
I may do it tomorrow morning, but feel free to beat me to it!

aleksasiriski:
> I may do it tomorrow morning, but feel free to beat me to it!

I'm on it, will be done in a couple of minutes...

aleksasiriski:
> I may do it tomorrow morning, but feel free to beat me to it!

It's done.. but I don't know if this:
```terraform
cni_k3s_settings = {
    "flannel" = {
      disable-network-policy = var.disable_network_policy
      flannel-backend        = var.encrypt_kubernetes_api ? "wireguard-native" : "vxlan"
    }
    "calico" = {
      disable-network-policy = true
      flannel-backend        = "none"
    }
    "cilium" = {
      disable-network-policy = true
      flannel-backend        = "none"
    }
  }
```

should be changed to this:
```terraform
cni_k3s_settings = {
    "flannel" = {
      disable-network-policy = var.disable_network_policy
      flannel-backend        = var.encrypt_kubernetes_api ? "wireguard-native" : "vxlan"
    }
    "calico" = {
      disable-network-policy = true
      flannel-backend        = var.encrypt_kubernetes_api ? "wireguard-native" : "none"
    }
    "cilium" = {
      disable-network-policy = true
      flannel-backend        = var.encrypt_kubernetes_api ? "wireguard-native" : "none"
    }
  }
```

mysticaltech:
@valkenburg-prevue-ch Great, initiative! It adequately addresses #527.

Also, I do agree with both of you that it needs to be optional. @aleksasiriski The first option is the right one, as calico and cilium get configured for this differently, folks can just use the *_values variable to do so.

aleksasiriski:
Shouldn't we add this encryption to calico and cilium as well?

otavio:
> > Please do. I'm using cillium as flannel had intermittent network failures. As I plan to add bare metal machine in the cluster for ci use the wireguard is required.

ifeulner (changes requested):
It's not the api but the whole network communication between the nodes and pods. So the name of the variable is a little bit misleading. Also it is this way only for flannel. As soon as you change the CNI this setting is not valid anymore.

aleksasiriski:
> It's not the api but the whole network communication between the nodes and pods. So the name of the variable is a little bit misleading. Also it is this way only for flannel. As soon as you change the CNI this setting is not valid anymore.

How should we rename it? `encrypted_network`? I'll look into enabling it for calico and cilium

valkenburg-prevue-ch:
How about `enable_wireguard`, and a comment in kube.ft.example explaining that right now this is only relevant when using flannel?

aleksasiriski:
> @aleksasiriski Definitely! But, the reason why I am more relaxed about this is that they have the `*_values` variable to add it if they want to. It just needs some reading of the helm values yaml.
> 
> However, why not make this easier for them, and piggyback on the `encrypt_kubernetes_api` boolean too! 🚀

There literally is no `calico_values`, only cilium, I added it my PR. Wouldn't it better to switch from patching the manifest to calico's helm chart?

aleksasiriski:
> I wonder if there is really any noticeable performance loss.

As I read in cilium's docs, when wireguard is enabled EVERY node makes a tunnel to EVERY other node. If you had 20 nodes that would mean n(n-1)/2 where n=20 -> 190 wireguard tunnels.

mysticaltech:
> > @aleksasiriski Definitely! But, the reason why I am more relaxed about this is that they have the `*_values` variable to add it if they want to. It just needs some reading of the helm values yaml.
> > However, why not make this easier for them, and piggyback on the `encrypt_kubernetes_api` boolean too! rocket
> 
> There literally is no `calico_values`, only cilium, I added it my [PR](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/536). Wouldn't it better to switch from patching the manifest to calico's helm chart?

Oh yes, had forgotten about that. The reason for that is that I actually did not find the calico helm chart, if you do find it, please let's more to helm!

aleksasiriski:
> > > @aleksasiriski Definitely! But, the reason why I am more relaxed about this is that they have the `*_values` variable to add it if they want to. It just needs some reading of the helm values yaml.
> > > However, why not make this easier for them, and piggyback on the `encrypt_kubernetes_api` boolean too! rocket
> > 
> > 
> > There literally is no `calico_values`, only cilium, I added it my [PR](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/536). Wouldn't it better to switch from patching the manifest to calico's helm chart?
> 
> Oh yes, had forgotten about that. The reason for that is that I actually did not find the calico helm chart, if you do find it, please let's move to helm!

I found [this](https://projectcalico.docs.tigera.io/getting-started/kubernetes/helm), but when running `helm show values` it's quite poor compared to cilium for example.

mysticaltech:
Great! @aleksasiriski If it supports wireguard, and a few other basic options, I think it's better than the current manifest approach!

aleksasiriski:
> Great! @aleksasiriski If it supports wireguard, and a few other basic options, I think it's better than the current manifest approach!

I'll add it to my PR then

aleksasiriski:
> Great! @aleksasiriski If it supports wireguard, and a few other basic options, I think it's better than the current manifest approach!

This is all that helm show values outputs:

```yaml
imagePullSecrets: {}

installation:
  enabled: true
  kubernetesProvider: ""

apiServer:
  enabled: true

certs:
  node:
    key:
    cert:
    commonName:
  typha:
    key:
    cert:
    commonName:
    caBundle:

# Resource requests and limits for the tigera/operator pod.
resources: {}

# Tolerations for the tigera/operator pod.
tolerations:
- effect: NoExecute
  operator: Exists
- effect: NoSchedule
  operator: Exists

# NodeSelector for the tigera/operator pod.
nodeSelector:
  kubernetes.io/os: linux

# Custom annotations for the tigera/operator pod.
podAnnotations: {}

# Custom labels for the tigera/operator pod.
podLabels: {}

# Image and registry configuration for the tigera/operator pod.
tigeraOperator:
  image: tigera/operator
  version: v1.29.0
  registry: quay.io
calicoctl:
  image: docker.io/calico/ctl
  tag: v3.25.0
```

mysticaltech:
@aleksasiriski There is also this page, that links to a helm chart, I do not know if it's the same, but it seems to be the main one. Also, there is the Tigera operator, maybe that is needed to deploy/manage Calico. https://github.com/projectcalico/calico/blob/master/calico/getting-started/kubernetes/helm.md

ifeulner:
As long as we don't know for sure what a good implementation would mean for other CNI providers we should just support what we know that it works. 
I used flannel with wg without any problems. If anyone has real experiences with Cilium et al. then we could support a setting that really works. Otherwise we should wait until someone is coming up with a setting. 
I am even asking if we need additional support here for Flannel as it obviously already works with the current configuration options (by adjusting two variables)

mysticaltech:
Yes, that makes a lot of sense @ifeulner!

However, if @aleksasiriski can make Calico work, then why not, it seems like a low-hanging fruit!

mysticaltech:
I mean, we have it working already, so it's just a push to transition to a helm chart if possible. It's something I attempted myself in the past.

mysticaltech:
Alright folks, this looks good to me and would properly add wireguard encryption to our cluster! We need not overcomplicate it with mingling concerns about how calico is deployed. 

So let's test this and merge! I will do a test run ASAP.

mysticaltech:
Ok merging to the staging branch and will do a round of testing before releasing. Great work @valkenburg-prevue-ch @aleksasiriski ! 🚀

PR #533, opened 2023-01-25
Emergency downgrade of the container-selinux package + tiny improvements

### LINKED ISSUE #528: apply_k3s_selinux fails on MicroOS snapshot of 2023-01-23
The script `apply_k3s_selinux = ["/sbin/semodule -v -i /usr/share/selinux/packages/k3s.pp"]` fails as of this morning, with the following output:

```
k3s-agent-large-hel1-djs:~ # /sbin/semodule -v -i /usr/share/selinux/packages/k3s.pp
Attempting to install module '/usr/share/selinux/packages/k3s.pp':
Ok: return value of 0.
Committing changes:
Conflicting name type transition rules
Binary policy creation failed at /var/lib/selinux/targeted/tmp/modules/400/k3s/cil:135
Failed to generate binary
/sbin/semodule:  Failed!
```

I have the strong suspicion that something changed in microos, as the image has been updated yesterday:
Screenshot 2023-01-24 at 11 16 26

Does anyone know how to work around this?

mysticaltech:
This fixes #528, by temporarily downgrading the `container-selinux` package, while we wait for the fix, most probably coming in the next version of `k3s-selinux`. See https://github.com/k3s-io/k3s-selinux/issues/36

Also, I lock the `container-selinux` package, so that an automatic upgrade, so not break the system.

ksnip_20230125-042006

Last but not least, this PR does small improvements, like switching back to the automatic mirror OpenSUSE MicroOS link, as previously suggested by @aleksasiriski, while preserving the known-working mirror link in the `kube.tf.example`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,6 +216,13 @@ kubectl delete plan k3s-agent -n system-upgrade
 kubectl delete plan k3s-server -n system-upgrade
 ```
 
+Also, note that after turning off nodes upgrades, you will need to manually upgrade the nodes when needed. You can do so by SSH'ing into each node and running the following commands (and don't forget to drain the node before with `kubectl drain <node-name>`):
+
+```sh
+transactional-update
+reboot
+```
+
 ### Individual Components Upgrade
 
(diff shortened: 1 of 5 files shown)


valkenburg-prevue-ch:
Impressive, thanks for the hard work and the fast fix!

PR #540, opened 2023-01-26
Feat/custom env

valkenburg-prevue-ch:
Commits related to this comment https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/discussions#discussioncomment-4779692

These changes allow you to set environment variables on the host nodes. Notably, HTTP(S)_PROXY which lets you route all the outbound traffic. Use this to work around the fact that some Hetzner IPs are blocked by various different container registries.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,5 @@
+# Security Policy
+
+## Reporting a Vulnerability
+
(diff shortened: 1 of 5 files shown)


mysticaltech:
@valkenburg-prevue-ch I love it, this is super practical! Thanks.

PR #542, opened 2023-01-26
Fix/auto assign ips

valkenburg-prevue-ch:
No need to compute the ips, have them be auto-assigned. The benefit is:
- less code to maintain,
- no worries when adding things to the network later which have not been imagined yet.


mysticaltech:
FYI for others, more info here https://github.com/orgs/kube-hetzner/discussions/541

valkenburg-prevue-ch:
Oh no... you're right!

Issue #535, opened 2023-01-25
403s on MicroOS download causing deployment failure


sandrinr:
If there is an error setting up a control plane node, rectifying that error state does not seem to be obvious.

In may case, when deploying a new cluster, the download of `openSUSE-MicroOS.x86_64-OpenStack-Cloud.qcow2` failed with `ERROR 403: Forbidden.` on one of the nodes.

Terraform correctly terminated the apply with:
```
│ Error: remote-exec provisioner error
│ 
│   with module.kube-hetzner.module.control_planes["1-0-control-plane-nbg1"].hcloud_server.server,
│   on .terraform/modules/kube-hetzner/modules/host/main.tf line 64, in resource "hcloud_server" "server":
│   64:   provisioner "remote-exec" {
│ 
│ error executing "/tmp/terraform_1335984070.sh": Process exited with status 8
```

Running `terraform apply` again correctly identifies the node that has to be replaced:
```
  # module.kube-hetzner.module.control_planes["1-0-control-plane-nbg1"].hcloud_server.server is tainted, so must be replaced
+/- resource "hcloud_server" "server" {
      + backup_window              = (known after apply)
      ~ datacenter                 = "nbg1-dc3" -> (known after apply)
      ~ id                         = "27970132" -> (known after apply)
      ~ ipv4_address               = "redacted" -> (known after apply)
      ~ ipv6_address               = "redacted" -> (known after apply)
      ~ ipv6_network               = "redacted" -> (known after apply)
        name                       = "k3s-control-plane-nbg1-yhf"
      ~ status                     = "running" -> (known after apply)
        # (15 unchanged attributes hidden)
    }
```

However, the application of the plan then exists with:
```
module.kube-hetzner.module.control_planes["1-0-control-plane-nbg1"].hcloud_server.server: Creating...
╷
│ Error: server name is already used (uniqueness_error)
│ 
│   with module.kube-hetzner.module.control_planes["1-0-control-plane-nbg1"].hcloud_server.server,
│   on .terraform/modules/kube-hetzner/modules/host/main.tf line 22, in resource "hcloud_server" "server":
│   22: resource "hcloud_server" "server" {
```

Is this expected behavior? Is further manual intervention expected/needed in such situations?

<details>
<summary>Version details</summary>

[terraform-hcloud-kube-hetzner](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner) version: 1.8.1
```
$ terraform version
[...]

PR #543, opened 2023-01-26
Feat/private ip during microos install

valkenburg-prevue-ch:
refactor: moving the shell scripting of the server installation outside of the hcloud_server block.
    
All the remote-exec blocks are now in a null-resource, instead of immediately in the hcloud_server block. The benefit is that now the private ip can be assigned (in the subnet) before microos is installed. With the current code base that has zero impact, but it opens up the possibility to ho to communicate safely in the private network at the installation stage already.

### DIFF
--- a/modules/host/main.tf
+++ b/modules/host/main.tf
@@ -183,7 +183,10 @@ resource "null_resource" "server_setup" {
 }
 
 resource "null_resource" "registries" {
-  depends_on = [hcloud_server.server]
+  depends_on = [
+    hcloud_server.server,
+    null_resource.server_setup,
+  ]
 
   triggers = {
     registries = var.k3s_registries
--- a/modules/host/main.tf
+++ b/modules/host/main.tf
@@ -43,12 +43,26 @@ resource "hcloud_server" "server" {
       user_data,
     ]
   }
+}
+
+resource "hcloud_server_network" "server" {
+  # auto-assign the ip.
+  server_id = hcloud_server.server.id
+  subnet_id = var.ipv4_subnet_id
+}
+
+resource "null_resource" "server_setup" {
+
+  # do this after the private ip assignment
+  depends_on = [
+    hcloud_server_network.server
+  ]
 
   connection {
     user           = "root"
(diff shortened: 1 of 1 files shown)


valkenburg-prevue-ch:
Hm, after actually testing this, the private ip is not know in the actual host during its rescue boot (the phase where microos is installed). Bummer, I was hopping to redistribute the microos image through the private network, descreasing the number of downloads.

mysticaltech:
@valkenburg-prevue-ch Interesting take, but please have a look at what @ifeulner is working on to fix the download issues. Though packer, super-duper elegant! 

You actually download on localhost and create a snapshot in your project before the terraform stage. This will be a game-changer! More info here https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/535#issuecomment-1403728279

mysticaltech:
I actually use that packer project to deploy MicroOS instances for docker-compose use and it runs amazingly well.

valkenburg-prevue-ch:
Ah yes, I should have checked that one more closely. Will use it. I was under the impression (my mistake) that it would become a part of this project's initialization, but I understand it will always be outside of it, and it will be a manual task.

valkenburg-prevue-ch:
Wait, but there is no flag in kube-hetzner yet to use a custom snapshot instead of running the whole microos-triple-reboot dance, right?

aleksasiriski:
> Wait, but there is no flag in kube-hetzner yet to use a custom snapshot instead of running the whole microos-triple-reboot dance, right?

It's a work in progress by @ifeulner . I assume it will be integrated into kube-hetzner and used as a single primary way of installing MicroOS.

mysticaltech:
Exactly @aleksasiriski, @valkenburg-prevue-ch He's working on it. The separate repo was just a proof of concept to launch the snapshot. 

But the integration with terraform comes is being worked on for v2.0. I am sure that he will share more as soon as he can, and we can all help him tune this up!

mysticaltech:
@valkenburg-prevue-ch Turning it into a draft for now, feel free to close or re-activate it.

Issue #548, opened 2023-01-27
Is touch /var/run/reboot-required wiser then systemctl restart k3s? 


valkenburg-prevue-ch:
https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/cf2ef6bbac78097015cd969c71882145b534df68/modules/host/main.tf#L199

The docs at https://docs.k3s.io/installation/private-registry state:
"In order for the registry changes to take effect, you need to restart K3s on each node."

if we replace 
```
touch /var/run/reboot-required
```
with
```
if systemctl status k3s > /dev/null; then 
  systemctl restart k3s; 
elif systemctl status k3s-agent > /dev/null; then 
  systemctl restart k3s-agent; 
fi      
```

would this change be a good or a bad idea?

mysticaltech:
@valkenburg-prevue-ch It's unwise because we need to node to be drained properly before restarting k3s on it. So by touching reboot-required, kured does all that magic for us.

mysticaltech:
There is another possibility, maybe we can create another [system-upgrade-controller plan](https://github.com/rancher/system-upgrade-controller), just for that, and activate it via labels. And it is extremely suited for working safely with a k3s/rke2 node.

valkenburg-prevue-ch:
Thanks for explaining, I thought so much! 

I would hope that workloads continue even if the k3s service is restarting for a few seconds. I mean, k3s is the orchestrator, but hopefully is not involved in every bit of communication of the cluster? I don't know what I'm talking about, this is just my story-telling understanding of k8s.

valkenburg-prevue-ch:
I cannot find info about this so quickly. Surely someone with more expertise knows the way to the answer.

mysticaltech:
Actually, the system-upgrade-controller has the possibility to execute commands, but we must double-check with their team if it's safe to just restart k3s. Also, it is very important to use it to make sure it does one node at a time, starting with control planes of course.

mysticaltech:
We would basically force a k3s upgrade to the same version (if possible), and that would cause the system upgrade controller to do all the magic, including the draining and k3s restart for us. 

An interesting starting point could be https://docs.k3s.io/upgrades/automated and our own current mechanism to upgrade based on checking the version channel https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/master/templates/plans.yaml.tpl

mysticaltech:
[...]

PR #550, opened 2023-01-27
fix: systemctl restart k3s instead of rebooting the machine.

valkenburg-prevue-ch:
This addresses the question in https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/548 , and after discussion we concluded that this restart is safe and immediate.

### DIFF
--- a/autoscaler-agents.tf
+++ b/autoscaler-agents.tf
@@ -130,10 +130,14 @@ resource "null_resource" "autoscaled_nodes_registries" {
     if cmp -s /tmp/registries.yaml /etc/rancher/k3s/registries.yaml || [ ! -f /etc/rancher/k3s/registries.yaml ]; then
       echo "No reboot required"
     else
-      echo "Update registries.yaml, reboot required"
+      echo "Update registries.yaml, restart of k3s required"
       mkdir -p /etc/rancher/k3s
       cp /tmp/registries.yaml /etc/rancher/k3s/registries.yaml
-      touch /var/run/reboot-required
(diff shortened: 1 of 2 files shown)


otavio:
Awesome. This raises one doubt to me: how can someone keep up with those improvements without recreating the cluster?

otavio:
Awesome. This raises one doubt to me: how can someone keep up with those improvements without recreating the cluster?

ifeulner (approved):
Still not a big fan of changing the nodes but in this case it is obviously well supported.
Cool!

mysticaltech:
@otavio This is our first feature that changes on live clusters! 

Already works now, but will trigger a reboot of nodes one by one. That change will remove the reboot! 🤞

mysticaltech:
@kube-hetzner/core This if it works is a game changer ! Because it means we can change any k3s config on the fly! 

@ifeulner Yes, we will keep that at a minimal with proper warning!

mysticaltech:
And credit for that goes to both @captnCC and @valkenburg-prevue-ch, for their hard work and constant push for improvement! 🚀

mysticaltech:
Please do review ASAP, so we can merge. I will do testing on my end too.

otavio (approved):

captnCC:
> Awesome. This raises one doubt to me: how can someone keep up with those improvements without recreating the cluster?

@otavio This changes should work with existing clusters, since this script is run every time the registries are updated!

PR #551, opened 2023-01-28
Fixed kube.tf.example

Julian-Louis:
Hello !
The default value of `enable_cert_manager` is `true`,
This PR fix the message in kube.tf.example saying that the default value is `false`

### DIFF
--- a/kube.tf.example
+++ b/kube.tf.example
@@ -346,8 +346,8 @@ module "kube-hetzner" {
   # By default, we allow ICMP ping in to the nodes, to check for liveness for instance. If you do not want to allow that, you can. Just set this flag to true (false by default).
   # block_icmp_ping_in = true
 
-  # You can enable cert-manager (installed by Helm behind the scenes) with the following flag, the default is "false".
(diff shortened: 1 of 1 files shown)


mysticaltech:
Thanks, @Julian-Louis for the notice. This change was introduced recently by mistake it seems, and will be reverted, the correct default is indeed false!

I will let you know when the change is pushed so we can close this PR.

mysticaltech:
Actually, let's keep it that way, why not, it's a useful addon that is used most of the time in clusters at some point of its operation.

aleksasiriski:
> Actually, let's keep it that way, why not, it's a useful addon that is used most of the time in clusters at some point of its operation.

I deliberatelly enabled it by default since I removed traefik's integrated acme support. It made sense to have it on by default because Traefik, which now must use cert-manager, is also enabled by default.

mysticaltech:
Good call, it's better that way @aleksasiriski! As this is ubiquitously used. It's a must-have service.

PR #557, opened 2023-01-30
initial usage of packer to create MicroOS snapshot

ifeulner:
and create with that the nodes. Still work to do...

Here an early integration with packer to create a snapshot of MicroOS upfront. 
Thought it might be useful to start with the 1.9.x branch and getting this stable before switching to RKE2. 

This early version integrates `packer` (please have it installed in your PATH) with terraform now, this way the variables can be reused.
It currently just creates a basic MicroOS snapshot, all the k3s - related package installation and adjustment is done as before.
Also currently the additional snapshot for the autoscaler is created. 

Of course this can be optimized now a lot, but I wanted to get some early feedback from you guys @mysticaltech @captnCC especially as we talked about it.
Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ To achieve this, we built up on the shoulders of giants by choosing [openSUSE Mi
 - [x] Possibility to have a **single node cluster** with a proper ingress controller.
 - [x] Can use Klipper as an **on-metal LB** or the **Hetzner LB**.
 - [x] Ability to **add nodes and nodepools** when the cluster is running.
-- [x] Possibility to turn on **Longhorn** and/or **Hetzner CSI**.
+- [x] Possibility to toggle **Longhorn** and **Hetzner CSI**.
 - [x] Choose between **Flannel, Calico, or Cilium** as CNI.
 - [x] Optional **Wireguard** encryption of the Kube network for added security.
 - [x] **Flexible configuration options** via variables and an extra Kustomization option.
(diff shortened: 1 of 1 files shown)


mysticaltech:
@ifeulner Good work. I fully agree it would be good to switch to it before RKE2.

Will deep-dive ASAP.

dictvm:
I gave your branch a try in an experimental cluster I'm working on. Thank you so much for this, this saves a lot of time.

Let me know if I can help test anything in particular. 

So far I've created a new cluster from scratch and added 2 nodes to an existing nodepool.

aleksasiriski:
I'll look into removing the additional image for autoscaler, @ifeulner if you have any lingering commits please do push them

ifeulner:
@aleksasiriski I already did some work there, give me some more days, as I am currently quite busy and I have some concrete ideas what to change...

aleksasiriski:
> @aleksasiriski I already did some work there, give me some more days, as I am currently quite busy and I have some concrete ideas what to change...

Alright, I made [this branch](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/tree/packer-integration) but I will wait for your commits before changing anything.

mysticaltech:
@ifeulner @aleksasiriski Something important that just came to mind is that we need to keep our module compatible with terraform cloud. So we need the ability to create the initial snapshot also with terraform.

Basically, when the `terraform apply` command is applied, IF THE SNAPSHOT IS NOT IN THE PROJECT it should default to create it via Terraform.

Now, in the project, for manual users, we could have a small bash script that first calls `packer`, before calling terraform, hence insuring that the snapshot is already in the project and ready to use.

ifeulner:
> @ifeulner @aleksasiriski Something important that just came to mind is that we need to keep our module compatible with terraform cloud. So we need the ability to create the initial snapshot also with terraform.

Hm, ok, that I haven't had in mind, but as far as I know terraform cloud also supports packer? Otherwise I have another idea, but that would need some more rewriting I think...

mysticaltech:
@ifeulner From what I have seen, they support packer build from the HashiCorp registry via the [HCP Packer Provider](https://developer.hashicorp.com/hcp/docs/packer/reference-image-metadata#hcp-terraform-provider), but not really what we want I think.

IMHO, the logic you wrote that we use now for the Autoscaler snapshot can be reused/rehashed, as plan B if the snapshot is not in the account (meaning a scenario like running from Terraform cloud without packer).

mysticaltech:
Something else that would be great folks, is the ability to regenerate the snapshot. 

Let's say I wait 6 months, and come to the cluster and want to deploy new nodes, it would be cool first to do a `terraform destroy -target snapshot` and then do a `terraform apply` again!

ifeulner:
> Something else that would be great folks, is the ability to regenerate the snapshot.
> 
> Let's say I wait 6 months, and come to the cluster and want to deploy new nodes, it would be cool first to do a `terraform destroy -target snapshot` and then do a `terraform apply` again!

Yeah, absolutely, I had this already in mind... will provide more on the weekend.

PR #582, opened 2023-02-12
Microos snapshot stage 2

ifeulner:
Hi, I don't know why but the last pull request has vanished..

Anyway, this know removes the initial snapshot and just uses the packer based snapshot for all created servers now, including the cluster autoscaler nodes. 
Please have a look and test it, hadn't hat much time for it yet. 

But as recently came the request, to make this work also in terraform cloud, I see the following alternatives:

1.  fall back to have the `packer` step outside of the terraform script (initial idea). Needs some additional "glue" scripting
2. avoid packer, do the snapshot with terraform - either by a separate script or include it here. Needs some additional "glue" scripting
3. Include the snapshot creation in the "host" module. Should be possible (when no `image_id`given, create first a snapshot, otherwise use the image_id. But Increases the complexity of the script

For me, using packer/ or a separate terraform script  to create the snapshot and assume the image_id in the terraform script would be the simpliest, cleanest way, as we can get rid of a lot of code in the main script. 

What do you think?

### DIFF
--- a/agents.tf
+++ b/agents.tf
@@ -8,6 +8,7 @@ module "agents" {
   for_each = local.agent_nodes
 
   name                         = "${var.use_cluster_name_in_node_name ? "${var.cluster_name}-" : ""}${each.value.nodepool_name}"
+  microos_image_id             = data.hcloud_image.microos_image.id
   base_domain                  = var.base_domain
   ssh_keys                     = length(var.ssh_hcloud_key_label) > 0 ? concat([local.hcloud_ssh_key_id], data.hcloud_ssh_keys.keys_by_selector[0].ssh_keys.*.id) : [local.hcloud_ssh_key_id]
   ssh_port                     = var.ssh_port
--- a/modules/host/main.tf
+++ b/modules/host/main.tf
@@ -21,8 +21,7 @@ resource "random_string" "identity_file" {
 
 resource "hcloud_server" "server" {
   name               = local.name
-  image              = "ubuntu-20.04"
-  rescue             = "linux64"
+  image              = var.microos_image_id
   server_type        = var.server_type
   location           = var.location
   ssh_keys           = var.ssh_keys
@@ -57,33 +56,6 @@ resource "hcloud_server" "server" {
     command = <<-EOT
       install -b -m 600 /dev/null /tmp/${random_string.identity_file.id}
       echo "${local.ssh_client_identity}" > /tmp/${random_string.identity_file.id}
-    EOT
-  }
-
-  # Install MicroOS
-  provisioner "remote-exec" {
-    connection {
-      user           = "root"
-      private_key    = var.ssh_private_key
-      agent_identity = local.ssh_agent_identity
-      host           = self.ipv4_address
-
-      # We cannot use different ports here as this runs inside Hetzner Rescue image and thus uses the
-      # standard 22 TCP port.
-      port = 22
-    }
-
-    inline = [
-      "set -ex",
-      "wget --timeout=5 --waitretry=5 --tries=5 --retry-connrefused --inet4-only ${var.opensuse_microos_mirror_link}",
-      "qemu-img convert -p -f qcow2 -O host_device $(ls -a | grep -ie '^opensuse.*microos.*qcow2$') /dev/sda",
-    ]
-  }
-
-  # Issue a reboot command.
-  provisioner "local-exec" {
-    command = <<-EOT
-      ssh ${local.ssh_args} -i /tmp/${random_string.identity_file.id} root@${self.ipv4_address} '(sleep 2; reboot)&'; sleep 3
     EOT
   }
 
--- a/modules/host/main.tf
+++ b/modules/host/main.tf
@@ -1,15 +1,15 @@
-resource "random_string" "server" {
-  length  = 3
-  lower   = true
-  special = false
-  numeric = false
-  upper   = false
 
+resource "random_id" "server_id" {
   keepers = {
     # We re-create the apart of the name changes.
     name = var.name
+    # Generate a new id each time we switch to a new image id?
+    # image_id = var.microos_image_id
   }
+
+  byte_length = 3
 }
+
 
 resource "random_string" "identity_file" {
   length  = 20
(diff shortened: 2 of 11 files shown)


templates/autoscaler-cloudinit.yaml.tpl
```diff
@@ -121,5 +116,12 @@ runcmd:
...
-- [systemctl, 'start', 'k3s-agent']
+# enable k3s-agent service
+- [systemctl, 'enable', 'k3s-agent']
+
+# reboot!
+power_state:
+    delay: now
+    mode: reboot
+    message: MicroOS rebooting to reflect changes
+    condition: true
```
mysticaltech, line 127: @ifeulner Why is this needed? As it already boots into the correct btrfs snapshot.

packer.tf
```diff
@@ -0,0 +1,44 @@
+locals {
+  packages = length(local.packages_to_install) > 0 ? "-var packages_to_install=${local.packages_to_install}" : ""
+}
+data "hcloud_image" "microos_image" {
+  with_selector = "microos-snapshot=yes,creator_id=${null_resource.packer.id}"
+  most_recent   = true
+}
+
+resource "null_resource" "packer" {
```
mysticaltech, line 9: Maybe the backend terraform mechanism can happen here?!

templates/autoscaler-cloudinit.yaml.tpl
ifeulner: But within the first boot the install of k3s happens, which will includes a new transactional-update. That's why also the k3s service is only enabled and not started.

agents.tf
```diff
@@ -8,6 +8,7 @@ module "agents" {
   for_each = local.agent_nodes
 
   name                         = "${var.use_cluster_name_in_node_name ? "${var.cluster_name}-" : ""}${each.value.nodepool_name}"
+  microos_image_id             = data.hcloud_image.microos_image.id
```
mysticaltech, line 11: Since at other places snapshot_id is used, we should keep the naming consistent, easy fix.

templates/autoscaler-cloudinit.yaml.tpl
mysticaltech: @ifeulner Yes, but what I do not understand is why the reboot at the end of cloud-init since it's not even booted yet.

mysticaltech: Our install of k3s does not do a transactional-update.

mysticaltech:
@ifeulner About packer and terraform, I'm all for 1. and it would be a pleasure working on a small bash script to make the magic happen.

So we would need pure terraform logic that would kick in if not snapshot is detected, meaning it did not run via the init.sh script that calls packer.

This is great, because we can do plenty of things like ensure that all the required CLIs are present, like of course terraform, packer and ideally also hcloud.

And ideally, we would all it via curl to always get the latest and greatest, like what k3s does. Honesly, bash is super easy to write with ChatGPT, like it ridiculous. So that's an ideal use case!

Also I really like the fact that it's optional. Folks that want to use packer then terraform directly can! And they can even use terraform directly and bypass packer to fetch the image online like we do now.

So really this bash script is a layer just for adding UX goodies, like in the future automatically fetching mirrors and testing them even. If we push the envelope we could even install homebrew for them and install all the required CLIs if not present. 

But at first let's keep it simple! Just save them to type the both the packer and terraform command is a great start. And called in the form of:

```
curl -sfL https://kube-hetzner.github.io/install | sh -
```

mysticaltech:
About the previous PR, it's here https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/557 and you yourself closed it.

And thanks for this new one, looking great! Will test ASAP 🙏 🚀

mysticaltech:
@ifeulner Let's forget the bash proposal above for now and instead use a `snapshot_deploy_method` variable, that would default to packer integrated and called via local-exec as you do already, but also have a pure terraform version like in the current master.

modules/host/main.tf
```diff
@@ -60,33 +59,6 @@ resource "hcloud_server" "server" {
     EOT
   }
 
-  # Install MicroOS
```
mysticaltech, line 63: @ifeulner So this logic could come back and execute based on the value 
of `snapshot_deploy_method` if it's equal to "terraform" for instance.

mysticaltech:
@ifeulner Please let me know what you think of the above proposals. Also, to keep things well compartimentalized, I propose moving the packer logic to the host module, or a new module of it's own, along with the terraform snapshot creation logic. Basically, making sure that each concerns is well compartimentalized, while staying compatible with terraform cloud when the right options are chosen (in our case `snapshot_deploy_method="terraform"` for instance).

mysticaltech:
One last thing, and maybe we can clarify this in a subsequent PR is that, we need to clarify which variables can be changed after the initial apply, with or without `terraform destroy -target null_ressource.kustomization`

ifeulner:
> @ifeulner Please let me know what you think of the above proposals. Also, to keep things well compartimentalized, I propose moving the packer logic to the host module, or a new module of its own, along with the terraform snapshot creation logic.
> 
> Basically, making sure that each concerns is well compartmentalized, while staying compatible with terraform cloud when the right options are chosen (in our case `snapshot_deploy_method="terraform"` for instance).

Hi @mysticaltech there are multiple options now, but it‘s more about „packaging“ and not functionality anymore.
I would prefer to keep it as simple as possible, though. So Option 1 would be my preference. 
But I understand you want to keep the terraform-only way, right?

mysticaltech:
@ifeulner Yes, option one is fine, but let definitely with support for terraform only! 🙏

otavio:
Using external logic for the snapshot allows for decoupling and easier code maintenance, but more critical is simplifying the snapshot update without needing to recreate the cluster.

mysticaltech:
@ifeulner I created this new branch [snapshot-deploy](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/tree/snapshot-deploy) and opened #598, we can continue there and more people can participate if they want to.

@gthieleb Thanks for sharing your experience, I like your ideas, reducing the remote-execs if possible would be ideal, please don't hesitate to open a PR targeting the new branch mentioned above.

@aleksasiriski Same for you man, if inspired the new branch above should be the target.

Let's make this awesome folks! 🙏

PR #594, opened 2023-02-18
Add "PubkeyAuthentication=yes" to ssh_args

decius7bc:
If your local SSH environment is configured to disable authentication by public/private keys, SSH connections fail due to no other authentication means remaining.

This small modification adds `PubkeyAuthentication=yes` to the `ssh_args` variable, thus always enabling public key authentication when running Kube-Hetzner.


mysticaltech:
Thanks @decius7bc, useful tweak!

PR #598, opened 2023-02-20
Snapshot deploy

mysticaltech:
Continuation of #582 by @ifeulner.


mysticaltech:
@aleksasiriski @gthieleb I am tagging you both as you both showed interest in the previous PR. Basically, here's the full context as of today (it can evolve, but we will reflect everything here):

- We are going to dissociate the packer code from terraform, and have a pure terraform working version, but with snapshot only. 

- If packer is run first and the snapshot is present and no indications of need to replace (could be a flag) is present, the terraform snapshot generation code will be skipped.

- Ideally, there would be a bash glue code, similar to `curl -sfL https://get.k3s.io | sh -`. It would pull the packer files, the kube.tf, even create a new folder and execute packer if wanted. THIS WOULD BE FULLY OPTIONAL of course, just nice but easy sugar. I have a pretty good idea of how that would work, so I can take this on.

- Any remote-exec functionality that can be moved elsewhere closer to its purpose should be, and everything that can be moved into modules should be. In the near future, we are planning to add support for RKE2, so laying the groundwork here to have things well modularized to allow just creating an RKE2 module later on and adding switching logic, would be ideal.

---

In recap: Packer separated, Terraform-only works, and as modularized as possible.

mysticaltech:
@ifeulner We discussed the above together already! @captnCC We miss you man, and hope to see you again soon, the project is moving and your input here if you are inspired is more than welcome. 

The rest of @kube-hetzner/core FYI people!

valkenburg-prevue-ch (approved):
Looks all good to me!

mysticaltech:
@valkenburg-prevue-ch Thanks for the review, it's good but will fail to run on Terraform cloud for instance, so changes are needed.

aleksasiriski:
I'll take a look and see what I can do to help ASAP

gthieleb:
I understand this is WIP. If its not on the todo list, I would appreciate an optional variable to be able to set a static image id. This can be used in case I've created a snapshot in another terraform state and want to reuse this. This can also be used to trigger or suppress the packer build.

ifeulner:
@mysticaltech et al: I will separate the snapshot creation and the main terraform script... pretty much what we had earlier, though. But earliest on the upcoming weekend.

aleksasiriski:
Could [this](https://cloud.hashicorp.com/products/packer) be used for Terraform Cloud? I don't know if it's required for Hetzner to support is.. There's a [provider](https://registry.terraform.io/providers/hashicorp/hcp/latest/docs) for it as well.

mysticaltech:
@aleksasiriski Exactly. So currently we have Terraform cloud support, and this needs to stay. The default version should deploy the snapshot with pure terraform as done now, we should have the ability to detect the presence of the snapshot in the project already, meaning it was deployed via packer locally. 

Also, we should have a to regenerate the snapshot via terraform if the said flag is true. By the same token, re-executing the packer command locally should also regenerate the snapshot if already present. Basically, we need to rely on the name of the snapshot, not the id, as this can change.

As for the hcp provider, it's not for us now, that's something else, people can use it if they want to, but we do not need to impose them to work with it.

mysticaltech:
@aleksasiriski @ifeulner I will also be working on that functionality soon enough. Just waiting to have an opening in time, as it does need a good few hours of focus. But if you folks get it done before, even better!

Needless to say, it needs to be super-awesome-magic, we are taking the project to the next level. The goal is stability, and maintainability, a great UX, but also readiness for RKE2.

ifeulner:
Hi, just pushed the changes to extract the snapshot generation into an explicit `packer build` step. 
Now it's a 2-step approach:
- first run `packer build .` within the packer-template directory
- then run `terraform apply`

Currently the snapshot is detected via da data-resource, this needs to be discussed. I also only tested yet a very small configuration, all variations must be checked.

Who can support?

Then the usual stuff is also still needed: testing, documentation, etc...

aleksasiriski:
> Hi, just pushed the changes to extract the snapshot generation into an explicit `packer build` step. Now it's a 2-step approach:
> 
>     * first run `packer build .` within the packer-template directory
> 
>     * then run `terraform apply`
> 
> 
> Currently the snapshot is detected via da data-resource, this needs to be discussed. I also only tested yet a very small configuration, all variations must be checked.
> 
> Who can support?
> 
> Then the usual stuff is also still needed: testing, documentation, etc...

But does it retain the option to not use packer at all to be compatible with Terraform Cloud?

aleksasiriski:
>     * first run `packer build .` within the packer-template directory

This would mean that people need to download that as well, or cd into `.terraform/modules/kube-hetzner/packer-template` and then run that command, instead of just having to use kube.tf.example.

I think you should revert to before where it was integrated within Terraform and add an option "use packer to build the image, or the slow old method which doesn't require packer"

Also running `packer build .` within that dir doesn't work OOTB because the API key is missing if not added to env..

```
Error: 1 error(s) occurred:

* token for auth must be specified

  on hcloud-microos-snapshot.pkr.hcl line 30:
  (source code not available)
  ```

**Comment by ifeulner** [comment]
> This would mean that people need to download that as well, or cd into `.terraform/modules/kube-hetzner/packer-template` and then run that command, instead of just having to use kube.tf.example.

In general, that's by intention, see the discussion above. 
Packaging is a different thing, I would suggest moving that to https://github.com/kube-hetzner/packer-hcloud-microos 
as the terraform module usually will not be downloaded. 

Packer is supported by HCP.

But yes, with two separated steps also more configuration / preparation is necessary. 

> My suggestion would be that you revert to before where it was integrated within Terraform and add an option to "use packer to build the image, or the slow old method which doesn't require packer" while making the default to use packer method. Users of Terraform Cloud can switch to the old method manually.

That would be a step back of my understanding of our discussion above, but you're welcome to contribute this. 

As said, packaging / usage is now something that needs to be discussed, that why I submitted the current state.

**Comment by mysticaltech** [comment]
@ifeulner Good work, we are getting closer. Now we still need pure terraform node creation (as discussed above) for cases like terraform cloud, in case a snapshot is not already detected. I can take care of this myself.

@aleksasiriski I believe your worries can be answered with some bash magic, but ideal run in a way like k3s does, via curl.

Will work on this ASAP. In the meantime folks, if inspired, follow your heart and PR to this branch!

**Comment by aleksasiriski** [comment]
Would it be possible to do a poll to check how many people use this tool with Terraform Cloud?

**Comment by mysticaltech** [comment]
@aleksasiriski It's not about a poll, it's a design choice. As a terraform module, it's super important that it works there! We had lots of people use it there since the beginning. And it's simple to do. Thanks for your patience folks!

File: packer-template/hcloud-microos-snapshot.pkr.hcl [hunk 1]
```diff
@@ -0,0 +1,91 @@
...
+
+locals {
+  needed_packages = join(" ", concat(["restorecond policycoreutils setools-console"], var.packages_to_install))
+}
+
+source "hcloud" "microos-snapshot" {
+  image       = "ubuntu-20.04"
+  rescue      = "linux64"
+  location    = "nbg1"
+  server_type = "cx21" # at least a disk size of >= 40GiB is needed to install MicroOS image
```
mysticaltech, line 34: @ifeulner Actually the minimum supported in cpx11 which also has 40GB. I will update this.

packer-template/hcloud-microos-snapshot.pkr.hcl
```diff
@@ -0,0 +1,91 @@
...
+      transactional-update --continue shell <<< "zypper --gpg-auto-import-keys install -y ${local.needed_packages}"
+      sleep 1 && udevadm settle && reboot
+      EOT
+    ]
+    expect_disconnect = true
+  }
+
+  # Ensure connection to MicroOS and do house-keeping
+  provisioner "shell" {
+    pause_before = "5s"
```
mysticaltech, line 78: @ifeulner Why is that needed, you just want to ensure the system is well initialized?

templates/autoscaler-cloudinit.yaml.tpl
```diff
@@ -72,26 +72,9 @@ write_files:
...
-%{ endif }
 
-- content: |
-    set -vx
-    curl -sfL https://get.k3s.io | INSTALL_K3S_SKIP_START=true INSTALL_K3S_SKIP_SELINUX_RPM=true INSTALL_K3S_CHANNEL=${k3s_channel} INSTALL_K3S_EXEC=agent sh -
-    /sbin/semodule -v -i /usr/share/selinux/packages/k3s.pp
-    systemctl start k3s-agent
+- content: ${base64encode(join("\n", install_k3s))}
+  encoding: base64
   path: /var/pre_install/install-k3s-agent.sh
```
mysticaltech, line 78: Since the system is virgin, we can further optimize this by removing the k3s uninstall step.

mysticaltech:
@ifeulner @aleksasiriski It may seem dumb, but it just dawned on me that folks that are going to use the terraform cloud, will just need to ensure themselves that a snapshot is already present on the project, otherwise it will fail. Not really such a big deal.

So basically this is pretty much in a good state already, again awesome work @ifeulner!

What remains is a small bash script to automatically check for the presence of the snapshot, check for the presence of the hcloud, terraform, and packer cli, automatically download the microos image, and running the packer command for them if they choose yes. It will also serve to download the packer hcl file, and initialize a folder with kube.tf. Basically making sure the UX is not degraded, and even improved.

Also, we must test well to elucidate paths of upgradability, so that we can provide our users with step by step instructions and commands to upgrade from 1.10.x, if possible. 

I will try my best to give that a shot over the week-end.

PR #599, opened 2023-02-20
Hcloud token env

gthieleb:
Make token usage optional to use `HCLOUD_TOKEN` environment variable instead. This helps in avoiding unintentional versioning the token within git and to use it within a pipeline.


mysticaltech:
Thanks @gthieleb, I had no idea this was possible! Very useful indeed.

mysticaltech:
Deployed in v1.9.4 🚀

Issue #601, opened 2023-02-21
[Feature Request]: Inject token to cloud-init using env var


gthieleb:
### Description

When using hcloud token via envvar (as described in #599) the token is currently not injected in cloudinit. 

Background: Hetzner does not offer an approach similar to AWS IAM instance role so we need to ship the token ourself.

We should think aboout a way to use an envvar and be able to inject the token anyway.

mysticaltech:
@gthieleb Does that mean that setting HCLOUD_TOKEN alone is not enough to successfully deploy? 

Because if so, we could just rename it TF_VAR_hcloud_token and [it would automatically be picked up](https://support.hashicorp.com/hc/en-us/articles/4547786359571-Reading-and-using-environment-variables-in-Terraform-runs). Please let me know so that I do the change to the kube.tf.

mysticaltech:
@gthieleb I when ahead an renamed the env variable to use to `TF_VAR_hcloud_token`, all you be fixed and that FR handled too! Please let me know if not.

See https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/master/kube.tf.example

PR #602, opened 2023-02-21
Apply hcloud token as var in module call

gthieleb:
This PR makes the hcloud token available as variable that can be injected via envvar. This is the solution proposed by @mysticaltech in #601.

### DIFF
--- a/kube.tf.example
+++ b/kube.tf.example
@@ -1,13 +1,18 @@
 locals {
-  # set your Hetzner API token here or define the TF_VAR_hcloud_token env within your shell for security purpose. It can be found in your Project > Security > API Token (Read & Write is required).
   hcloud_token = ""
+}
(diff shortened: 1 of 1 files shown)


mysticaltech:
@gthieleb I see where I got it wrong. However, redeclaring the variable is not needed.

What you can try is just doing this test:

`hcloud_token = hcloud_token ?? local.hcloud_token
`
And remove the default value of terraform in variables.tf so that it takes the value of null and not "" like now.

mysticaltech:
So at the end, this PR should, theoretically (and IMHO) change just one line.

mysticaltech:
@gthieleb It does not work, just gave it a try.

mysticaltech:
@gthieleb I did some changes to this PR, see 603. I tested and it works, and also is backward compatible which was important.

PR #603, opened 2023-02-21
Hcloud token env 2

mysticaltech:
Suplants #602

### DIFF
--- a/kube.tf.example
+++ b/kube.tf.example
@@ -1,18 +1,17 @@
 locals {
-  hcloud_token = ""
-}
-variable "hcloud_token" {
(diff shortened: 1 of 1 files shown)


mysticaltech:
@gthieleb Deployed in v1.9.5 🚀

gthieleb:
Looks good to me. Thanks for your fast response. Reimplementing variables for submodules does not looks tidy but I dont know an alternative way.

mysticaltech:
@gthieleb Exactly! There was no other way to make it work. And anyways, this kube.tf file is just an example, at the end of the day folks are free to use the underlying kube-hetzner module the way they wish, but at least those are good defaults!

PR #604, opened 2023-02-21
feat: variable for the INSTALL_K3S_EXEC k3s server var

lmaraite:
This adds the possibility to customize the INSTALL_K3S_EXEC environment variable for the k3s server daemon.
This can be useful to pass additional arguments, for example api server arguments, to enable oidc or other stuff.

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -45,7 +45,7 @@ locals {
   apply_k3s_selinux = ["/sbin/semodule -v -i /usr/share/selinux/packages/k3s.pp"]
 
   install_k3s_server = concat(local.common_commands_install_k3s, [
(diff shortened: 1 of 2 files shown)


mysticaltech:
@lmaraite Thanks for this, please provide some more context and examples. We already have `control_planes_custom_config`, are you sure it's cannot be done with that?

lmaraite:
> @lmaraite Thanks for this, please provide some more context and examples. We already have `control_planes_custom_config`, are you sure it's cannot be done with that?

Ok I'm sorry I actually had not seen this argument. This covers all use cases I've wanted to implement with this feature. So thanks for your help :)

PR #611, opened 2023-02-27
feat: agent_nodepools: add floating IPs option

JurgenTreeum:
This modification is primarily intended to enable the implementation of the Cilium egress gateway (https://docs.cilium.io/en/v1.12/gettingstarted/egress-gateway) and allows assigning a unique floating IP to each node in the selected node pool, as well as configuring routing through the assigned floating IP.

nodepool example:

```
{
  name        = "egress",
  server_type = "cpx11",
  location    = "fsn1",
  labels = [
    "node.kubernetes.io/role=egress"
  ],
  taints = [
    "node.kubernetes.io/role=egress:NoSchedule"
  ],
  floating_ip = true # <--
  count = 1
},
```

Brief summary of the changes made:

An optional floating_ip parameter has been added to the agent_nodes parameters (default value: false).

A floating IP is leased and assigned to each node in the selected node pool.

The following actions are performed via remote-exec on each node in the selected node pool:
A new network interface configuration is saved for ETH0 (+ routing through the floating IP).

```
 # /etc/sysconfig/network/ifcfg-eth0:
 BOOTPROTO='static'
 STARTMODE='auto'
 # its important: floating-ip should be the main IP
 IPADDR={FLOATING_IP}/32
 IPADDR_1={NODE_PRIMARY_IP}/32
```

```
# /etc/sysconfig/network/ifroute-eth0:
172.31.1.1 - 255.255.255.255 eth0
default 172.31.1.1 - eth0 src {FLOATING_IP}

```
Applied immediately to the running OS:

```
ip addr add {FLOATING_IP}/32 dev eth0
ip route replace default via 172.31.1.1 dev eth0 src {FLOATING_IP}

# its important: floating IP should be first on the interface IP list, move main IP to the second position
ip addr del {NODE_PRIMARY_IP}/32 dev eth0
ip addr add {NODE_PRIMARY_IP}/32 dev eth0
```

Additionally:
When leasing a floating IP, a protection flag is set to prevent accidentally losing the leased floating IP during future cluster reconfigurations (such as reassigning the floating IP to another node pool). This work is done manually using terraform state rm / terraform import.
If it is necessary to remove the configuration with the floating IP, it is necessary to manually remove the protection flag for the floating IP via the Hetzner console.

### DIFF
--- a/agents.tf
+++ b/agents.tf
@@ -144,8 +144,6 @@ resource "hcloud_floating_ip" "agents" {
   type          = "ipv4"
   labels        = local.labels
   home_location = each.value.location
-
-  delete_protection = true
 }
 
 resource "hcloud_floating_ip_assignment" "agents" {
--- a/agents.tf
+++ b/agents.tf
@@ -137,3 +137,61 @@ resource "null_resource" "configure_longhorn_volume" {
     hcloud_volume.longhorn_volume
   ]
 }
+
+resource "hcloud_floating_ip" "agents" {
+  for_each = { for k, v in local.agent_nodes : k => v if(lookup(v, "floating_ip", false)) }
+
+  type          = "ipv4"
+  labels        = local.labels
+  home_location = each.value.location
+
+  delete_protection = true
+}
+
+resource "hcloud_floating_ip_assignment" "agents" {
+  for_each = { for k, v in local.agent_nodes : k => v if(lookup(v, "floating_ip", false)) }
+
+  floating_ip_id = hcloud_floating_ip.agents[each.key].id
+  server_id      = module.agents[each.key].id
+
+  depends_on = [
+    null_resource.agents
+  ]
+}
+
+resource "null_resource" "configure_floating_ip" {
+  for_each = { for k, v in local.agent_nodes : k => v if(lookup(v, "floating_ip", false)) }
+
+  triggers = {
+    agent_id = module.agents[each.key].id
+    floating_ip_id = hcloud_floating_ip.agents[each.key].id
+  }
+
+  # Start the k3s agent and wait for it to have started
+  provisioner "remote-exec" {
+    inline = [
+      "echo \"BOOTPROTO='static'\nSTARTMODE='auto'\nIPADDR=${hcloud_floating_ip.agents[each.key].ip_address}/32\nIPADDR_1=${module.agents[each.key].ipv4_address}/32\" > /etc/sysconfig/network/ifcfg-eth0",
+      "echo \"172.31.1.1 - 255.255.255.255 eth0\ndefault 172.31.1.1 - eth0 src ${hcloud_floating_ip.agents[each.key].ip_address}\" > /etc/sysconfig/network/ifroute-eth0",
+
+      "ip addr add ${hcloud_floating_ip.agents[each.key].ip_address}/32 dev eth0",
+      "ip route replace default via 172.31.1.1 dev eth0 src ${hcloud_floating_ip.agents[each.key].ip_address}",
+
+      # its important: floating IP should be first on the interface IP list
+      # move main IP to the second position
+      "ip addr del ${module.agents[each.key].ipv4_address}/32 dev eth0",
+      "ip addr add ${module.agents[each.key].ipv4_address}/32 dev eth0",
+    ]
+  }
+
+  connection {
+    user           = "root"
+    private_key    = var.ssh_private_key
+    agent_identity = local.ssh_agent_identity
+    host           = module.agents[each.key].ipv4_address
+    port           = var.ssh_port
+  }
+
+  depends_on = [
+    hcloud_floating_ip_assignment.agents
+  ]
+}
--- a/agents.tf
+++ b/agents.tf
@@ -163,11 +163,10 @@ resource "null_resource" "configure_floating_ip" {
   for_each = { for k, v in local.agent_nodes : k => v if(lookup(v, "floating_ip", false)) }
 
   triggers = {
-    agent_id = module.agents[each.key].id
+    agent_id       = module.agents[each.key].id
     floating_ip_id = hcloud_floating_ip.agents[each.key].id
   }
 
-  # Start the k3s agent and wait for it to have started
   provisioner "remote-exec" {
     inline = [
       "echo \"BOOTPROTO='static'\nSTARTMODE='auto'\nIPADDR=${hcloud_floating_ip.agents[each.key].ip_address}/32\nIPADDR_1=${module.agents[each.key].ipv4_address}/32\" > /etc/sysconfig/network/ifcfg-eth0",
--- a/README.md
+++ b/README.md
@@ -50,6 +50,8 @@ To achieve this, we built up on the shoulders of giants by choosing [openSUSE Mi
 - [x] Possibility to toggle **Longhorn** and **Hetzner CSI**.
 - [x] Choose between **Flannel, Calico, or Cilium** as CNI.
 - [x] Optional **Wireguard** encryption of the Kube network for added security.
+- [x] Optional assign [Floating IPs](https://docs.hetzner.com/cloud/floating-ips/faq) (with default routing via Floating IP) to a selected nodepool.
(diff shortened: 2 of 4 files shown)


agents.tf
```diff
@@ -137,3 +137,61 @@ resource "null_resource" "configure_longhorn_volume" {
...
+
+      "ip addr add ${hcloud_floating_ip.agents[each.key].ip_address}/32 dev eth0",
+      "ip route replace default via 172.31.1.1 dev eth0 src ${hcloud_floating_ip.agents[each.key].ip_address}",
+
+      # its important: floating IP should be first on the interface IP list
+      # move main IP to the second position
+      "ip addr del ${module.agents[each.key].ipv4_address}/32 dev eth0",
+      "ip addr add ${module.agents[each.key].ipv4_address}/32 dev eth0",
+    ]
+  }
```
mysticaltech, line 184: If that config can be moved to cloud-init it would be best, as for other setups it lives there already. And conditional needs to be conditional on the need for it of course.

mysticaltech:
@JurgenTreeum Very interesting feature, thanks for that. If you could add some more explanation to the kube.tf it would be best, especially what needs to happen when scaling the nodepool up or down, to prevent failure.

Also, did you actually try the cilium egress feature? Because k3s is configured to work on with eth1, if you want it also to work with eth1, you will need to add it to devices in the k3s config.

Also, I see the need for a global flag to enable or disable that feature, unless we can really solely on the nodepool flag that defaults to false. Up to you, let me know what you think.

Lastly, please do test it well, and confirm that it's working as intended. If manual work can be reduced it should be.

agents.tf
JurgenTreeum: We considered the option of putting this logic into the host module - cloud-init, but the following conditions are not met at the server creation stage:

    The floating IP is not yet assigned to the server (as the server has not been created yet).
    We don't know the primary IP that will be assigned to the server.

Both the Floating-IP and primary IP are required to be specified in the /etc/sysconfig/network/ifcfg-eth0 configuration, and these changes can only be made after the server is ready.

JurgenTreeum:
> @JurgenTreeum Very interesting feature, thanks for that. If you could add some more explanation to the kube.tf it would be best, especially what needs to happen when scaling the nodepool up or down, to prevent failure.

Yes, we will add an extended description later.

> Also, did you actually try the cilium egress feature? Because k3s is configured to work on with eth1, if you want it also to work with eth1, you will need to add it to devices in the k3s config.

Yes, we have tried the cilium egress gateway and it works as intended by cilium authors. To do this, you need to specify the following cilium configuration:
`cilium_values = <<EOT
ipam:
  operator:
    clusterPoolIPv4PodCIDRList:
      - 10.42.0.0/16
kubeProxyReplacement: strict
l7Proxy: "false"
bpf:
  masquerade: "true"
egressGateway:
  enabled: "true"
EOT`

And here is what the automatically generated (by the kube-hetzner/terraform-hcloud-kube-hetzner module) cilium configuration looks like:

`devices: eth1
ipam:
  operator:
    clusterPoolIPv4PodCIDRList:
    - 10.42.0.0/16
`


> Also, I see the need for a global flag to enable or disable that feature, unless we can really solely on the nodepool flag that defaults to false. Up to you, let me know what you think.

We believe that such functionality should be implemented by separately dedicated worker pools and left to the discretion of the DevOps engineer. For some, a single node with a single Floating IP may be sufficient, while others may want to create multiple worker pools (each with several nodes, each with its own Floating IP) and distribute them across different availability zones.

> Lastly, please do test it well, and confirm that it's working as intended. If manual work can be reduced it should be.

Here is the minimal configuration for testing the functionality:
`locals {
  cluster_ipv4_cidr = "10.42.0.0/16"
}

module "kube-hetzner" {
  providers = {
    hcloud = hcloud
  }

  source = "github.com/treeumcorp/terraform-hcloud-kube-hetzner?ref=feat%2Fagent-pool-floating-ips"

  ssh_port = 64022

  ssh_public_key  = file("~/.ssh/id_ed25519.pub")
  ssh_private_key = file("~/.ssh/id_ed25519")

  cluster_ipv4_cidr = local.cluster_ipv4_cidr

  control_plane_nodepools = [
    {
      server_type = "cpx11",
      location    = "fsn1",
      labels      = [],
      taints      = [],
      count       = 1
    },
  ]

  agent_nodepools = [
    {
      name        = "mainpool",
      server_type = "cpx11"
      location    = "fsn1",
      floating_ip = false
      labels      = [],
      taints      = [],
      count       = 1
    },
    {
      name        = "egresspool",
      server_type = "cpx11"
      location    = "fsn1",
      floating_ip = true
      labels = [
        "node.kubernetes.io/role=egress"
      ],
      taints = [
        "node.kubernetes.io/role=egress:NoSchedule"
      ],
      count = 1
    },
  ]
  ingress_controller        = "none"
  automatically_upgrade_k3s = false
  automatically_upgrade_os  = false
  initial_k3s_channel       = "v1.24"
  restrict_outbound_traffic = false
  cni_plugin                = "cilium"
  create_kubeconfig         = true
  create_kustomization      = false
  cilium_values             = <<EOT
ipam:
  operator:
    clusterPoolIPv4PodCIDRList:
      - ${local.cluster_ipv4_cidr}
kubeProxyReplacement: strict
l7Proxy: "false"
bpf:
  masquerade: "true"
egressGateway:
  enabled: "true"
EOT
}

provider "hcloud" {
  token = var.hcloud_token
}

terraform {
  required_version = ">= 1.3.3"
  required_providers {
    hcloud = {
      version = ">= 1.35.2"
    }
  }
}

output "kubeconfig" {
  sensitive = true
}

variable "hcloud_token" {
  sensitive = true
}`

After deploying the cluster, we launch a test POD and check the external IP from which the test POD exits to the external network:

`apiVersion: v1
kind: Pod
metadata:
  name: mediabot
  labels:
    org: empire
    class: mediabot
spec:
  containers:
    - name: mediabot
      image: docker.io/tgraf/netperf
`

`kubectl exec mediabot -- wget -q -O - http://ifconfig.me/`

We make sure that the IP in the response belongs to the 0th node in the mainpool.

The next step is to deploy the CiliumEgressGatewayPolicy:

`apiVersion: cilium.io/v2
kind: CiliumEgressGatewayPolicy
metadata:
  name: example-egress-policy
spec:
  destinationCIDRs:
    - 0.0.0.0/0
  egressGateway:
    nodeSelector:
      matchLabels:
        node.kubernetes.io/role: egress

    # Specify the IP address used to SNAT traffic matched by the policy.
    # It must exist as an IP associated with a network interface on the instance.
    # egressIP: {FLOATING_IP}

    # Alternatively it's possible to specify the interface to be used for egress traffic.
    # In this case the first IPv4 assigned to that interface will be used as egress IP.
    interface: eth0

  selectors:`

Then  repeat the check for the external IP:

`kubectl exec mediabot -- wget -q -O - http://ifconfig.me/`

Then make sure that the IP in the response corresponds to the rented FLOATING IP.

P.S. We are actively implementing this change in our infrastructure and are already testing it in practice. It will take some time to confirm its stable operation.

mysticaltech:
@JurgenTreeum Perfect! Basically, you can use the docs folder if you want to create a small md file with all the explanations, and link to it from from kube.tf, that would be super useful. If you feel the need for it, if not just linking to the cilium doc would be fine too, but all the examples you show above, including your cilium config and test kube.tf do justify having a small explanation doc for it!

Just reformatting the explanations you gave above in proper markdown for instance would be enough.

Then we are good to go! Thanks again.

ifeulner:
@JurgenTreeum nice feature, thanks a lot for contribution! i am thinking if this feature can be used to avoid to have public IPs on all nodes (only on the egress nodes then)...?
Problem is that we need internet access during initial provisioning of the nodes, that still remains... anyway, really nice feature

JurgenTreeum:
> 

Hi. Formatting fixed.

mysticaltech:
@JurgenTreeum Did a few tweaks and released under v1.9.8, thanks! 🙏

JurgenTreeum:
> @JurgenTreeum Did a few tweaks and released under v1.9.8, thanks! pray

Thank you too

PR #617, opened 2023-02-28
Added backup option for hcloud servers

frknakk:
I was missing a function to enable automatic backups in Hetzner, so here it is :)

Is it possible to make variables optional? I started using terraform a week ago and i couldn't find an option for that.

### DIFF
--- a/modules/host/variables.tf
+++ b/modules/host/variables.tf
@@ -73,7 +73,7 @@ variable "server_type" {
   type        = string
 }
 
-variable "enable_backups" {
+variable "backups" {
   description = "Enable automatic backups via Hetzner"
   type        = bool
   default     = false
--- a/modules/host/variables.tf
+++ b/modules/host/variables.tf
@@ -73,9 +73,10 @@ variable "server_type" {
   type        = string
 }
 
-variable "backups" {
-  description = "Enable backups"
+variable "enable_backups" {
+  description = "Enable automatic backups via Hetzner"
   type        = bool
+  default     = false
 }
 
 variable "packages_to_install" {
--- a/modules/host/variables.tf
+++ b/modules/host/variables.tf
@@ -73,6 +73,11 @@ variable "server_type" {
   type        = string
 }
 
+variable "backups" {
+  description = "Enable backups"
+  type        = bool
+}
+
 variable "packages_to_install" {
   description = "Packages to install"
   type        = list(string)
--- a/agents.tf
+++ b/agents.tf
@@ -18,7 +18,7 @@ module "agents" {
   placement_group_id           = var.placement_group_disable ? 0 : hcloud_placement_group.agent[floor(each.value.index / 10)].id
   location                     = each.value.location
   server_type                  = each.value.server_type
-  enable_backups               = each.value.enable_backups
+  backups                      = each.value.backups
   ipv4_subnet_id               = hcloud_network_subnet.agent[[for i, v in var.agent_nodepools : i if v.name == each.value.nodepool_name][0]].id
   packages_to_install          = local.packages_to_install
   dns_servers                  = var.dns_servers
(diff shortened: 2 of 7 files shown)


aleksasiriski:
> I was missing a function to enable automatic backups in Hetzner, so here it is :)
> 
> Is it possible to make variables optional? I started using terraform a week ago and i couldn't find an option for that.

you can set the default to true and add explanation in kube.tf.example with commented option

modules/host/variables.tf
```diff
@@ -73,6 +73,11 @@ variable "server_type" {
   type        = string
 }
 
+variable "backups" {
+  description = "Enable backups"
+  type        = bool
+}
+
```
ianwesleyarmstrong, line 80:
```suggestion terraform
variable "backups" {
  description = "Enable backups"
  type             = bool
  default         = false
}
```

Not sure why the whitespace in the preview is messed up, but this is how you set optional/default variables!

modules/host/variables.tf
```diff
@@ -73,6 +73,11 @@ variable "server_type" {
   type        = string
 }
 
+variable "backups" {
```
ianwesleyarmstrong, line 76: Would the variable name `enable_backups` be clearer about it's intent?

modules/host/variables.tf
aleksasiriski: > Not sure why the whitespace in the preview is messed up, but this is how you set optional/default variables!

Just run `terraform fmt -recursive`

modules/host/variables.tf
aleksasiriski: > Would the variable name `enable_backups` be clearer about its intent?

Yes please

modules/host/variables.tf
ianwesleyarmstrong: Definitely required after applying the suggestion, but I was commenting on how the PR suggestion UI handles the whitespace. If you move the cursor up/down, the variables are all aligned, but not in the preview /shrug

Screenshot 2023-02-28 at 6 20 37 PM

frknakk:
Renamed the variable to "enable_backups" & made it optional. But i also had to add an object definition of control_plane/agent nodes and make the variable optional in there too. Now it's working, at least for me.

Is this a good solution to this?

modules/host/variables.tf
aleksasiriski: One line is probably with spaces and the other is using TAB which goes to exact width as the spaces, but is interpreted different in the posted comment.

mysticaltech:
@frknakk Thanks for this. If we have the variable at the nodepool level, then we do not need to have it at the global level too, it does not make sense to me. Best keep it at the nodepool level.

Also, I am sorry you have to do the change, please revert it back to backups, as this is what is used in the [hcloud_server](https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs/resources/server) module, and we have all other params using the same names as that. 

When the changes above are good and tested, we can merge! 🚀

frknakk:
> @frknakk Thanks for this. If we have the variable at the nodepool level, then we do not need to have it at the global level too, it does not make sense to me. Best keep it at the nodepool level.
> 
> Also, I am sorry you have to do the change, please revert it back to backups, as this is what is used in the [hcloud_server](https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs/resources/server) module, and we have all other params using the same names as that.
> 
> When the changes above are good and tested, we can merge! 🚀

If i remove the variable from `modules/host/variables.tf` i get this error:
Error Message

As far as i understand this is the flow:
global variable `control_plane_nodepools` (in /kube.tf, variables defined by /variables.tf)
-> copied via for-in into local variable `control_plane_nodes` (in /locals.tf)
-> copied via for_each into module `control_planes` (in /control_planes.tf)
-> module `control_planes` is an instance of the file /modules/host/main.tf (variables defined by /modules/host/variables.tf)

Is that correct? I think in that case the variable has to be defined in both `/variables.tf` and `/modules/host/variables.tf`

mysticaltech:
@frknakk It is correct! It was confusion on my end, it is indeed only at the nodepool level, so perfect!

mysticaltech (approved):

aleksasiriski (approved):

PR #632, opened 2023-03-09
Update Hetzner cloud API IP address

tripadvisor101:
Since currently dynamic IP gathering logic from DNS is not implemented yet, updating API IP manually to a new value.


mysticaltech:
Thanks @tripadvisor101, indeed! However please see https://github.com/hetznercloud/csi-driver/issues/204#issuecomment-1457820457.

Turns out what we have trying to clear it for IN traffic is not needed, as it's only used for out traffic, and HTTP out traffic is already allowed by default for any IPs. However, until it is removed from the codebase, having the correct value is better! 

Thank you, will merge and release ASAP 🙏

PR #638, opened 2023-03-10
Fix/longhorn encryption at rest

valkenburg-prevue-ch:
This was one of the most painful debugging sessions, with an insanely small code-change to fix it. Please pad me on the back here.. 

Longhorn encryption at rest now works correctly! cryptsetup does not hang, volumes get created and encrypted and mounted. The key was installing a >10 year old package lvm2, which creates the right udev rules.

See here for some digging: https://github.com/longhorn/longhorn/discussions/5332

This solves the open question of encrypting longhorn volumes, https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/discussions/448 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ To achieve this, we built up on the shoulders of giants by choosing [openSUSE Mi
 - [x] Can use Klipper as an **on-metal LB** or the **Hetzner LB**.
 - [x] Ability to **add nodes and nodepools** when the cluster is running.
 - [x] Possibility to toggle **Longhorn** and **Hetzner CSI**.
+- [x] Encryption at rest fully functional in both **Longhorn** and **Hetzner CSI**.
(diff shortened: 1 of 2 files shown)


mysticaltech:
@valkenburg-prevue-ch Beautiful!! Well done and clear examples... 🥳 🙏

FYI @ifeulner @JustinGuese

mysticaltech:
Released in v1.10.1, thanks again @valkenburg-prevue-ch 🩵 🚀

JustinGuese:
Yeah nice!! Will test, definitely helpful...

JustinGuese:
So basically lvm2 was missing?

valkenburg-prevue-ch:
@JustinGuese Yeah, exactly, that's literally all it was. But figuring that out was incredibly tough. Usually you get an error message somewhere and then the first hit on Google is your solution. Not this time.

PR #660, opened 2023-03-17
added workaround to avoid empty k3s configs in screenshot

bulnv:
Added workaround to preserve k3s configuration files from deletion with k3s-uninstall

### DIFF
--- a/templates/autoscaler-cloudinit.yaml.tpl
+++ b/templates/autoscaler-cloudinit.yaml.tpl
@@ -64,13 +64,13 @@ write_files:
     X3QBAAAAAQAAAAEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA==
   path: /etc/selinux/sshd_t.pp
 
-- content: ${base64encode(k3s_config)}
+- content: ${k3s_config}
   encoding: base64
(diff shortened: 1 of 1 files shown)


mysticaltech:
@bulnv Are you sure about this, did this change fix it for you?

Because write_files in cloud-init will create the required path if missing. But maybe indeed, the k3s uninstallation will delete it... If that's the case, it would explain it.

mysticaltech:
Then it's a matter of timing depending on the node. Great catch!! Will merge and do final testing later on.

ifeulner:
Was a few days off, so I just read now this thread - but can't see really what this patch changes? What do you mean with timing? There could indeed be an issue dependent on when the snapshot has been created... unfortunately the current snapshot generation is unreliable - this is not the case anymore when the snapshot is created from the outside, e.g. via packer.

mysticaltech:
@ifeulner No worries. Now it's realiable, I just had changed the location of the k3s config file (you had it correct in the first place, tmp was the right place), and it got deleted by the k3s uninstall on some nodes depending on the node size I guess, so timing. This PR is correct and restores the bug that I introduced during another fix.

mysticaltech (approved):

PR #661, opened 2023-03-17
Set allow_scheduling_on_control_plane before using it

mattanja:
Before this, the label exclude-from-external-load-balancers is set to true on control planes even if allow_scheduling_on_control_plane is set to true in the kube.tf file.

Please note that this is completely untested since I'm new to terraform. Since this is the only location of the label in the project, I think it should be correct though.


mattanja:
Sorry, read the contribute description after opening this PR. Will reopen another PR according to your guideline.

PR #662, opened 2023-03-17
Fix/invert loadbalancer label on control planes

mattanja:
Hi, I can't find the "staging" branch mentioned in the contribution section of the readme, so I'm targeting master for now?

I've had the issue of the label `node.kubernetes.io/exclude-from-external-load-balancers` being set to true with `allow_scheduling_on_control_plane=true` in my configuration. After searching for the issue it seem to me that the conditions are not being set correctly.
There are two separate concerns mixed: the loadbalancer targeting the control planes (set via an "exclude" meaning inverted label) and the scheduling of load on the control planes (set via the ":NoSchedule" taint).

The condition for the loadbalancer label seems to have been inverted. (I'm not sure how it would have been possible that this was not detected before, which makes me doubt my solution.)

I have now tested the cluster setup with 3 control planes with `allow_scheduling_on_control_plane=true` and the label was set to `node.kubernetes.io/exclude-from-external-load-balancers: "false"` which should be correct. However, the LoadBalancer still did not target the servers. Maybe there are additional issues?

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -101,10 +101,14 @@ locals {
   # disable k3s extras
   disable_extras = concat(["local-storage"], local.using_klipper_lb ? [] : ["servicelb"], ["traefik"], var.enable_metrics_server ? [] : ["metrics-server"])
 
+  # Determine if scheduling should be allowed on control plane nodes, which will be always true for single node clusters and clusters using the klipper lb or if scheduling is allowed on control plane nodes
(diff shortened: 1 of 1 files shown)


mysticaltech:
@mattanja Thanks for noticing, I will investigate and test your changes.

mysticaltech:
@mattanja Your fix was accurate! Thanks for this. It's been released in 1.10.6 🙏🚀

If you still do not see the control plane as target list, please try to see if there are any remaining NoSchedule taints on those. Normally there should not be when allow_scheduling_on_control_planes is set to true!

Let me know, and don't hesitate to open an issue to investigate this deeper, or provide an additional fix.

PR #689, opened 2023-03-30
Update variables.tf

Wuppie007:
Apparently k8s only support 3 entries for DNS servers. 

To avoid these warning "Nameserver limits were exceeded, some nameservers have been omitted, the applied nameserver line is: 8.8.8.8 8.8.4.4 1.1.1.1"

randomly removed the last entry ;)

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -472,7 +472,7 @@ variable "use_control_plane_lb" {
 
 variable "dns_servers" {
   type        = list(string)
-  default     = ["8.8.8.8", "8.8.4.4", "1.1.1.1", "1.0.0.1"]
+  default     = ["8.8.8.8", "8.8.4.4", "1.1.1.1"]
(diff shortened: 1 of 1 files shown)


mysticaltech:
@Wuppie007 Thanks for this. Please don't hesitate to tweak the `kube.tf.example` file, and the description of the variable to inform of that limit. 

If you are feeling adventurous, see how validations are done for some other variables (or just use chatgpt 4 🤐) to add a validation for 3 max values, it's just another attribute in the variable object.

aleksasiriski:
I would argue to use both cloudflare's DNS servers and one Google, or even Quad9 as the third instead of Google

Wuppie007:
Updated now for three different companies: ["8.8.8.8","1.1.1.1","9.9.9.9"]

mysticaltech:
Released in v2.0.2, thanks for your contribution both. FYI, I finally opted for @aleksasiriski's recommendation to put Cloudflare first.

PR #693, opened 2023-04-01
added cifs-utils and csi-driver-smb

aleksasiriski:
For Hetzner Storage Boxes to work correctly with [this csi driver](https://github.com/kubernetes-csi/csi-driver-smb) [cifs-utils](https://software.opensuse.org/package/cifs-utils) package is needed. Since nfs-utils and iscsi are installed regardless if longhorn is used or not, I would argue a small package like cifs-utils should be preinstalled by default as well.

EDIT: I went ahead and added csi-driver-smb as optional Helm chart, as I already said this can be leveraged by Hetzner Storage Boxes so I believe it is a welcome addition to the project. Also, now it makes more sense to preinstall cifs-utils by default.


locals.tf
```diff
@@ -390,6 +390,8 @@ persistence:
   %{if var.disable_hetzner_csi~}defaultClass: true%{else~}defaultClass: false%{endif~}
   EOT
 
+  csi_driver_smb_values = var.csi_driver_smb_values
```
mysticaltech, line 393: Please, let's follow the same as other helm charts, default from locals gets used if custom not provided.

templates/ccm.yaml.tpl
```diff
@@ -1,3 +1,4 @@
+---
```
mysticaltech, line 1: Just curious about those added everywhere, why do you feel adding those are a good practice?

mysticaltech (changes requested):
Hey @aleksasiriski, looks good, just two little questions, and please don't forget to add this to `kube.tf.example`. Thanks! 🚀

locals.tf
aleksasiriski: The default from locals should be empty string, so I just let it copy from var since by default it's empty. But I'll fix it so it looks more unified.

templates/ccm.yaml.tpl
aleksasiriski: My personal preference is to not have them at the start of files, but I noticed that most of the files had them so I just made everything the same.

aleksasiriski:
I've tested these changes succesfully, whole cluster build and deploy.

aleksasiriski:
I've encountered an issue..

Old nodes made by v1 work flawlessly with this setup, smb share mounts to host and I can list files and container get PVC mounted and can list files. The setup for v1 was identical except that I used old `additional_packages` var to add `cifs-utils`.

Now that I made this PR and upgraded my cluster, and added 2 nodes with packer from v2, even though `cifs-utils` is installed and I can list the files from smb share on the host without issue, if a pod is run on the new nodes it cannot access files from the smb share.

TLDR:
1. cifs-utils is installed correctly
2. smb mounts to both v1 and v2 hosts correctly when making a PV and PVC, files are accessable
3. pods on v1 work as expected, pods on v2 error out with (exactly the same pods, I just set antiaffinity for them and made replicas to equal number of nodes):
```
ls: cannot open directory '/var/storage/media': Permission denied
```

aleksasiriski:
I've tested this, making a cluster with v1.10.9 and deploying the same Helm chart results in perfectly working Storage Box setup (note that I'm not even installing cifs-utils, just a normal cluster without Longhorn even) with my Autoscan deployment.

So something changed when the script switched to packer method. I'll investigate further...

aleksasiriski:
@mysticaltech Could it be related to all these SELinux additions I see in the commit history?

aleksasiriski:
kube-hetzner v2.0.0 works without issues, I'm testing to see if 2.0.1 works but I believe the culprit is there and not in 2.0.2

mysticaltech:
@aleksasiriski Yes, you guess right, SELinux was activated for k3s in 2.0.1. But it's super easy to fix. 

Run this command sudo `grep "avc:  denied" /var/log/audit/audit.log | audit2allow -M my_custom_policy && mv my_custom_policy.te my_custom_policy_all_denied.te`
, grab the content of the .te, and add the the kube-hetzner policy in the locals.tf. Search for "policy", you will see it. Then rebuild.

But ideally, apply the .pp with semodule -i to test if it works, and if it generates other error, run the above command again, until it works for good and only then copy to the locals.tf policy, then delete the snapshot (you can use the cleanup script) and try again. It should work like a charm!

aleksasiriski:
The issue is mentioned here, I'll try adapting kube-hetzner's SELinux to work with smb shares, as I would like for it to stay enabled. https://github.com/kubernetes-csi/csi-driver-smb/issues/132#issuecomment-709917748

mysticaltech:
It will work, super easy flow, you'll see. If you just append the content it's fine. I can organize it better with my friend gpt-4 later on 😜

aleksasiriski:
This policy doesn't make sense to me, I guess I need to manually set `virt_use_samba` SELinux boolean to ON and then generate the current policy to a file?

```
module my_custom_policy 1.0;

require {
	type kernel_t;
	type bin_t;
	type kernel_generic_helper_t;
	class key { read view };
	class file execute_no_trans;
	class unix_dgram_socket create;
}

#============= kernel_generic_helper_t ==============
allow kernel_generic_helper_t bin_t:file execute_no_trans;
allow kernel_generic_helper_t kernel_t:key { read view };
allow kernel_generic_helper_t self:unix_dgram_socket create;
```

aleksasiriski:
Setting `setsebool -P virt_use_samba 1` manually fixes the issue, but I have to find out how to convert that to a policy entry

aleksasiriski:
https://unix.stackexchange.com/questions/330861/is-it-possible-to-enable-an-selinux-boolean-in-a-policy-module

I guess it must be a shell command?

aleksasiriski:
Just created a whole new cluster and it finally works without issues! You can give it another look if you want, but I believe it's ready to be merged into 2.0.3

mysticaltech:
Perfect, will do tomorrow morning ASAP, good work! 🥳

mysticaltech (approved):

aleksasiriski:
Oh yeah, will remember that. Let's not be Kubernetes...
image

mysticaltech:
> This policy doesn't make sense to me, I guess I need to manually set `virt_use_samba` SELinux boolean to ON and then generate the current policy to a file?
> 
> ```
> module my_custom_policy 1.0;
> 
> require {
> 	type kernel_t;
> 	type bin_t;
> 	type kernel_generic_helper_t;
> 	class key { read view };
> 	class file execute_no_trans;
> 	class unix_dgram_socket create;
> }
> 
> #============= kernel_generic_helper_t ==============
> allow kernel_generic_helper_t bin_t:file execute_no_trans;
> allow kernel_generic_helper_t kernel_t:key { read view };
> allow kernel_generic_helper_t self:unix_dgram_socket create;
> ```

This does not seem to pose an obvious security risk, so will add them too. I want to make sure that our system is not overly restrictive while  we maintain SELinux on (voluntarily did not create an option to switch it off). 

Also see https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/697, we are now mapping the playing field of needed permissions to operate standard kubernetes tools.

PR #702, opened 2023-04-05
K3s v1.26

aleksasiriski:


mysticaltech:
@aleksasiriski Just FYI, the importance of testing cannot be understated. Turns out the autoscaler v1.26.0 had a segmentation error when trying to scale up. v1.26.2 on the other hand works. Pushing that now!

aleksasiriski:
>  @aleksasiriski Just FYI, the importance of testing cannot be understated. Turns out the autoscaler v1.26.0 had a segmentation error when trying to scale up. v1.26.2 on the other hand works. Pushing that now!

Sorry for that, but I did ask above if we should be using latest patch release

mysticaltech:
@aleksasiriski Ah my bad, yes, that would be a good idea in the future. Also, it would be cool at some point to migrate the autoscaler deployment to helm. Which I guess will make it easier to upgrade post deploy!

mysticaltech:
Also, hopefully we can deploy some basic automatic testing in the future, from Github actions directly! 🤞 And passing the checks would be a requirement before being allowed to merge by the system.

PR #706, opened 2023-04-09
add k3s ver to null resource kustomization

aleksasiriski:
`terraform apply` now updates the k3s ver without errors

I just had a question: is it possible to have a channel for cluster autoscaler as well? So we aren't fixated on a patch version but instead always pull latest?

### DIFF
--- a/init.tf
+++ b/init.tf
@@ -100,6 +100,7 @@ resource "null_resource" "kustomization" {
     ])
     # Redeploy when versions of addons need to be updated
     versions = join("\n", [
+      coalesce(var.initial_k3s_channel, "N/A"),
       coalesce(var.cluster_autoscaler_version, "N/A"),
       coalesce(var.hetzner_ccm_version, "N/A"),
       coalesce(var.hetzner_csi_version, "N/A"),


mysticaltech (approved):

mysticaltech:
Good to see this works! 

About the cluster autoscaler, I have been thinking about this too. The problem is that it does not have "channels", but instead like you say, we could just remove the tag so that it fetches the latest and greatest. I think this would work great!

mysticaltech:
I tried without the tags, it doesn't work. I guess we will have to change to helm to fix the autoscaler version thing, because the version is set by default in the default helm values.yaml.

nikolaszimmermann:
NOTE: I'm closely tracking upstream progress, and would wish for more helpful commit messages -- I can deduce that the k3s version was added to versions field of the kustomizaton null_resource from the commit, the intersting part: why? is only answered here, and not in the commit message.

No criticisim, just asking to make my downstream life easier :-)

aleksasiriski:
> NOTE: I'm closely tracking upstream progress, and would wish for more helpful commit messages -- I can deduce that the k3s version was added to versions field of the kustomizaton null_resource from the commit, the intersting part: why? is only answered here, and not in the commit message.
> 
> No criticisim, just asking to make my downstream life easier :-)

I was lazy, will have to change that habit. Noted.

mysticaltech:
Great tip @nikolaszimmermann! We will improve.

@aleksasiriski FYI https://github.com/Nutlope/aicommits

PR #722, opened 2023-04-14
Add use_arm option to packer template

### LINKED ISSUE #716: Feature Request: Add Support for Hetzner's New ARM Servers - CAX11, CAX21, CAX31, CAX41
### Description

Request to add new feature in kube-hetzner.  

Support for Hetzner's newly released **ARM** server lineup: **CAX11, CAX21, CAX31, and CAX41.** I believe that incorporating support for these new servers will greatly benefit the community and enhance the overall user experience.

Kakadus:
This adds another variable to the packer template, so that users can easily enable an arm packer build.
closes #716

### DIFF
--- a/packer-template/hcloud-microos-snapshot.pkr.hcl
+++ b/packer-template/hcloud-microos-snapshot.pkr.hcl
@@ -22,7 +22,7 @@ variable "opensuse_microos_mirror_link_x86_64" {
 
 variable "opensuse_microos_mirror_link_arm64" {
   type    = string
-  default = "https://ftp.gwdg.de/pub/opensuse/ports/aarch64/tumbleweed/appliances/openSUSE-MicroOS.aarch64-16.0.0-OpenStack-Cloud-Snapshot20230403.qcow2"
+  default = "https://ftp.gwdg.de/pub/opensuse/ports/aarch64/tumbleweed/appliances/openSUSE-MicroOS.aarch64-OpenStack-Cloud.qcow2"
 }
 
 # If you need to add other packages to the OS, do it here in the default value, like ["vim", "curl", "wget"]
--- a/packer-template/hcloud-microos-snapshot.pkr.hcl
+++ b/packer-template/hcloud-microos-snapshot.pkr.hcl
@@ -2,6 +2,11 @@
  * Creates a MicroOS snapshot for Hetzner Cloud
  */
 
+variable "use_arm" {
+  type = bool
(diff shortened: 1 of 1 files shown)


Kakadus:
It tested it, packer successfully created an arm64 snapshot

aleksasiriski:
This option shouldn't be needed in my opinion, we should by default make both x86 and arm snapshot and autodetect if a node is CAX or not and run the corresponding logic. This would allow to easily have hybrid cluster if someone wants that functionality.

I'll try to implement it today, but PRs are always appreciated just like this one (:

korewaChino:
Just tried this, seems to fail deploying due to it still trying to apply an x86 snapshot

aleksasiriski:
I'm working on it right now, will need testers later.

Kakadus:
> Just tried this, seems to fail deploying due to it still trying to apply an x86 snapshot

Jeah, I was simply focussing on packer

Kakadus:
> > Just tried this, seems to fail deploying due to it still trying to apply an x86 snapshot
> 
> Jeah, I was simply focussing on packer

After reapplying with terraform, the cluster was created successfully. Did you change the server types in your `kube.tf`?

ichbestimmtnicht:
> This option shouldn't be needed in my opinion, we should by default make both x86 and arm snapshot and autodetect if a node is CAX or not and run the corresponding logic. This would allow to easily have hybrid cluster if someone wants that functionality.
> 
> I'll try to implement it today, but PRs are always appreciated just like this one (:

As it is cheaper and would double the RAM, maybe even make arm the new default for controlplane nodes?

aleksasiriski:
> As it is cheaper and would double the RAM, maybe even make arm the new default for controlplane nodes?

Interesting idea, if that works in a hybrid setup I would be all for it. @mysticaltech thoughts?

aleksasiriski:
I've made another [PR](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/723) that needs thorough testing.

mysticaltech:
Thanks @Kakadus, this is now supplanted by #723 but thanks a lot anyways, PRs are always welcome! 🙏

aleksasiriski:
> As it is cheaper and would double the RAM, maybe even make arm the new default for controlplane nodes?

We could do this when arm servers become available in more locations, as they are currently only in FSN it would make Super HA (control planes in different locations) impossible.

PR #735, opened 2023-04-18
Added k3s_exec_server_args variable.

valkenburg-prevue-ch:
This adds k3s_exec_server_args, which lets you pass arguments to `k3s server` like so:
```
k3s_exec_server_args = "--kube-apiserver-arg enable-admission-plugins=PodTolerationRestriction,PodNodeSelector"
```

See the updated README with and example of a currently very relevant use-case with mixed arm64/amd64 clusters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -574,6 +574,41 @@ For more details, see [Longhorn's documentation](https://longhorn.io/docs/1.4.0/
 
 </details>
 
+<details>
+<summary>Default nodeSelectors let you assign namespaces to arm64/amd64 nodes</summary>
+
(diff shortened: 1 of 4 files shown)


aleksasiriski:
If you have tested this, it can be merged.

valkenburg-prevue-ch:
yes, I verified with and without  (i.e. using default) values for k3s_exec_server_args, this works alright.

mysticaltech (approved):

mysticaltech (approved):
Awesome!

PR #766, opened 2023-04-26
Update hcloud-microos-snapshots.pkr.hcl

wejdross:
replacing default image to working one

### DIFF
--- a/packer-template/hcloud-microos-snapshots.pkr.hcl
+++ b/packer-template/hcloud-microos-snapshots.pkr.hcl
@@ -12,7 +12,7 @@ variable "hcloud_token" {
 # You can find a working mirror at https://download.opensuse.org/tumbleweed/appliances/openSUSE-MicroOS.x86_64-OpenStack-Cloud.qcow2.mirrorlist
 variable "opensuse_microos_x86_mirror_link" {
   type    = string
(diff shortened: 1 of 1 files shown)


mysticaltech:
@wejdross That really depends on your location, on our end, testing showed that the mirror we have set by default works better. Experience showed us that mirror auto-selection is not as robust as it should be, but good that it worked for you. Where are you located more or less?

mysticaltech:
@kube-hetzner/core Maybe we should try giving the mirror auto-selection a try again? What do you think?

otavio (approved):
Fixed my error too.

wejdross:
> @wejdross That really depends on your location, on our end, testing showed that the mirror we have set by default works better. Experience showed us that mirror auto-selection is not as robust as it should be, but good that it worked for you. Where are you located more or less?

Its central Switzerland

aleksasiriski (changes requested):
Please change to `https://download.opensuse.org/ports/aarch64/tumbleweed/appliances/openSUSE-MicroOS.aarch64-OpenStack-Cloud.qcow2` for the ARM mirror link as well. After that, it can be merged.

mysticaltech:
Ok, thanks for the info folks, and for confirming this helped you too @octavio. I think there a good chance OpenSUSE on their side are also improving the mirror autoselect, I would agree with this.

aleksasiriski (approved):

aleksasiriski:
@mysticaltech Should we merge to staging, or directly to master?

mysticaltech:
Thanks @wejdross!! And @aleksasiriski  🚀

mysticaltech:
Just released in v2.1.3 !

PR #774, opened 2023-04-29
Introduce ssh_max_auth_tries variable

pschiffe:
The default value of 2 may be too strict if there's many SSH keys in the ssh-agent. Some parts of this project don't support specific key selection in the ssh-agent, such as `data "remote_file" "kubeconfig"`


mysticaltech (approved):
Thanks @pschiffe, glad you came through with this! Really appreciate it, it will help others!  🙏

mysticaltech:
Just released in v2.1.4 !!

PR #778, opened 2023-05-01
added ssh_connection_timeout to resources

dan1el-k:
# Release notes
This PR introduces a new variable called `ssh_connection_timeout`, which is set to `10m` (10 minutes) as default.
This gives the local ssh client used by terraform more time before aborting. The initial default of terraform is 5 minutes (see: https://developer.hashicorp.com/terraform/language/resources/provisioners/connection)

# Additional information
In some crowded regions and under some circumstances the full network-wise availability of the nodes might be delayed after a restart, which causes the installation to fail. 

In my specific case, I wasn't able to install a simple cluster (3 control-plane nodes, 2 workers) spanned over Falkenstein, Nuernberg, Helsinki within about 50! attempts. After investigating and switching  (from WSL2 to native Linux) as well as tweaking (various ssh connection and key options) of my local infrastructurel, I figured out that some of the nodes were usually just fully available via ssh after 7-8 minutes. Which nodes in which region was always different based on time and day (so not reproducable).

Increasing the ssh timeout to 20m using this new variable finally eased the situation for me and let me install a cluster in almost one shot.

# Linked issues

might cover: https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/discussions/526

### DIFF
--- a/kustomization_user.tf
+++ b/kustomization_user.tf
@@ -11,9 +11,13 @@ resource "null_resource" "kustomization_user" {
     agent_identity = local.ssh_agent_identity
     host           = module.control_planes[keys(module.control_planes)[0]].ipv4_address
     port           = var.ssh_port
+    timeout        = var.ssh_connection_timeout
   }
 
   provisioner "remote-exec" {
+    connection {
+      timeout = var.ssh_connection_timeout
+    }
     inline = [
       "echo 'Create kustomize dir'",
       "mkdir -p /var/user_kustomize"
@@ -21,16 +25,25 @@ resource "null_resource" "kustomization_user" {
   }
 
   provisioner "file" {
+    connection {
+      timeout = var.ssh_connection_timeout
+    }
     source      = "extra-manifests/"
     destination = "/var/user_kustomize"
   }
 
   provisioner "file" {
+    connection {
+      timeout = var.ssh_connection_timeout
+    }
     content     = templatefile("extra-manifests/kustomization.yaml.tpl", var.extra_kustomize_parameters)
     destination = "/var/user_kustomize/kustomization.yaml"
   }
 
   provisioner "remote-exec" {
+    connection {
+      timeout = var.ssh_connection_timeout
+    }
     inline = [
       "rm /var/user_kustomize/kustomization.yaml.tpl",
       "kubectl apply -k /var/user_kustomize/"
--- a/agents.tf
+++ b/agents.tf
@@ -15,6 +15,7 @@ module "agents" {
   ssh_public_key               = var.ssh_public_key
   ssh_private_key              = var.ssh_private_key
   ssh_additional_public_keys   = length(var.ssh_hcloud_key_label) > 0 ? concat(var.ssh_additional_public_keys, data.hcloud_ssh_keys.keys_by_selector[0].ssh_keys.*.public_key) : var.ssh_additional_public_keys
+  ssh_connection_timeout       = var.ssh_connection_timeout
   firewall_ids                 = [hcloud_firewall.k3s.id]
   placement_group_id           = var.placement_group_disable ? null : hcloud_placement_group.agent[floor(index(keys(local.agent_nodes), each.key) / 10)].id
   location                     = each.value.location
@@ -52,6 +53,7 @@ resource "null_resource" "agents" {
     agent_identity = local.ssh_agent_identity
     host           = module.agents[each.key].ipv4_address
     port           = var.ssh_port
+    timeout        = var.ssh_connection_timeout
   }
 
   # Generating k3s agent config file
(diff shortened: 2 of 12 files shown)


kustomization_user.tf
```diff
@@ -11,26 +11,39 @@ resource "null_resource" "kustomization_user" {
     agent_identity = local.ssh_agent_identity
     host           = module.control_planes[keys(module.control_planes)[0]].ipv4_address
     port           = var.ssh_port
+    timeout        = var.ssh_connection_timeout
   }
 
   provisioner "remote-exec" {
+    connection {
+      timeout = var.ssh_connection_timeout
```
mysticaltech, line 19: @dan1el-k This is redundant, I believe it's enough to set it at the ressource level like for the other params, same for the file provisioner.

mysticaltech (changes requested):
Looking good, just some redundant stuff that can be removed IMHO.

kustomization_user.tf
mysticaltech: Ok, thanks for the clarification, appreciate it!

mysticaltech:
@kube-hetzner/core Would love your input on this one. @dan1el-k addresses an important issue that can happen from time to time, especially for automated deployments.

mysticaltech (dismissed):

mysticaltech:
@dan1el-k Something worth exploring is updating to the latest version of terraform and trying again without the redundancy. Please upgrade to the latest and greatest and try again. Because I just asked GPT-4 and it said that provisioners should inherit from the ressource level (which is as it should be, as it's the case for other SSH settings), it also said that it could vary from terraform version to terraform version. So it would really be worth it to try upgrading to the latest stable.

valkenburg-prevue-ch:
Fwiw, I run a dev machine inside hetzner cloud, from which I run all terraform commands, because my personal internet connection is too unstable. With that setup, I have not yet run into the issue described here.

mysticaltech:
> Fwiw, I run a dev machine inside hetzner cloud, from which I run all terraform commands, because my personal internet connection is too unstable. With that setup, I have not yet run into the issue described here.

Very interesting, thanks for sharing.

mysticaltech:
@dan1el-k Did you try again with the info I shared above?

dan1el-k:
So, I spent some more days debugging the behaviour.

I updated to `Terraform v1.4.6`.
I can confirm after several tests, that the **timeout is NOT inherited from the resources to the provisioners**. 

## Packer `.hcl` files
I also needed to add the timeout to the latest extensions on how the snapshots are now created using Packer. Before that, I wasn't able to create a snapshot in 10 attempts. Defining the `ssh_timeout` on `source` and tried another 10 times - same result. Adding the ssh_timeout to the provisioners, suddenly it finished the first time (after 45 min 😃 . I think the 5 minutes statement from the main readme is a bit too optimistic for a full clusters. I took me min. 4-5 hours pure runtime in best case to have a cluster running 😆 )
grafik

## Null resources 
Also here I got timeouts in various `null resources` but 95% of the times it failed at the `init.tf` "null resources kustomization" and the `kubeconfig.tf`. 

### Before adding the ssh_timeout to the "provisioners"
grafik

### After adding the ssh_timeout to the "provisioners"

grafik
or
grafik



But here I even observed many failures after adding timeouts to any resources. I think its related to the the fact that the file, local and remote provisioners are used inflationary, but shouldnt' (https://developer.hashicorp.com/terraform/language/resources/provisioners/file)
> Important: Use provisioners as a last resort. There are better alternatives for most situations. Refer to [Declaring Provisioners](https://developer.hashicorp.com/terraform/language/resources/provisioners/syntax) for more details.

--> I will open an discussion to talk about talk about this and think about splitting the steps or any alternative or more mature solutions for the job (like ansible)

## ssh connectivity to Hetzner

In general I would argue that the ssh timeout issue is not caused client wise. I tried different devices (Windows, Linux, Mac) with different internet connections (landline and mobile) even using a VPN directly connected to Frankfurt germany. All showed the same behaviour. It seems Hetzner is doing some DDoS protection or healtch checking of the backend service from a managed additionaly firewall in their network stack or whatever, and therefore causing delays after reboots but also drops of running connections...

mysticaltech:
@dan1el-k Super weird, we never saw it before. Can you please try again while using cloudflare https://one.one.one.one/?

mysticaltech:
@dan1el-k This seems like a very specific use case, as we have not seen any other complaints on the matter. Hence my hesitation to merge. @kube-hetzner/core Please don't hesitate to share you thoughts on this one!

aleksasiriski:
If we get more requests for this feature we can merge, otherwise I think this is user related and needs fixing on their side (internet connection).

mysticaltech:
@aleksasiriski Exactly!

PR #780, opened 2023-05-02
Fix/optional skip GitHub calico

valkenburg-prevue-ch:
data "github_release" "..." can be skipped for those resources which are disabled by the user.


mysticaltech (approved):
Yes Sir!

Issue #772, opened 2023-04-29
Cluster wide IPv6?


riemervdzee:
### Description

I'm currently debugging an issue on my cluster where my workloads 'sometimes' cannot access an external website/endpoint. I narrowed this down to ipv4/6 issues; when coredns returns an ipv4 address my services work, if it returns an ipv6 address then opening a socket fails (errors like `Network unreachable`, `ENETUNREACH`). 

When I log into the containers I see no default route for ipv6 and cannot ping via ipv6, the same happens when I log into a controlplane or agent node.

I've tested this with both Cilium and Flannel as CNI's. Is it possible to enable ipv6 somewhere, or configure coredns to use ipv4 only?

### Kube.tf file

```terraform
module "kube-hetzner" {
  providers = {
    hcloud = hcloud
  }
  hcloud_token = var.hcloud_token

  source = "kube-hetzner/kube-hetzner/hcloud"
  version = "2.1.1"

  ssh_port = 2222
  ssh_public_key = file("...")
  ssh_private_key = file("....")

  network_region = "eu-central" # change to `us-east` if location is ash

  control_plane_nodepools = [
    {
      name        = "control-plane-fsn1",
      server_type = "cpx21",
      location    = "nbg1",
      labels      = [],
      taints      = [],
      count       = 1
    },
  ]
  agent_nodepools = [
    {
      name        = "agent-lg-nbg1",
      server_type = "cpx21",
      location    = "nbg1",
      labels      = [],
      taints      = [],
      count       = 0
    },
  ]

  cluster_name = "k3s-test"
  restrict_outbound_traffic = false
}

provider "hcloud" {
  token = var.hcloud_token
}

terraform {
  required_version = ">= 1.3.3"
  required_providers {
    hcloud = {
      source  = "hetznercloud/hcloud"
      version = ">= 1.35.2"
    }
  }
}

output "kubeconfig" {
  value     = module.kube-hetzner.kubeconfig
  sensitive = true
}

variable "hcloud_token" {
  sensitive = true
  default   = ""
}
```

### Screenshots

Screenshot when logged in a container

<img width="433" alt="Screenshot 2023-04-29 at 12 49 58" src="https://user-images.githubusercontent.com/510299/235298806-675bb2d9-e5db-4e9f-ab38-3fbcdec1f8c8.png">

Screenshot when logged in from the controlplane node. The same happens on agents
[...]

PR #792, opened 2023-05-07
Properly guide cloud-init to use NetworkManager

riemervdzee:
# Summary
Cloud-init does not generate the connections config correctly for NetworkManager as stated in issue https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/772. By creating the necessary file, cloud-init will properly configure NetworkManager and resolve the IPv6 issue on our Nodes. IPv6 for Pods is step two.

# Detailed info

Cloud-init correctly reads the configuration given by Hetzner, but incorrectly detects systemd-network as the network service instead of the actual NetworkManager. This is due to a missing `/etc/NetworkManager/NetworkManager.conf` file, which is required according to cloud-init's source code: https://github.com/canonical/cloud-init/blob/main/cloudinit/net/network_manager.py#L381

By creating this file before generating the network config, cloud-init is correctly guided to use NetworkManager and a valid file is written to `/etc/NetworkManager/system-connections/cloud-init-eth0.nmconnection`. The NetworkManager.conf file creation is performed in the packer file because the network configuration is generated early in the process, making it too late to create the file in `cloudinit_write_files_common` for example.

### DIFF
--- a/packer-template/hcloud-microos-snapshots.pkr.hcl
+++ b/packer-template/hcloud-microos-snapshots.pkr.hcl
@@ -61,6 +61,11 @@ locals {
     rm -rf /etc/ssh/ssh_host_*
     sleep 1 && udevadm settle
   EOT
+
+  cloud_init_network = <<-EOT
+    echo 'Make sure to use NetworkManager'
+    touch /etc/NetworkManager/NetworkManager.conf
+  EOT
 }
 
 # Source for the MicroOS x86 snapshot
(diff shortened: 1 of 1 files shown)


valkenburg-prevue-ch (approved):
Looks good to me from a code perspective, haven't tested.

mysticaltech (approved):
Looking stunning! Will run a quick basic test and merge later tonight.

mysticaltech:
⭐ FYI people, here's how to upgrade existing clusters https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/discussions/794 ⭐

PR #803, opened 2023-05-12
[Feature] Add ability to turn on SWAP for the ControlPlane and Agent nodes

karbowiak:
It's a pretty straight forward change, by default it's business as usual - no swap.

But pr. control plane/agent it can be enabled.

By default if it's just turned on, it'll create a swap file in `/var/swapfile` that's the size of the RAM, so if 3GB memory, then it's a 3GB SWAP file.

It can however be overruled by using `swap_size`, for example `swap_size = "8G"` for 8GB of SWAP.

Examples:

```
  control_plane_nodepools = [
    {
      name        = "master",
      server_type = "cax11",
      location    = "fsn1",
      labels      = [],
      taints      = [],
      count       = 3,
      swap        = true, # Enable swap on the agent nodes. (default is false)
      swap_size   = 0     # Set this to 0 to automatically set it to the same amount of RAM on the node, example: swap_size = 2G, swap_size = 8G etc. (Important to remember the suffix)
    }
  ]

  agent_nodepools = [
    {
      name        = "worker",
      server_type = "cax11",
      location    = "fsn1",
      labels      = [],
      taints      = [],
      count       = 1,
      swap        = true, # Enable swap on the agent nodes. (default is false)
      swap_size   = "8G"     # Set this to 0 to automatically set it to the same amount of RAM on the node, example: swap_size = 2G, swap_size = 8G etc. (Important to remember the suffix)
    }
  ]
  ```
  
  I updated the docs as well.

### CODE REVIEW

**Review by aleksasiriski — DISMISSED** [review]

**Review by mysticaltech — DISMISSED** [review]
@karbowiak Looking good!

@ifeulner Please have a look if you can.

**Comment by mysticaltech** [comment]
@karbowiak Now it's a lot more elegant! Well done.

**Comment by aleksasiriski** [comment]
Is this a cluster-wide option now? If so, I believe I liked the previous approach a lot more.. since this means that both 16gb ram and 2gb ram machines would all use the same swap size.. Also, what if someone wanted to have one node pool of lower performance with swap and correct labels to use node selector on?

**Comment by karbowiak** [comment]
@aleksasiriski i wholeheartedly agree with you, sadly there isn't really a way to do it both ways 😅 

It's one way or another - the current cloud-init way is only a revert away as well, so not a big deal to go back.

**Comment by karbowiak** [comment]
So, should i revert the latest change, and make it pr. node again? 😅

**Comment by ifeulner** [comment]
> I would lean towards the previous solution and add more precautions to it. For example, if swap is enabled we could add a label to the node pool automatically `node.kubernetes.io/server-swap=enabled` and for zram `node.kubernetes.io/server-zram=enabled`. Then users would be advised in the README to carefully read about implications of using swap, and to add node selectors with the aforementioned labels to their services, whether it be affinity or antiaffinity.
> 
> @mysticaltech @ifeulner What's your opinion on this proposal?

Hi @aleksasiriski I don't see why this could not work with cloud-init? 
Adjusting Swap / zram is an OS-related topic, so it should be in the host module. But of course it could / should be adjustable per node-group then, as swap also often depends on disk / RAM configuration of the VM.

> 
> > Ingo: I am quite a big fan of immutable infrastructure. Essential things such as swap should not change after a node has been created.
> 
> I don't think this poses a real problem even in the previous implementation, just like how k3s is only setup once, swap and zram would be in the same boat so no need to worry about that.

See my comment above.

**Comment by aleksasiriski** [comment]
@karbowiak The changes for initializing swap are correct now, but should also be adjustable for each node group.

**Comment by mysticaltech** [comment]
@karbowiak I really like @aleksasiriski's proposal below, please implement if you can. 

> I would lean towards the previous solution and add more precautions to it. For example, if swap is enabled we could add a label to the node pool automatically `node.kubernetes.io/server-swap=enabled` and for zram `node.kubernetes.io/server-zram=enabled`. Then users would be advised in the README to carefully read about implications of using swap, and to add node selectors with the aforementioned labels to their services, whether it be affinity or antiaffinity.

And as both @ifeulner and @aleksasiriski reminded, the setting of the swap size should be on a per nodepool level. Now, this can easily be done automatically via cloud-init runcmd. 

Here's some inspiration from our good friend GPT4:

```bash
#!/bin/bash
# Calculate Swap space
RAM=`free | grep Mem | awk '{print $2}'`
Swap=$((RAM * 2))
SwapInGB=$((Swap / 1024 / 1024))

# Disable all swap processes
sudo swapoff -a

# Create a swap file with desired size
sudo fallocate -l ${SwapInGB}G /swapfile

# Set the correct permissions
sudo chmod 600 /swapfile

# Make it a swap file
sudo mkswap /swapfile

# Enable the swap file
sudo swapon /swapfile

# Make the change permanent
echo '/swapfile swap swap defaults 0 0' | sudo tee -a /etc/fstab
```

mysticaltech:
@karbowiak Did you get a chance to try the above? Would love to push this with the next release.

karbowiak:
> @karbowiak Did you get a chance to try the above? Would love to push this with the next release.

yes i did not 😅 had a busy week and all that, BUT i will look at it right now

mysticaltech:
@karbowiak Go, go, go!  Let's launch faster than SpaceX lol 🚀🚀🚀

modules/host/variables.tf
```diff
@@ -118,3 +118,15 @@ variable "cloudinit_runcmd_common" {
...
+  type        = bool
+  default     = false
+  description = "Enable swap on the nodes."
+}
+
+variable "swap_size" {
+  type        = string
+  default     = "2G"
+  description = "Size of the swap file."
+}
```
ifeulner, line 132: I don't really understand why both is needed? Why not enable swap if swap_size >= 0 ? 
But maybe it's more clearer this way... 

Also I'd prefer a separation between control nodes and agent nodes... in a production environment I would never enable swap on the control nodes... but maybe (really only maybe) on the agent nodes. 
For a demo cluster with just control nodes it might make sense, though.

ifeulner (changes requested):
I'd like to see this flag separated for control and agent nodes.

Silvest89:
Finalized https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/1019

PR #804, opened 2023-05-12
[Feature] Add ability to turn on ZRAM for ControlPlane and Agent nodes

karbowiak:
Similar to https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/803 this is a fairly straight forward change.
If it's not defined (or defined as false) it doesn't enable - so there should be no problems with backwards compatibility.

(Only thing might be the /tmp/zram.service that gets created - not entirely sure how to prevent that 😅)

It does not have a default size, so if someone does not provide a size, it'll not enable during provisioning.

Examples for `kube.tf`

```
  control_plane_nodepools = [
    {
      name        = "master",
      server_type = "cax11",
      location    = "fsn1",
      labels      = [],
      taints      = [],
      count       = 3,
      zram        = true # It's only enabled, but doesn't have a size - so it won't actually be enabled
    }
  ]

  agent_nodepools = [
    {
      name        = "worker",
      server_type = "cax11",
      location    = "fsn1",
      labels      = [],
      taints      = [],
      count       = 1,
      zram        = true # This is enabled
      zram_size   = "1024M" # And it has a size defined, so it will get setup and started
    }
  ]
  ```
  
  Docs were also updated 😅

### CODE REVIEW

**Review by mysticaltech — APPROVED** [review]
Good job!

**Review by otavio — APPROVED** [review]

**Review by ifeulner — APPROVED** [review]
I am still not a big fan of this and would never use it on a production cluster, but maybe it helps for demo environments. 
Thanks a lot anyways @karbowiak for providing this implementation!

File: kube.tf.example [hunk 1]
```diff
@@ -98,7 +98,14 @@ module "kube-hetzner" {
       location    = "fsn1",
       labels      = [],
       taints      = [],
-      count       = 1
+      count       = 1,
+
+      # Keep in mind, both zram _AND_ zram_size has to be defined, otherwise zram will not be enabled
```
aleksasiriski, line 103: Why even have zram as an option? If zram_size > 0 it sholud be enabled, otherwise disabled.

aleksasiriski (changes requested):
Also, another approach would be to only keep `zram` as an option and always set the size to the default % of system memory as described [here](https://forums.fedoraforum.org/showthread.php?325086-Is-there-any-way-to-set-more-percentage-of-memory-for-ZRAM)

mysticaltech:
@Silvest89 What's your opnion on supporting zRam? Do you think it would be valuable?

Silvest89:
> @Silvest89 What's your opnion on supporting zRam? Do you think it would be valuable?

@mysticaltech 
I've heard it has benefits especially in low ram environments.
I would do it in the cloudinit instead of using a systemd service.

mysticaltech:
@Silvest89 That would be awesome! Thanks.

mysticaltech:
Superseded by #1060. Thanks for starting this @karbowiak.

PR #805, opened 2023-05-13
Add hcloud_network and hcloud_ssh_key ids to output so other resources outside of the module can refer to them

schmidp:
In my setup I also create another VMs that should use the same key as created by the module.

I also use the same network.

Using data resources and depends_on leads to recreation of resources every time a value in the kube-hetzner module changes.

### DIFF
--- a/output.tf
+++ b/output.tf
@@ -3,6 +3,16 @@ output "cluster_name" {
   description = "Shared suffix for all resources belonging to this cluster."
 }
 
+output "network_id" {
+  value       = hcloud_network.k3s.id
+  description = "The ID of the HCloud network."
+}
+
+output "ssh_key_id" {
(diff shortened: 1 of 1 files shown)


aleksasiriski (approved):

mysticaltech (approved):

mysticaltech:
@schmidp Thanks for this PR. So these outputs fix your issues, correct?

schmidp:
yes it does. if you use the output, you don't have to use a data source that depends on module which flags resources that depend on the data source to recreate if the module changes.

I just needed those two values, but the module should output more in the future for anything others would like to reference.

mysticaltech:
@schmidp Thanks, perfect!

PR #809, opened 2023-05-17
A lot of QoL changes to Traefik - nothing breaking

aleksasiriski:
1. Added `traefik_pod_disruption_budget` which is enabled by default
2. Added `traefik_resource_limits` which are enabled by default
3. Added `traefik_additional_ports` which allow users to add custom ports to Traefik, for example port 22 for ssh is added with these rules:
```terraform
traefik_additional_ports = [
    {
      name        = "ssh",
      port        = 22,
      exposedPort = 22
    }
  ]
```

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -502,6 +502,12 @@ resources:
     cpu: "300m"
     memory: "150Mi"
 %{endif~}
+%{if var.traefik_autoscaling~}
+autoscaling:
+  enabled: true
+  minReplicas: ${local.ingress_replica_count}
+  maxReplicas: ${local.ingress_max_replica_count}
+%{endif~}
   EOT
 
   rancher_values = var.rancher_values != "" ? var.rancher_values : <<EOT
--- a/locals.tf
+++ b/locals.tf
@@ -461,6 +461,7 @@ ports:
       trustedIPs:
         - 127.0.0.1/32
         - 10.0.0.0/8
+%{if var.traefik_additional_ports != ""~}
 %{for option in var.traefik_additional_ports~}
   ${option}:
     proxyProtocol:
@@ -472,6 +473,7 @@ ports:
         - 127.0.0.1/32
         - 10.0.0.0/8
 %{endfor~}
+%{endif~}
 %{endif~}
 podDisruptionBudget:
 %{if var.traefik_pod_disruption_budget~}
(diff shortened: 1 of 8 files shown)


aleksasiriski:
I'm testing this heavily now, will say when it's ready to be merged.

locals.tf
```diff
@@ -462,11 +462,45 @@ ports:
...
+%{endif~}
 additionalArguments:
+  - "--entrypoints.tcp=true"
+%{if var.traefik_additional_options != ""~}
 %{for option in var.traefik_additional_options~}
-- "${option}"
+  - "${option}"
 %{endfor~}
+%{endif~}
+%{if var.traefik_resource_limits~}
```
mysticaltech, line 496: @aleksasiriski Just out of curiosity, what's the reasoning behind the limits?

mysticaltech: What I am worried about is that resource limits by default will limit the amount of requests that can be processed by the ingress controller, am I wrong?

mysticaltech:
Apart from the above question, it looks great!

locals.tf
aleksasiriski: Those were the default resource limits commented out in Traefik Helm chart.

If a deployment doesn't have any resource limits, it get killed first if another deployment with resource limits needs those resources.

Best solution for High Traffic would probably HPA enabled as well, so it can be upscaled when needed and downscaled otherwise.

aleksasiriski:
I still haven't fully tested `traefik_additional_ports`, will say when I'm done.

locals.tf
mysticaltech: @aleksasiriski Ok good info, thanks. But what do you mean by HPA?

aleksasiriski: Horizontal Pod Autoscaler, increases number of replicas when the load increases and decreases them when load decreases. It's included in Traefik Helm chart

aleksasiriski:
Tested deploying a cluster (my production cluster to be precise), and it's working as expected. Traefik HPA scales up when needed, and scales down otherwise. Everything deploys successfully and I can see in Hetzner Load Balancer webui that the port `22` for ssh is added.

Will test later if a gitea instance with that port works, after that it can be merged.

locals.tf
mysticaltech: Ok I see, beautiful!

mysticaltech (approved):
Awesome!

mysticaltech:
@aleksasiriski Is this good to fly?

aleksasiriski:
Haven't had time to test gitea, but this PR doesn't break old behaviour.

mysticaltech:
Ok, perfect!

mysticaltech (approved):

PR #822, opened 2023-05-24
Update generate-docs.yaml - Bypass branch protection

aleksasiriski:
Added ssh key for bypassing branch protection rules for Github Actions

This is the reference: https://stackoverflow.com/a/76135647


mysticaltech (approved):
Awesome! I think that's enough.

PR #841, opened 2023-06-13
Fix k3s docs link about backup and restore in kube.tf.example

RelativeSure:


mysticaltech (approved):
@RelativeSure Thanks for that!

PR #844, opened 2023-06-15
Drop default ssh firewall rule when custom ssh port is used

MaximilianMeister:
It is not necessary to open the standard port 22 when using a custom port

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -157,15 +157,7 @@ locals {
       source_ips  = ["0.0.0.0/0", "::/0"]
     },
 
-    # Allow all traffic to the ssh ports
-    {
-      description = "Allow Incoming SSH Traffic"
-      direction   = "in"
-      protocol    = "tcp"
-      port        = "22"
-      source_ips  = ["0.0.0.0/0", "::/0"]
-    }
-    ], var.ssh_port == 22 ? [] : [
+    # Allow all traffic to the ssh port
     {
       description = "Allow Incoming SSH Traffic"
       direction   = "in"


aleksasiriski:
Have you tested building a complete cluster from scratch with this patch? I forgot if nodes are firstly added to the firewall and only afterwards the port is changed from the default, which would make this patch break the cluster setup or if it was the other way around, which means it's perfectly safe to merge this patch.

MaximilianMeister:
@aleksasiriski Yeah, I tested it with a 3 CP + 2 Worker node cluster and checked the firewall by looking into hetzner UI and also trying to ssh into a machine and it seems to do the trick

locals.tf
```diff
@@ -157,15 +157,16 @@ locals {
       source_ips  = ["0.0.0.0/0", "::/0"]
     },
 
-    # Allow all traffic to the ssh ports
+    # Allow all traffic to the ssh port
```
aleksasiriski, line 160: Please combine this into a single rule, there's no need for checking if the port is 22, just set it to `var.ssh_port` since it has the default value of 22 if nothing is set in kube.tf.

aleksasiriski (changes requested):

aleksasiriski (approved):

mysticaltech:
@MaximilianMeister Indeed that is legacy logic left over from the days when we used to enter rescue mode to install the OS. Thanks for the find!

mysticaltech (approved):

MaximilianMeister:
> Indeed that is legacy logic left over from the days when we used to enter rescue mode to install the OS. Thanks for the find!

Ah good to know. Thanks!

PR #827, opened 2023-05-26
Improve kustomization support

nikolaszimmermann:
Currently only one configurable 'extra-manifests/kustomization.yaml.tpl' template is available. Extend that to all file named *.yaml.tpl and create individual terraform resources for al of them, individually. This way one can precisely taint and rebuild what is necessary.

Allow to run custom deploy commands after the initial kubectl apply -f.

TODO: Improve change log, add example to docs

PR #848, opened 2023-06-18
Fix error with extra-manifests

kayiplar:
Using the `fileexists` function on the `extra-manifests` directory throws the following error when it exists: 

```

╷
│ Error: Error in function call
│ 
│   on .terraform/modules/k8s.kube-hetzner/kustomization_user.tf line 2, in locals:
│    2:   extra_manifests_exists       = fileexists("extra-manifests")
│     ├────────────────
│     │ while calling fileexists(path)
│ 
│ Call to function "fileexists" failed: "extra-manifests" is a directory, not
│ a file.
```

The function can be used only on a file, not a directory. 

A safer solution would be to use the `try` function if there is some error with `fileset`.

### DIFF
--- a/kustomization_user.tf
+++ b/kustomization_user.tf
@@ -1,6 +1,5 @@
 locals {
-  extra_manifests_exists       = fileexists("extra-manifests")
-  user_kustomization_templates = local.extra_manifests_exists ? fileset("extra-manifests", "*.yaml.tpl") : toset([])
+  user_kustomization_templates = try(fileset("extra-manifests", "*.yaml.tpl"), [])
 }
 
 resource "null_resource" "kustomization_user_setup" {


mysticaltech:
@kayiplar I hadn't tried that one, thank you! See the context in which it was introduces at the end of https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/827.

mysticaltech (approved):
I'm hesitating about the toset() being removed, but I assumed you have tried it, so merging, thanks!

mysticaltech:
@kayiplar Finally added toset back, but it's the same actually. It's been released in v2.2.3 🚀

@nikolaszimmermann FYI this fix, my previous "fix" wasn't ideal as I had not tested it fully. Now everything should be good.

nikolaszimmermann:
Thanks a lot for taking care!Am 20.06.2023 um 04:09 schrieb Karim Naufal ***@***.***>:﻿
@kayiplar Finally added toset back, but it's the same actually. It's been released in v2.2.3 🚀
@nikolaszimmermann FYI this fix, my previous "fix" wasn't ideal as I had not tested it fully. Now everything should be good.

Issue #726, opened 2023-04-15
[Feature Request]: Labels/Taints for autoscaler_nodepools


Fabb111:
### Description

Since autoscaler respects labels and taints for scheduling pods, it would be great to be able to set labels and taints on the autoscaler node pools similar to the normal node pools. If that's already possible somehow, it would be great to document it since I couldn't find anything related to it.

mysticaltech:
Yep, it's possible! Will do ASAP, otherwise PR welcome, it needs to follow the same from as normal agents to propagate the values. So not a lot to be done, but it does need testing!

ifeulner:
We are currently  already applying the default label / taints... maybe a set dedicated to the autoscaling nodes should be considered.
(see code [here](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/488e107b599d538a732b8d7134b85197c967de38/autoscaler-agents.tf#L82))

mysticaltech:
@ifeulner Yes, definitely, it's actually quite straightforward, just need to spend some time on it.

sambonbonne:
I'm trying to implement this feature but it does not seem that straightforward: the cluster autoscaler can only take one `HCLOUD_CLOUD_INIT` variable, which contains the taints and labels we want to set.

The solution I'm trying to implement is to split `templates/autoscaler.yaml.tpl` to make a dedicated template for the Deployment and template it with different names and labels (`cluster-autoscaler-${autoscaler_pool_name}`) to be able to give each autoscaler a different `HCLOUD_CLOUD_INIT` variable.

It means it will start one autoscaler deployment per autoscaled pool, which may not be desirable to everyone.

Should I continue this way or did I miss something that would make the configuration simpler? If I continue, be assured my pull request will contain some documentation update in order to warn users about this behavior.

valkenburg-prevue-ch:
I don't know of a simpler solution, and I would be impressed and happy if your plan is going to work!

-------- Oorspronkelijk bericht --------
Op 14 jun. 2023 16:36, schreef Sam D :

sambonbonne:
I started working on it today but I currently have an issue: the autoscaler does not trigger the scale up of the first node (even if I set `min_nodes = 1`).

I will continue to work on it next week.

If anyone want to try it, you can change the source of the `kube-hetzner` module in your `kube.tf` to:

```hcl
[...]

PR #849, opened 2023-06-19
[Feature] Labels and taints for autoscaler nodepools

sambonbonne:
This pull requests add the feature asked in #726: the ability to add taints and labels for autoscaler nodepools.

Plus, it adds the ability to give custom extra arguments to the cluster autoscaler.

## Modifications 

* added `autoscaler_labels` and `autoscaler_taints` optional variables, which are each a `list(string)` like `agent_nodepools[*].labels` and `agent_nodepools[*].taints` variables
* added the `autoscaler_labels` and `autoscaler_taints` values in the `k3s_config` file of the cloudinit coniguration
* added `cluster_autoscaler_extra_args` optional variable, which is a `list(string)` too
* added the `cluster_autoscaler_extra_args` values in the autoscaler template (`templates/autoscaler.yaml.tpl`
* added these three new variables with description and examples in `kube.tf.example`
* added these three new variables in the Terraform documentation (`docs/terraform.md`)

## Tests done

All those tests are working:

* not using any autoscaler node pools
* using autoscaler node pools as before (without taints, labels nor extra args)
* using autoscaler node pools with taints, labels or both (but without extra args)
* using autoscaler node pools with extra args (but without labels nor taints)
* using autoscaler node pools with labels, taints and extra args


aleksasiriski (approved):
Looks good to me

mysticaltech (approved):
@sambonbonne Superb PR. That's a valuable addition. Thanks for that 🙏

Issue #852, opened 2023-06-21
[Bug]: `extra_kustomize_deployment_commands` seems to be required


FelixSab:
### Description

I've encountered a problem when using the kube-hetzner module with Terraform. Even though `extra_kustomize_deployment_commands` is stated to be an optional variable in the comments, I've received an error when I don't define it in my `kube.tf` file:

```
│ Error: remote-exec provisioner error
│ 
│   with module.kube-hetzner.null_resource.kustomization_user_deploy[0],
│   on .terraform/modules/kube-hetzner/kustomization_user.tf line 80, in resource "null_resource" "kustomization_user_deploy":
│   80:   provisioner "remote-exec" {
│ 
│ invalid empty string in 'scripts'
```

This error indicates that the variable `extra_kustomize_deployment_commands` is being used even when it's not defined, which leads to an error. I've managed to workaround this issue by defining the variable as follows:

```terraform
# Extra commands to be executed after the `kubectl apply -k` (useful for post-install actions, e.g. wait for CRD, apply additional manifests, etc.).
extra_kustomize_deployment_commands="echo"
```

### Kube.tf file

```terraform
locals {

  hcloud_token = "xxxxxxxxxxx"
}

module "kube-hetzner" {
  providers = {
    hcloud = hcloud
  }
  hcloud_token = var.hcloud_token != "" ? var.hcloud_token : local.hcloud_token


  source = "kube-hetzner/kube-hetzner/hcloud"


  ssh_public_key  = file("~/.ssh/id_ed25519.pub")
  ssh_private_key = file("~/.ssh/id_ed25519")


  network_region = "eu-central"
  control_plane_nodepools = [
    {
      name        = "control-plane-nbg1",
      server_type = "cpx11",
      location    = "nbg1",
      labels      = [],
      taints      = [],
    count = 1 }
  ]

  agent_nodepools = [
    {
      name                 = "internal-services",
      server_type          = "cx21",
      location             = "nbg1",
      labels               = ["internal", "production"],
      taints               = [],
      longhorn_volume_size = 0,
      count                = 1,
      backups              = true
    },
    {
      name                 = "production",
      server_type          = "cx21",
      location             = "nbg1",
      labels               = ["external", "production"],
      taints               = [],
      longhorn_volume_size = 0,
    count = 1 },
    {
      name                 = "database",
      server_type          = "cx31",
      location             = "nbg1",
      labels               = ["database"],
      taints               = [],
[...]

PR #853, opened 2023-06-22
[Bug] add default value to extra_kustomize_deployment_commands

FelixSab:
https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/852


mysticaltech:
@FelixSab Thank you for that. However, please have a look at https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/854, it seems like a more optimal solution. Please do not hesitate to confirm if it works for you! 🙏

PR #854, opened 2023-06-22
[Bug] kustomization_user_deploy and var.extra_kustomize_deployment_commands

kayiplar:
To avoid issues when the variable `var.extra_kustomize_deployment_commands`, we can use the `compact` function to strip elements from the list that are `""` or `null`.

With the current default, the following error occurs while applying:
```
╷
│ Error: remote-exec provisioner error
│ 
│   with module.k8s[0].module.kube-hetzner.null_resource.kustomization_user_deploy[0],
│   on .terraform/modules/k8s.kube-hetzner/kustomization_user.tf line 80, in resource "null_resource" "kustomization_user_deploy":
│   80:   provisioner "remote-exec" {
│ 
│ invalid empty string in 'scripts'
╵
```


mysticaltech:
@kayiplar That is much more elegant indeed, will test and merge ASAP, thanks!

PR #855, opened 2023-06-22
fix longhorn prevent downscaling

ErikFilbert:
when using autoscaler and longhorn at the same time, the longhorn instance manager is deployed on the autoscaler nodes.
When downscaling, these pods cannot be unscheduled.
Longhorn has integrated the "kubernetesClusterAutoscalerEnabled" configuration switch into the Helmchart to solve this problem.

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -378,6 +378,9 @@ spec:
 
   longhorn_values = var.longhorn_values != "" ? var.longhorn_values : <<EOT
 defaultSettings:
+%{if length(var.autoscaler_nodepools) != 0~}
+  kubernetesClusterAutoscalerEnabled: true
+%{endif~}
   defaultDataPath: /var/longhorn
 persistence:
   defaultFsType: ${var.longhorn_fstype}


aleksasiriski:
This feature is still [experimental](https://longhorn.io/docs/1.4.2/high-availability/k8s-cluster-autoscaler/), but I guess people could use NoSchedule on autoscaler nodepools if they wanted to not risk anything? This should be discussed.

aleksasiriski (approved):
It's experimental feature, but the PR is good.

mysticaltech:
@aleksasiriski Our autoscaler now supports labels too, maybe it's doable for longhorn to exclude certain nodes via labels.

mysticaltech (approved):
But yeah, @ErikFilbert PR looks good and if it works for you, it's a good enough validation IMHO. Will merge ASAP.

ErikFilbert:
Hi @aleksasiriski , @mysticaltech ,
do you know which label/taint can be used to prevent longhorn from deploying to autoscaler?
if that works, it might be a better solution than the experimental longhorn configuration.

aleksasiriski:
This should be relevant: https://longhorn.io/docs/1.4.2/advanced-resources/deploy/taint-toleration/

But I don't think it's a "better" solution, just an alternative. I believe we can merge this but explicitly say that it's experimental, and offer taints for autoscales nodepools as an alternative method.

PR #859, opened 2023-06-27
Autocaler logging and kube.tf tweaks

mysticaltech:
Added a bunch of customization for the autoscaler logging, because of https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/discussions/708.

And in the process, noticed that we had multiple missing variables not mentioned in the kube.tf, so added those.


mysticaltech:
@aleksasiriski If you coudl review this it would be great. Thanks!

kube.tf.example
```diff
@@ -535,6 +582,12 @@ module "kube-hetzner" {
   # Don't create the kustomize backup. This can be helpful for automation.
   # create_kustomization = false
 
+  # MicroOS x86 snapshot IDs to be used. Per default empty, the most recent image created using createkh will be used.
```
aleksasiriski, line 585: Change to "MicroOS snapshot IDs", without "x86".

kube.tf.example
```diff
@@ -535,6 +582,12 @@ module "kube-hetzner" {
   # Don't create the kustomize backup. This can be helpful for automation.
   # create_kustomization = false
 
+  # MicroOS x86 snapshot IDs to be used. Per default empty, the most recent image created using createkh will be used. 
+  # We recommend the default, but if you want to use specific IDs you can. 
+  # You can fetch the ids with the hcloud cti by running the "hcloud image list --selector 'microos-snapshot=yes'" command.
```
aleksasiriski, line 587: It should say "hcloud cli", not "cti".

aleksasiriski (approved):
Good additions, if it's been tested it can be merged. It's not a lot of changes so it should work from the looks of it.

kube.tf.example
mysticaltech: Thanks!!

mysticaltech:
I don't know if it's hetzner or something, I only was able to deploy on CAX instances. Will wait some more.

ifeulner (approved):
Hi @mysticaltech looks good, valuable additions! thanks!

PR #875, opened 2023-07-04
feat: added variable for ping address

valkenburg-prevue-ch:
Default is 1.1.1.1, but if you use a proxy, you may want to ping the proxy instead.


mysticaltech (approved):
Awesome!

Issue #850, opened 2023-06-20
[Bug]: Terraform Apply does not update k3s config.yaml


TroyKomodo:
### Description

When changing a value which would normally change the state of the `/etc/rancher/config.yaml` after deploy terraform apply returns successfully but the apply does not actually change the state of the servers.

### Kube.tf file

```terraform
locals {
  cluster_ipv4_cidr = "10.42.0.0/16"
}

module "kube-hetzner" {
  providers = {
    hcloud = hcloud
  }

  hcloud_token = var.hcloud_token

  source = "kube-hetzner/kube-hetzner/hcloud"

  ssh_public_key = file("~/.ssh/id_ed25519.pub")
  ssh_private_key = file("~/.ssh/id_ed25519")

  network_region = "eu-central"

  control_plane_nodepools = [
    {
      name        = "control-plane-fsn1",
      server_type = "cpx11",
      location    = "fsn1",
      labels      = [],
      taints      = [],
      count       = 1
    },
    {
      name        = "control-plane-nbg1",
      server_type = "cpx11",
      location    = "nbg1",
      labels      = [],
      taints      = [],
      count       = 1
    },
    {
      name        = "control-plane-hel1",
      server_type = "cpx11",
      location    = "hel1",
      labels      = [],
      taints      = [],
      count       = 1
    }
  ]

  agent_nodepools = [
    {
      name        = "agent-small",
      server_type = "cpx11",
      location    = "fsn1",
      labels      = [],
      taints      = [],
      count       = 1
    },
  ]

  load_balancer_type     = "lb11"
  load_balancer_location = "fsn1"

  autoscaler_nodepools = [
    {
      name        = "autoscaled-medium"
      server_type = "cpx41"
      location    = "fsn1"
      min_nodes   = 1
      max_nodes   = 5
    }
  ]

  etcd_s3_backup = {
    etcd-s3-endpoint            = "xxx.r2.cloudflarestorage.com"
    etcd-s3-access-key          = "xxx"
    etcd-s3-secret-key          = "xxx"
    etcd-s3-bucket              = "xxx"
  }

  control_planes_custom_config = {
    etcd-snapshot-schedule-cron = "0 */6 * * *"
  }

  use_control_plane_lb = "true"

  cni_plugin = "cilium"

  cluster_ipv4_cidr = local.cluster_ipv4_cidr
  
  cilium_values = <<EOT
ipam:
  mode: "cluster-pool"
  operator:
[...]

PR #881, opened 2023-07-09
Dynamic reconfiguration of k3s on control planes

mraerino:
The idea is to allow for the config provisioner to re-run whenever the config changes and thereby apply it to the control plane node.

This is more of a PoC than anything and also a request for comments.

- How do you imagine dynamic reconfiguration of the control plane to work?
- Should there be a rollback mechanism in case the new config does not work?
- Can we do the nodes one-at-a-time to not cause control plane downtime if the config is faulty?
- Should we do a node restart instead of just restarting the k3s unit?
- How can we work around the fact that applying the config on node creation is different than when it is changed on an existing node?

_Related to https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/850_

### DIFF
--- a/control_planes.tf
+++ b/control_planes.tf
@@ -76,6 +76,72 @@ resource "hcloud_load_balancer_service" "control_plane" {
   listen_port      = "6443"
 }
 
+locals {
+  k3s-config = { for k, v in local.control_plane_nodes : k => merge(
+    {
+      node-name = module.control_planes[k].name
+      server = length(module.control_planes) == 1 ? null : "https://${
+        var.use_control_plane_lb ? hcloud_load_balancer_network.control_plane.*.ip[0] :
+        module.control_planes[k].private_ipv4_address == module.control_planes[keys(module.control_planes)[0]].private_ipv4_address ?
+        module.control_planes[keys(module.control_planes)[1]].private_ipv4_address :
+      module.control_planes[keys(module.control_planes)[0]].private_ipv4_address}:6443"
+      token                       = random_password.k3s_token.result
+      disable-cloud-controller    = true
+      disable                     = local.disable_extras
+      kubelet-arg                 = local.kubelet_arg
+      kube-controller-manager-arg = local.kube_controller_manager_arg
+      flannel-iface               = local.flannel_iface
+      node-ip                     = module.control_planes[k].private_ipv4_address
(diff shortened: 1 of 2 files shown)


variables.tf
```diff
@@ -642,7 +642,7 @@ variable "enable_wireguard" {
 }
 
 variable "control_planes_custom_config" {
-  type        = map(any)
+  type        = any
```
mysticaltech, line 645: Why this change?

control_planes.tf
```diff
@@ -93,44 +129,34 @@ resource "null_resource" "control_planes" {
...
-      )
-    )
-
+    content     = yamlencode(local.k3s-config[each.key])
     destination = "/tmp/config.yaml"
   }
 
+  provisioner "remote-exec" {
+    inline = [
+      "mv /tmp/config.yaml /etc/rancher/k3s/config.yaml",
```
mysticaltech, line 138: Here we can use the same kind of custom logic used for k3s_registries to backup, and revert to the backed-up version if k3s is not successful in restarting.

mysticaltech:
@mraerino Very interesting initial work, left a few comments, and I would like that we also tackle agent nodes in the same way, but let's corner control planes first indeed 🙏

As for your questions above, the methodology that was used for k3s_registries seems to work, so we create a backup either way if initial or not, and if k3s fails to start we restaure the backup. In cases of change (not initial), this will ensure we do not break a working cluster.

variables.tf
mysticaltech: Yes, just saw, and merged it, makes sense! Thanks @mraerino 🙏

control_planes.tf
mysticaltech: @mraerino Do you need help here? It's a good PR, just need this small tweak.

Silvest89:
@mysticaltech This is taking way too long, I will be taking this over since I need it in my current cluster. And I have already been doing a lot of custom changes ;]

mysticaltech:
@Silvest89  PR most welcome, please see the logic behind `k3s_registries` for inspiration. We basically need the same thing.

mysticaltech:
@mraerino Thanks for starting this. @Silvest89 Built up the remaining needed functionality.

PR #886, opened 2023-07-12
drop default for var.hcloud_token

hvraven:
the token is used to create the hcloud secret in the kube-system namespace. Previously it could be left empty, breaking the setup during the kustomization stage. This change forces the user to set a value.

Closes: #885


mysticaltech (approved):
Thanks!!

PR #901, opened 2023-07-21
create control_plane_lb_type and default to lb11 so it is independent…

supernovae:
create control_plan_lb_type var and default it to lb11 so the control_plane load balancer is independently sized from ingress load balancer.


aleksasiriski (approved):

mysticaltech (approved):
Thanks @supernovae, LGTM 🚀

PR #908, opened 2023-07-26
Fix name for control-lb + prevent infinite waiting when using lb_host…

mnbro:
1. Correlates '"hcloud_load_balancer" "cluster"' name between control_planes.tf and data.tf
2. When using lb_hostname="some.fqdn" terraform is stuck in 'Waiting for load-balancer to get an IP...', this is because `kubectl get -n traefik service/traefik --output=jsonpath='{.status.loadBalancer.ingress[0]}'`
return only `{"ip":"1.1.1.1"}`

### DIFF
--- a/init.tf
+++ b/init.tf
@@ -320,7 +320,7 @@ resource "null_resource" "kustomization" {
       local.has_external_load_balancer ? [] : [
         <<-EOT
       timeout 360 bash <<EOF
-      until [ -n "\$(kubectl get -n ${lookup(local.ingress_controller_namespace_names, local.ingress_controller)} service/${lookup(local.ingress_controller_service_names, local.ingress_controller)} --output=jsonpath='{.status.loadBalancer.ingress[0].${var.lb_hostname != "" ? "hostname" : "ip"}}' 2> /dev/null)" ]; do
+      until [ -n "\$(kubectl get -n ${lookup(local.ingress_controller_namespace_names, local.ingress_controller)} service/${lookup(local.ingress_controller_service_names, local.ingress_controller)} --output=jsonpath='{.status.loadBalancer.ingress[0].ip}' 2> /dev/null)" ]; do
           echo "Waiting for load-balancer to get an IP..."
           sleep 2
       done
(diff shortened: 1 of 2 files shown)


init.tf
```diff
@@ -322,7 +322,7 @@ resource "null_resource" "kustomization" {
       local.has_external_load_balancer ? [] : [
         <<-EOT
       timeout 360 bash <<EOF
-      until [ -n "\$(kubectl get -n ${lookup(local.ingress_controller_namespace_names, local.ingress_controller)} service/${lookup(local.ingress_controller_service_names, local.ingress_controller)} --output=jsonpath='{.status.loadBalancer.ingress[0].${var.lb_hostname != "" ? "hostname" : "ip"}}' 2> /dev/null)" ]; do
+      until [ -n "\$(kubectl get -n ${lookup(local.ingress_controller_namespace_names, local.ingress_controller)} service/${lookup(local.ingress_controller_service_names, local.ingress_controller)} --output=jsonpath='{.status.loadBalancer.ingress[0].ip}' 2> /dev/null)" ]; do
```
aleksasiriski, line 325: If you use the option `lb_hostname` this will hang. This should be reverted (I fixed it a long time ago).

Is anyone even using a control plane lb for kubectl / ssh (if it's even configured to work with ssh)? @mysticaltech

aleksasiriski (changes requested):

aleksasiriski (approved):

mysticaltech (approved):

mysticaltech:
@mnbro Thanks for this fix! I wonder how the old logic ended up this way 🤦🏻

mysticaltech:
@mnbro Had to revert the change for the lb name in data.tf as it did not deploy and was not backward compatible, I missed it in the review, the rest was ok. Thanks for the main hanging fix and no worries!

mysticaltech:
@mnbro Yes, it's because you had changed it. You could try renaming it manually to k3s over at Hetzner. Also make sure to set the cluster variable.

mysticaltech:
Anyways, that change would break backward compatibility. On top of that, this lb is not associated with the control planes. There's another one for that.

PR #917, opened 2023-08-01
[Feature] Add enable_public_interface option for control plane lb

mnbro:
Add option to disable public interface for control plane LB


mysticaltech (approved):
Thanks for this @mnbro.

Issue #914, opened 2023-07-30
[Feature Request]: Using only private network


ceyhunn:
### Description

Hi,

It would be nice to use only private network (including servers, cp load balancer) with router (for internet access) and wireguard app server (for vpn access) for creating k3s clusters.

mysticaltech:
@ceyhunn Thanks for your suggestion! Private network is already possible, you can close all the incoming ports, please search for "firewall" in kube.tf.example. Other solutions have been discussed in the part but it's not practical to implement in this project directly. However you can add it yourself on the side, as this project can be a module that integrates in a greater terraform project.

ceyhunn:
Hi @mysticaltech , 
thank you for the answer, but hetzner firewall is applied to servers, not load balancers, hetzner load balancers can't be safe if public ip of it is enabled.

mysticaltech:
@ceyhunn How can the LB be reached if it doesn't have a public IP! 😅 And that's up to Hetzner really, they ensure the security of it.

M4t7e:
I think the only way would be routing from somewhere to the internal network via VPN or something similar. Some kind of internal-only app maybe. The use case should be solved with #917 and #918.

ceyhunn:
> I think the only way would be routing from somewhere to the internal network via VPN or something similar. Some kind of internal-only app maybe. The use case should be solved with #917 and #918.

Yes, that is I want. I already said about that. 

> @ceyhunn How can the LB be reached if it doesn't have a public IP! sweat_smile And that's up to Hetzner really, they ensure the security of it.

Not correct, I think, control plane load balancer and k3s servers have to be private and access to them through private network by vpn, also not need to set additional firewall rules and pay additional money for the servers' public ip.  Steps I can share with you:
1. Create wireguard vpn server (with public ip enabled)
2. Create router server in order to give internet access for servers (with public ip enabled).
3. Create private network for servers 
4. And eventually steps that creating a k3s cluster.

ceyhunn:
Hi @M4t7e ,
Can you please add option to disable public ip of servers, and connect to servers with ssh via private ip, which created by private network.

mysticaltech:
@ceyhunn If you close the ssh port, the servers are basically inaccessible from the outside world. No need to disable the public interface, and this is not something we are considering doing right now.

PR #918, opened 2023-08-01
Added option to disable public lb network

M4t7e:
Added option to disable the public load balancer network.
Together with https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/917 it should solve the use case behind https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/914 completely.


M4t7e:
A bit more context what can be achieved with this. Basically this option controls the `Public Network` setting from the load balancer:
image

If you want to "connect" your K3s project in Hetzner with another project or a dedicated infrastructure via a VPN and want to have internal-only applications (not exposed to the public internet) behind the LB, you can disable the public network access to the LB and still reach it via the internal network IP address (the same internal address the LB uses to connect to the ingress controller nodes).

mysticaltech (approved):
Makes sense, thanks @M4t7e!

PR #924, opened 2023-08-04
Revert to the old LB Availability Check

M4t7e:
Revert the last left-over from #908 so that the LB availability check is also performed when `lb_hostname` is used.

I tested it in one cluster deployment using `lb_hostname` and it seems to work fine.


mysticaltech (approved):
Wonderful!

PR #925, opened 2023-08-04
fix waiting for load-balancer IP in case no IP is assigned

MoergJ:
fix waiting for load-balancer IP in case no IP is assigned, but a hostname (e.g. when lb_hostname is set)


MoergJ:
Change basically already in staging

PR #926, opened 2023-08-05
Unify NetworkManager Connection Names

M4t7e:
This PR is mainly a cosmetic change to unify NetworkManager connection names to make life easier for humans interacting with `nmcli`.

Before the change:
```bash
# nmcli connection show
NAME                UUID                                  TYPE      DEVICE 
cloud-init eth0     1dd9a779-d327-56e1-8454-c65e2556c12c  ethernet  eth0   
Wired connection 1  fdcd1a0a-bb69-3017-a940-96773f29fb8d  ethernet  eth1   
lo                  9e961e02-e328-4a5f-98ed-227fbf4b189d  loopback  lo 
```

After the change:
```bash
# nmcli connection show
NAME  UUID                                  TYPE      DEVICE 
eth0  1dd9a779-d327-56e1-8454-c65e2556c12c  ethernet  eth0   
eth1  2a211c06-56de-35bb-967d-5b0c8378449e  ethernet  eth1   
lo    086d46a4-41f2-4089-81ce-0517216dbe80  loopback  lo  
```

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -576,7 +576,7 @@ fi
 EOF
 
   cloudinit_write_files_common = <<EOT
-# Script to rename the private interface to eth1
+# Script to rename the private interface to eth1 and unify NetworkManager connection naming
 - path: /etc/cloud/rename_interface.sh
   content: |
     #!/bin/bash
@@ -594,6 +594,18 @@ EOF
     ip link set $INTERFACE down
     ip link set $INTERFACE name eth1
     ip link set eth1 up
+
+    eth0_connection=$(nmcli -g GENERAL.CONNECTION device show eth0)
+    nmcli connection modify "$eth0_connection" \
+      con-name eth0 \
+      connection.interface-name eth0
+
+    eth1_connection=$(nmcli -g GENERAL.CONNECTION device show eth1)
+    nmcli connection modify "$eth1_connection" \
+      con-name eth1 \
+      connection.interface-name eth1
+
+    systemctl restart NetworkManager
   permissions: "0744"
 
 # Disable ssh password authentication


mysticaltech (approved):
@M4t7e LGTM, tested and running well huh?

M4t7e:
@mysticaltech Yes, I test everything before creating a PR. But tbh, I can only test to some extend because we have a lot of different settings, so I can't test manually all combinations. Agent, Server and Egress Nodes are working fine with this change and I tested it with a few cluster deployments.

The only place where we use `nmcli` is here: https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/9f2e7ebde6617e864cf608709fc8a0dd8e07aee2/agents.tf#L180-L187
Here the connection name is obtained by the device name `eth0`. So no hard-coded connection name is used.

Other places where NetworkManager is touched are here:
- https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/9f2e7ebde6617e864cf608709fc8a0dd8e07aee2/locals.tf#L705-L719
- https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/9f2e7ebde6617e864cf608709fc8a0dd8e07aee2/packer-template/hcloud-microos-snapshots.pkr.hcl#L70-L77

My assumption currently is, that cloud-init will never touch the config again after the very first init. If it would so, only the `eth0` connection name is reverted back to `cloud-init eth0`. `eth1` seems to be not managed by cloud-init at all. I tested that as well.

I'm in no hurry with this PR. So if anyone wants to take a look, I'd appreciate that :slightly_smiling_face: 

PS: A re-run of cloud-init would revert not only the `eth0` connection name in NetworkManager back to the old one, but would also remove the config from [agents.tf](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/9f2e7ebde6617e864cf608709fc8a0dd8e07aee2/agents.tf#L180-L187). The floating IP configuration would completely disappear! Maybe in the future we can find a better way to integrate nicely with cloud-init in general rather than overriding what it did on the first pass. Then we could run it again without worrying.

mysticaltech:
Thanks for the explanation @M4t7e! Yes, in the future if we can make the whole flow that we currently use more robust to cloud-init it would be great. But normally it shouldn't run again. But if it does, we can run these scripts automatically after each reboot.

Btw, I also do not test every combination, just one test is good enough 🙏

mysticaltech:
I will merge it ASAP today or tomorrow.

PR #930, opened 2023-08-08
Performance Improvements and Native Routing Mode for Cilium

M4t7e:
This PR modifies the Cilium configuration to achieve the best possible performance and also native routing functionality for Cilium, allowing configuration of a fully integrated network model. For more background information, please have a look here: https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/discussions/911

**Changes:**
- Fixed decoupling issue between CCM IPAM and the Cilium IPAM. Now Cilium honors the IPAM from the CCM and Hetzner Network Routes are set accordingly to allow native routing directly to pod IPs as well. See also https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/902#issuecomment-1648189450
- Updated Cilium options according to the latest deprecation warnings: https://docs.cilium.io/en/latest/operations/upgrade/#deprecated-options
- The Cilium version is now maintained explicitly. This PR sets it to 1.14.0 and some settings require that as the minimum version.
- Enabled per endpoint routes instead of routing via the `cilium_host` interface.
- Native routing mode has been added so that even a fully integrated network model (flat model) can be configured with Cilium
- Enabled eBPF-based masquerading, considered the most efficient implementation.
- Added LoadBalancer & NodePort XDP Acceleration (eBPF is operating directly in the networking driver instead of a higher layer)
- Added option to enable the Egress Gateway for Cilium (disabled by default)

The native routing mode is disabled by default as we have been using tunnel mode so far. As we gain more experience with this setup, I would suggest changing the default to native. This also appears to be the setting Hetzner uses when testing their CCM with Cilium.

I've done my best to test many different cases, but I'd be very grateful for a thorough review :slightly_smiling_face:

### DIFF
--- a/README.md
+++ b/README.md
@@ -372,17 +372,21 @@ cluster_ipv4_cidr = local.cluster_ipv4_cidr
 
 cilium_values = <<EOT
 ipam:
-  operator:
-    clusterPoolIPv4PodCIDRList:
-      - ${local.cluster_ipv4_cidr}
-kubeProxyReplacement: strict
-l7Proxy: "false"
+  mode: kubernetes
+k8s:
+  requireIPv4PodCIDR: true
+kubeProxyReplacement: true
+routingMode: native
+ipv4NativeRoutingCIDR: "10.0.0.0/8"
+endpointRoutes:
+  enabled: true
+loadBalancer:
+  acceleration: native
 bpf:
-  masquerade: "true"
+  masquerade: true
 egressGateway:
-  enabled: "true"
-extraConfig:
-  mtu: "1450"
+  enabled: true
+MTU: 1450
 EOT
 ```
 
--- a/kube.tf.example
+++ b/kube.tf.example
@@ -532,6 +532,18 @@ module "kube-hetzner" {
   # Also, see the cilium_values at towards the end of this file, in the advanced section.
   # cni_plugin = "cilium"
 
+  # You can choose the version of Cilium that you want. By default we keep the version up to date and configure Cilium with compatible settings according to the version.
+  # cilium_version = "v1.14.0"
+  
+  # Set native-routing mode ("native") or tunneling mode ("tunnel"). Default: tunnel
+  # cilium_routing_mode = "native"
+
+  # Used when Cilium is configured in native routing mode. The CNI assumes that the underlying network stack will forward packets to this destination without the need to apply SNAT. Default: value of "cluster_ipv4_cidr"
+  # cilium_ipv4_native_routing_cidr = "10.0.0.0/8"
+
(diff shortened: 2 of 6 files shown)


mysticaltech (approved):
@M4t7e Beautiful! That is a work of art my friend 🙏

mysticaltech:
Will give 24h-48h for others in the team to review if they want, and will merge, this is huge! 🚀

README.md
```diff
@@ -372,17 +372,21 @@ cluster_ipv4_cidr = local.cluster_ipv4_cidr
...
+  acceleration: native
 bpf:
-  masquerade: "true"
+  masquerade: true
 egressGateway:
-  enabled: "true"
-extraConfig:
-  mtu: "1450"
+  enabled: true
+MTU: 1450
```
ifeulner, line 389: Is the explicit setting go the MTU necessary? - does the auto detection not work on Hetzner? 
Just curious...

kube.tf.example
```diff
@@ -662,20 +674,26 @@ module "kube-hetzner" {
...
+loadBalancer:
+  acceleration: native
+bpf:
+  masquerade: true
 encryption:
   enabled: true
   type: wireguard
-extraConfig:
-  mtu: "1450"
+MTU: 1450
```
ifeulner, line 696: see above

locals.tf
```diff
@@ -349,19 +351,43 @@ locals {
...
   type: wireguard
 %{endif~}
-extraConfig:
-  mtu: "1450"
+%{if var.cilium_egress_gateway_enabled}
+egressGateway:
+  enabled: true
+%{endif~}
+
+MTU: 1450
```
ifeulner, line 390: see above

ifeulner (approved):
@M4t7e cool stuff need to test this asap, but currently working on other areas. But thanks a lot!

aleksasiriski:
@M4t7e Are there any known issues with native routing on Hetzner?

M4t7e:
@aleksasiriski No issues I am aware of. With the new setup the Hetzner CCM takes care about all Pod routes and updates the Hetzner Network Routes accordingly. The old setup bypassed the Hetzner CCM entirely and that was leading to connectivity issues if a native routing setup (`routingMode: native`) was used.

Old (dedicated IPAM that bypasses HCCM):
```yml
ipam:
 operator:
  clusterPoolIPv4PodCIDRList:
   - ${var.cluster_ipv4_cidr}
```

New (native K8s IPAM with HCCM support):
```yml
ipam:
  mode: kubernetes
k8s:
  requireIPv4PodCIDR: true
```

PR #946, opened 2023-08-27
Add k3s-token and k3s-endpoint as output so other modules can re-use it.

nupplaphil:
Hi there,

I'm currently building a terraform environment to add root-server to this cluster.
This is my current state of work: https://github.com/nupplaphil/terraform-root-kube-hetzner

Since I don't want to mix up root server and hcloud server, I choose a different repo for it.

The only thing missing is the "glue" between the hcloud terraform module and my root terraform module.
--> the k3s token and endpoint is missing.

With these two outputs, I was successfully to add a root-server to a cluster out of this project :)

btw the cilium config
```yaml
loadBalancer:
  # Enable LoadBalancer & NodePort XDP Acceleration (direct routing (routingMode=native) is recommended to achieve optimal performance)
  acceleration: native
```

doesn't work because XDP is incompatible with the vlan interface of the vSwitch subnet.
--> I manually overwrite the `cilium_values`

### DIFF
--- a/output.tf
+++ b/output.tf
@@ -37,6 +37,14 @@ output "ingress_public_ipv6" {
   value       = (local.has_external_load_balancer || var.load_balancer_disable_ipv6) ? null : data.hcloud_load_balancer.cluster[0].ipv6
 }
 
+output "k3s_endpoint" {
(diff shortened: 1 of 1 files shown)


mysticaltech (approved):
Cool, super interesting!

mysticaltech:
@nupplaphil If you could add an example on how to use metal in the examples folder, it would be great. Including your custom cilium values. As a lot of people want to use this.

mysticaltech:
@schlichtanders Doesn't that answer the need of #953 and #949?

PR #953, opened 2023-08-30
Possibility to set k3s_token manually for restoring a cluster

### LINKED ISSUE #949: k3s_token is not exported to the user
### Summary
First time I am trying the etcd backup recovery and to my surprises, the k3s token is needed for restoration. 
https://docs.k3s.io/datastore/backup-restore

This token however is a randomly created string which is never stored on the user side.
Hence, if the cluster crashes, there is no way to restore even if etcd backup was setup correctly (at least this is my current understanding)

### PoC

I tried to restore a backup testwise using k3s on docker
```bash
docker run \
  --privileged \
  --name k3s-server-1 \
  --hostname k3s-server-1 \
  -p 6443:6443 \
  -it --rm \
  rancher/k3s:v1.26.7-k3s1 \
  server \
  --cluster-init \
  --cluster-reset \
  --etcd-s3 \
  --cluster-reset-restore-path=$SNAPSHOT_NAME \
  --etcd-s3-endpoint=$TF_VAR_etcd_s3_endpoint \
  --etcd-s3-bucket=$TF_VAR_etcd_s3_bucket \
  --etcd-s3-access-key=$TF_VAR_etcd_s3_access_key \
  --etcd-s3-secret-key=$TF_VAR_etcd_s3_secret_key
```
It outputs
```
WARN[0000] remove /var/lib/rancher/k3s/agent/etc/k3s-agent-load-balancer.json: no such file or directory 
WARN[0000] remove /var/lib/rancher/k3s/agent/etc/k3s-api-server-agent-load-balancer.json: no such file or directory 
FATA[0000] /var/lib/rancher/k3s/server/token does not exist, please pass --token to complete the restoration 
```

Of course to replicate on another end, you need a backup (on S3) and set the environment variables (or the server args) respectively.

### Impact
etcd backup is not working for a whole cluster loss

### Kube.tf file

```terraform
locals {
  # You have the choice of setting your Hetzner API token here or define the TF_VAR_hcloud_token env
  # within your shell, such as such: export TF_VAR_hcloud_token=xxxxxxxxxxx
  # If you choose to define it in the shell, this can be left as is.

  # Your Hetzner token can be found in your Project > Security > API Token (Read & Write is required).
  hcloud_token = "xxxxxxxxxxx"
}

module "kube-hetzner" {
  providers = {
    hcloud = hcloud
  }
  hcloud_token = var.hcloud_token != "" ? var.hcloud_token : local.hcloud_token

  # Then fill or edit the below values. Only the first values starting with a * are obligatory; the rest can remain with their default values, or you
  # could adapt them to your needs.

  # * For local dev, path to the git repo
  # source = "../../kube-hetzner/"
  # If you want to use the latest master branch
  # source = "github.com/kube-hetzner/terraform-hcloud-kube-hetzner"
  # For normal use, this is the path to the terraform registry
  source = "kube-hetzner/kube-hetzner/hcloud"

  ###################
  # FIXING VERSIONS #
  ###################

  # You can optionally specify a version number
  version = "2.2.0"

  # https://github.com/hetznercloud/hcloud-cloud-controller-manager
  hetzner_ccm_version = "v1.15.0"
  # https://github.com/hetznercloud/csi-driver
  hetzner_csi_version = "v2.3.2"

  # If you want to specify the Kured version, set it below - otherwise it'll use the latest version available.
  # https://github.com/kubereboot/kured
  kured_version = "1.13.1"

  # Allows you to specify either stable, latest, testing or supported minor versions.
  # see https://rancher.com/docs/k3s/latest/en/upgrades/basic/ and https://update.k3s.io/v1-release/channels
  # ⚠️ If you are going to use Rancher addons for instance, it's always a good idea to fix the kube version to latest - 0.01,
  # ⚠️ Rancher currently only supports v1.25 and earlier versions: https://github.com/rancher/rancher/issues/41113
  # The default is "v1.26".
  initial_k3s_channel = "v1.26"

  # You can choose the version of Calico that you want. By default, the latest is used.
  # More info on available versions can be found at https://github.com/projectcalico/calico/releases
  # Please note that if you are getting 403s from Github, it's also useful to set the version manually. However there is rarely a need for that!
  calico_version = "v3.26.0"

  #######################
  # END FIXING VERSIONS #
  #######################


  # Note that some values, notably "location" and "public_key" have no effect after initializing the cluster.
  # This is to keep Terraform from re-provisioning all nodes at once, which would lose data. If you want to update
  # those, you should instead change the value here and manually re-provision each node. Grep for "lifecycle".

  # Customize the SSH port (by default 22)
  # ssh_port = 2222

  # * Your ssh public key
  ssh_public_key = file("~/.ssh/hetzner_id_ed25519.pub")
  # * Your private key must be "ssh_private_key = null" when you want to use ssh-agent for a Yubikey-like device authentification or an SSH key-pair with a passphrase.
  ssh_private_key = file("~/.ssh/hetzner_id_ed25519")

  # You can also add additional SSH public Keys which are saved in the hetzner cloud by a label.
  # See https://docs.hetzner.cloud/#label-selector
  # ssh_hcloud_key_label = "role=admin"

  # If you want to use an ssh key that is already registered within hetzner cloud, you can pass its id.
  # If no id is passed, a new ssh key will be registered within hetzner cloud.
  # It is important that exactly this key is passed via `ssh_public_key` & `ssh_private_key` vars.
  # hcloud_ssh_key_id = ""

  # If you must change the network CIDR you can do so below, but it is highly advised against.
  # network_ipv4_cidr = "10.0.0.0/8"

  # If you must change the cluster CIDR you can do so below, but it is highly advised against.
  # Cluster CIDR must be a part of the network CIDR!
  # cluster_ipv4_cidr = "10.42.0.0/16"

  # For the control planes, at least three nodes are the minimum for HA. Otherwise, you need to turn off the automatic upgrades (see README).
  # **It must always be an ODD number, never even!** Search the internet for "splitbrain problem with etcd" or see https://rancher.com/docs/k3s/latest/en/installation/ha-embedded/
  # For instance, one is ok (non-HA), two is not ok, and three is ok (becomes HA). It does not matter if they are in the same nodepool or not! So they can be in different locations and of various types.

  # Of course, you can choose any number of nodepools you want, with the location you want. The only constraint on the location is that you need to stay in the same network region, Europe, or the US.
  # For the server type, the minimum instance supported is cpx11 (just a few cents more than cx11); see https://www.hetzner.com/cloud.

  # IMPORTANT: Before you create your cluster, you can do anything you want with the nodepools, but you need at least one of each, control plane and agent.
  # Once the cluster is up and running, you can change nodepool count and even set it to 0 (in the case of the first control-plane nodepool, the minimum is 1).
  # You can also rename it (if the count is 0), but do not remove a nodepool from the list.

  # The only nodepools that are safe to remove from the list are at the end. That is due to how subnets and IPs get allocated (FILO).
  # You can, however, freely add other nodepools at the end of each list if you want. The maximum number of nodepools you can create combined for both lists is 255.
  # Also, before decreasing the count of any nodepools to 0, it's essential to drain and cordon the nodes in question. Otherwise, it will leave your cluster in a bad state.

  # Before initializing the cluster, you can change all parameters and add or remove any nodepools. You need at least one nodepool of each kind, control plane, and agent.
  # The nodepool names are entirely arbitrary, you can choose whatever you want, but no special characters or underscore, and they must be unique; only alphanumeric characters and dashes are allowed.

  # If you want to have a single node cluster, have one control plane nodepools with a count of 1, and one agent nodepool with a count of 0.

  # Please note that changing labels and taints after the first run will have no effect. If needed, you can do that through Kubernetes directly.

  # ⚠️ When choosing ARM cax* server types, for the moment they are only available in fsn1.
  # Muli-architecture clusters are OK for most use cases, as container underlying images tend to be multi-architecture too.

  # * Example below:

  control_plane_nodepools = [
    {
      server_type = "cax11",
      location    = "fsn1",
      labels      = [
        # no longer needed, but may be interesting when using klipper as the load balancer
        # as this seems to be the final option which counts
        # taken from https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/447#issuecomment-1350597300
        # "node.kubernetes.io/exclude-from-external-load-balancers=true",
      ],
      taints      = [],
      count       = 1

      # backups = true
    },
  ]

  agent_nodepools = [
    # Arm based nodes, currently available only in FSN location
    {
      name        = "agent-arm-largest",
      server_type = "cax41",
      location    = "fsn1",
      labels      = [],
      taints      = [],
      count       = 1,
    }
  ]
  # E.g to enable monitoring for etcd, proxy etc:
  # control_planes_custom_config = {
  #  etcd-expose-metrics = true,
  #  kube-controller-manager-arg = "bind-address=0.0.0.0",
  #  kube-proxy-arg ="metrics-bind-address=0.0.0.0",
  #  kube-scheduler-arg = "bind-address=0.0.0.0",
  # }

  # You can enable encrypted wireguard for the CNI by setting this to "true". Default is "false".
  # FYI, Hetzner says "Traffic between cloud servers inside a Network is private and isolated, but not automatically encrypted."
  # Source: https://docs.hetzner.com/cloud/networks/faq/#is-traffic-inside-hetzner-cloud-networks-encrypted
  # It works with all CNIs that we support.
  # Just note, that if Cilium with cilium_values, the responsability of enabling of disabling Wireguard falls on you.
  enable_wireguard = true

  ### The following values are entirely optional (and can be removed from this if unused)

  # You can refine a base domain name to be use in this form of nodename.base_domain for setting the reserve dns inside Hetzner
  base_domain = "cloud.jolin.io"

  # Cluster Autoscaler
  # Providing at least one map for the array enables the cluster autoscaler feature, default is disabled
  # By default we set a compatible version with the default initial_k3s_channel, to set another one,
  # have a look at the tag value in https://github.com/kubernetes/autoscaler/blob/master/charts/cluster-autoscaler/values.yaml
  # ⚠️ Based on how the autoscaler works with this project, you can only choose either x86 instances or ARM server types for ALL autocaler nodepools.
  # Also, as mentioned above, for the time being ARM cax* instances are only available in fsn1.
  # If you are curious, it's ok to have a multi-architecture cluster, as most underlying container images are multi-architecture too.
  # * Example below:
  autoscaler_nodepools = [
    {
      name        = "autoscaled-arm-largest"
      server_type = "cax41"
      location    = "fsn1"
      min_nodes   = 0  # somehow a first autoscaler node is spawned even if `min_nodes = 0`, for updates see https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/756
      max_nodes   = 4
    }
  ]

  # Enable etcd snapshot backups to S3 storage.
  # Just provide a map with the needed settings (according to your S3 storage provider) and backups to S3 will
  # be enabled (with the default settings for etcd snapshots).
  # Cloudflare's R2 offers 10GB, 10 million reads and 1 million writes per month for free.
  # For proper context, have a look at https://docs.k3s.io/backup-restore.
  etcd_s3_backup = {
    etcd-s3-endpoint   = var.etcd_s3_endpoint
    etcd-s3-access-key = var.etcd_s3_access_key
    etcd-s3-secret-key = var.etcd_s3_secret_key
    etcd-s3-bucket     = var.etcd_s3_bucket
  }

  # To enable Hetzner Storage Box support, you can enable csi-driver-smb, default is "false".
  # enable_csi_driver_smb = true

  # To use local storage on the nodes, you can enable Longhorn, default is "false".
  # See a full recap on how to configure agent nodepools for longhorn here https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/discussions/373#discussioncomment-3983159
  # Also see Longhorn best practices here https://gist.github.com/ifeulner/d311b2868f6c00e649f33a72166c2e5b
  # enable_longhorn = true

  # By default, longhorn is pulled from https://charts.longhorn.io.
  # If you need a version of longhorn which assures compatibility with rancher you can set this variable to https://charts.rancher.io.
  # longhorn_repository = "https://charts.rancher.io"

  # The namespace for longhorn deployment, default is "longhorn-system".
  # longhorn_namespace = "longhorn-system"

  # The file system type for Longhorn, if enabled (ext4 is the default, otherwise you can choose xfs).
  # longhorn_fstype = "xfs"

  # how many replica volumes should longhorn create (default is 3).
  # longhorn_replica_count = 1

  # When you enable Longhorn, you can go with the default settings and just modify the above two variables OR you can add a longhorn_values variable
  # with all needed helm values, see towards the end of the file in the advanced section.
  # If that file is present, the system will use it during the deploy, if not it will use the default values with the two variable above that can be customized.
  # After the cluster is deployed, you can always use HelmChartConfig definition to tweak the configuration.

  # Also, you can choose to use a Hetzner volume with Longhorn. By default, it will use the nodes own storage space, but if you add an attribute of
  # longhorn_volume_size (⚠️ not a variable, just a possible agent nodepool attribute) with a value between 10 and 10000 GB to your agent nodepool definition, it will create and use the volume in question.
  # See the agent nodepool section for an example of how to do that.

  # To disable Hetzner CSI storage, you can set the following to "true", default is "false".
  # disable_hetzner_csi = true

  # https://github.com/hetznercloud/hcloud-cloud-controller-manager
  # hetzner_ccm_version = "v1.15.0"
  # https://github.com/hetznercloud/csi-driver
  # hetzner_csi_version = "v2.3.2"

  # If you want to specify the Kured version, set it below - otherwise it'll use the latest version available.
  # https://github.com/kubereboot/kured
  # kured_version = "1.13.1"

  # If you want to enable the Nginx ingress controller (https://kubernetes.github.io/ingress-nginx/) instead of Traefik, you can set this to "nginx". Default is "traefik".
  # By the default we load optimal Traefik and Nginx ingress controller config for Hetzner, however you may need to tweak it to your needs, so to do,
  # we allow you to add a traefik_values and nginx_values, see towards the end of this file in the advanced section.
  # After the cluster is deployed, you can always use HelmChartConfig definition to tweak the configuration.
  # If you want to disable both controllers set this to "none"
  # ingress_controller = "nginx"

  # You can change the number of replicas for selected ingress controller here. The default 0 means autoselecting based on number of agent nodes (1 node = 1 replica, 2 nodes = 2 replicas, 3+ nodes = 3 replicas)
  # ingress_replica_count = 1

  # Use the klipperLB (similar to metalLB), instead of the default Hetzner one, that has an advantage of dropping the cost of the setup.
  # Automatically "true" in the case of single node cluster (as it does not make sense to use the Hetzner LB in that situation).
  # It can work with any ingress controller that you choose to deploy.
  # Please note that because the klipperLB points to all nodes, we automatically allow scheduling on the control plane when it is active.
  # enable_klipper_metal_lb = "true"

  # If you want to configure additional arguments for traefik, enter them here as a list and in the form of traefik CLI arguments; see https://doc.traefik.io/traefik/reference/static-configuration/cli/
  # They are the options that go into the additionalArguments section of the Traefik helm values file.
  # Example: traefik_additional_options = ["--log.level=DEBUG", "--tracing=true"]
  # traefik_additional_options = []

  # By default traefik is configured to redirect http traffic to https, you can set this to "false" to disable the redirection.
  # traefik_redirect_to_https = false

  # If you want to disable the metric server set this to "false". Default is "true".
  # enable_metrics_server = false

  # If you want to allow non-control-plane workloads to run on the control-plane nodes, set this to "true". The default is "false".
  # True by default for single node clusters, and when enable_klipper_metal_lb is true. In those cases, the value below will be ignored.
  # allow_scheduling_on_control_plane = true

  # If you want to disable the automatic upgrade of k3s, you can set below to "false".
  # Ideally, keep it on, to always have the latest Kubernetes version, but lock the initial_k3s_channel to a kube major version,
  # of your choice, like v1.25 or v1.26. That way you get the best of both worlds without the breaking changes risk.
  # For production use, always use an HA setup with at least 3 control-plane nodes and 2 agents, and keep this on for maximum security.

  # The default is "true" (in HA setup i.e. at least 3 control plane nodes & 2 agents, just keep it enabled since it works flawlessly).
  # automatically_upgrade_k3s = false

  # The default is "true" (in HA setup it works wonderfully well, with automatic roll-back to the previous snapshot in case of an issue).
  # IMPORTANT! For non-HA clusters i.e. when the number of control-plane nodes is < 3, you have to turn it off.
  automatically_upgrade_os = false

  # If you need more control over kured and the reboot behaviour, you can pass additional options to kured.
  # For example limiting reboots to certain timeframes. For all options see: https://kured.dev/docs/configuration/
  # The default options are: `--reboot-command=/usr/bin/systemctl reboot --pre-reboot-node-labels=kured=rebooting --post-reboot-node-labels=kured=done --period=5m`
  # Defaults can be overridden by using the same key.
  # kured_options = {
  #   "reboot-days": "su"
  #   "start-time": "3am"
  #   "end-time": "8am"
  #   "time-zone": "Local"
  # }

  # Allows you to specify either stable, latest, testing or supported minor versions.
  # see https://rancher.com/docs/k3s/latest/en/upgrades/basic/ and https://update.k3s.io/v1-release/channels
  # ⚠️ If you are going to use Rancher addons for instance, it's always a good idea to fix the kube version to latest - 0.01,
  # ⚠️ Rancher currently only supports v1.25 and earlier versions: https://github.com/rancher/rancher/issues/41113
  # The default is "v1.26".
  # initial_k3s_channel = "stable"

  # The cluster name, by default "k3s"
  cluster_name = "jolin"

  # Extra k3s registries. This is useful if you have private registries and you want to pull images without additional secrets.
  # Or if you want to proxy registries for various reasons like rate-limiting.
  # It will create the registries.yaml file, more info here https://docs.k3s.io/installation/private-registry.
  # Note that you do not need to get this right from the first time, you can update it when you want during the life of your cluster.
  # The default is blank.
  /* k3s_registries = <<-EOT
    mirrors:
      hub.my_registry.com:
        endpoint:
          - "hub.my_registry.com"
    configs:
      hub.my_registry.com:
        auth:
          username: username
          password: password
  EOT */

  # Additional environment variables for the host OS on which k3s runs. See for example https://docs.k3s.io/advanced#configuring-an-http-proxy .
  # additional_k3s_environment = {
  #   "CONTAINERD_HTTP_PROXY" : "http://your.proxy:port",
  #   "CONTAINERD_HTTPS_PROXY" : "http://your.proxy:port",
  #   "NO_PROXY" : "127.0.0.0/8,10.0.0.0/8,",
  # }

  # Additional commands to execute on the host OS before the k3s install, for example fetching and installing certs.
  # preinstall_exec = [
  #   "curl https://somewhere.over.the.rainbow/ca.crt > /root/ca.crt",
  #   "trust anchor --store /root/ca.crt",
  # ]
  preinstall_exec = [
    # This is adding node-taints to the config
    # it is safest to do it here, because node-taints may have already been added, and setting --node-taint attribute to k3s agent args will clean out all the other node-taints
    # we check whether the csi-node-driver can be loaded
    # if we get 403 response, this means that we got a bad IP
    # adapted from https://stackoverflow.com/questions/53526188/can-i-have-curl-print-just-the-response-code
    "echo 'I am here' > /tmp/iamhere"
    ,
    # <<-EOF
    # [ -f /etc/rancher/k3s/config.yaml ] && \
    # [ "403" == $(curl -IL --silent --write-out "%{http_code}\n" -o /dev/null https://registry.k8s.io/v2/sig-storage/csi-node-driver-registrar/manifests/v2.7.0) ] && \
    # sed -i -e 's/^"node-taint": \[]/"node-taint":/' -e 's/^"node-taint":/"node-taint":\n- jolin=badip:NoSchedule/' /etc/rancher/k3s/config.yaml
    # EOF
    # ,

    # we can simply add a second config to be aggregated
    # see https://docs.k3s.io/installation/configuration
    <<-EOF
    mkdir -p /etc/rancher/k3s/config.yaml.d && \
    [ "403" == $(curl -IL --silent --write-out "%{http_code}\n" -o /dev/null https://registry.k8s.io/v2/sig-storage/csi-node-driver-registrar/manifests/v2.7.0) ] && \
    cat > /etc/rancher/k3s/config.yaml.d/jolin.yaml <<END
    node-taint+:
    - jolin=badip:NoSchedule
    END
    EOF
    ,
  ]

  # Additional flags to pass to the k3s server command (the control plane).
  # k3s_exec_server_args = "--kube-apiserver-arg enable-admission-plugins=PodTolerationRestriction,PodNodeSelector"
  # we need to repeat the default kubelet-arg as CLI arguments take precendence. See https://docs.k3s.io/installation/configuration#configuration-file
  # default kubelet_arg = ["cloud-provider=external", "volume-plugin-dir=/var/lib/kubelet/volumeplugins"]

  # If you want to allow all outbound traffic you can set this to "false". Default is "true".
  # restrict_outbound_traffic = false

  # Adding extra firewall rules, like opening a port
  # More info on the format here https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs/resources/firewall
  extra_firewall_rules = [
    # {
    #   description = "For Postgres"
    #   direction       = "in"
    #   protocol        = "tcp"
    #   port            = "5432"
    #   source_ips      = ["0.0.0.0/0", "::/0"]
    #   destination_ips = [] # Won't be used for this rule
    # },
    # {
    #   description = "To Allow ArgoCD (or ssh-keyscan) access to resources via SSH"
    #   direction       = "out"
    #   protocol        = "tcp"
    #   port            = "22"
    #   source_ips      = [] # Won't be used for this rule
    #   destination_ips = ["0.0.0.0/0", "::/0"]
    # }
    {
      description = "Allow any outward access. To Allow ArgoCD (or ssh-keyscan) access to resources via SSH, access to Databases via special ports, etc."
      direction       = "out"
      protocol        = "tcp"
      port            = "any"
    }
  ]

  # If you want to configure a different CNI for k3s, use this flag
  # possible values: flannel (Default), calico, and cilium
  # As for Cilium, we allow infinite configurations via helm values, please check the CNI section of the readme over at https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/#cni.
  # Also, see the cilium_values at towards the end of this file, in the advanced section.
  # cni_plugin = "cilium"

  # You can choose the version of Calico that you want. By default, the latest is used.
  # More info on available versions can be found at https://github.com/projectcalico/calico/releases
  # Please note that if you are getting 403s from Github, it's also useful to set the version manually. However there is rarely a need for that!
  # calico_version = "v3.25.0"

  # If you want to disable the k3s default network policy controller, use this flag!
  # Both Calico and Ciliun cni_plugin values override this value to true automatically, the default is "false".
  # disable_network_policy = true

  # If you want to disable the automatic use of placement group "spread". See https://docs.hetzner.com/cloud/placement-groups/overview/
  # We advise to not touch that setting, unless you have a specific purpose.
  # The default is "false", meaning it's enabled by default.
  # placement_group_disable = true

  # By default, we allow ICMP ping in to the nodes, to check for liveness for instance. If you do not want to allow that, you can. Just set this flag to true (false by default).
  # block_icmp_ping_in = true

  # You can enable cert-manager (installed by Helm behind the scenes) with the following flag, the default is "true".
  # enable_cert_manager = false

  # IP Addresses to use for the DNS Servers, set to an empty list to use the ones provided by Hetzner, defaults to ["1.1.1.1", "8.8.8.8", "9.9.9.9"].
  # The number of different DNS servers is limited to 3 by Kubernetes itself.
  # dns_servers = []

  # When this is enabled, rather than the first node, all external traffic will be routed via a control-plane loadbalancer, allowing for high availability.
  # The default is false.
  # see https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/discussions/302
  use_control_plane_lb = false

  # Let's say you are not using the control plane LB solution above, and still want to have one hostname point to all your control-plane nodes.
  # You could create multiple A records of to let's say cp.cluster.my.org pointing to all of your control-plane nodes ips.
  # In which case, you need to define that hostname in the k3s TLS-SANs config to allow connection through it. It can be hostnames or IP addresses.
  # additional_tls_sans = ["cp.cluster.my.org"]

  # Oftentimes, you need to communicate to the cluster from inside the cluster itself, in which case it is important to set this value, as it will configure the hostname
  # at the load balancer level, and will save you from many slows downs when initiating communications from inside. Later on, you can point your DNS to the IP given
  # to the LB. And if you have other services pointing to it, you are also free to create CNAMES to point to it, or whatever you see fit.
  # If set, it will apply to either ingress controllers, Traefik or Ingress-Nginx.
  lb_hostname = "cloud.jolin.io"

  # You can enable Rancher (installed by Helm behind the scenes) with the following flag, the default is "false".
  # ⚠️ Rancher currently only supports Kubernetes v1.25 and earlier, you will need to set initial_k3s_channel to a supported version: https://github.com/rancher/rancher/issues/41113
  # When Rancher is enabled, it automatically installs cert-manager too, and it uses rancher's own self-signed certificates.
  # See for options https://rancher.com/docs/rancher/v2.0-v2.4/en/installation/resources/advanced/helm2/helm-rancher/#choose-your-ssl-configuration
  # The easiest thing is to leave everything as is (using the default rancher self-signed certificate) and put Cloudflare in front of it.
  # As for the number of replicas, by default it is set to the numbe of control plane nodes.
  # You can customized all of the above by adding a rancher_values variable see at the end of this file in the advanced section.
  # After the cluster is deployed, you can always use HelmChartConfig definition to tweak the configuration.
  # IMPORTANT: Rancher's install is quite memory intensive, you will require at least 4GB if RAM, meaning cx21 server type (for your control plane).
  # ALSO, in order for Rancher to successfully deploy, you have to set the "rancher_hostname".
  # enable_rancher = true

  # If using Rancher you can set the Rancher hostname, it must be unique hostname even if you do not use it.
  # If not pointing the DNS, you can just port-forward locally via kubectl to get access to the dashboard.
  # If you already set the lb_hostname above and are using a Hetzner LB, you do not need to set this one, as it will be used by default.
  # But if you set this one explicitly, it will have preference over the lb_hostname in rancher settings.
  # rancher_hostname = "rancher.xyz.dev"

  # When Rancher is deployed, by default is uses the "latest" channel. But this can be customized.
  # The allowed values are "stable" or "latest".
  # rancher_install_channel = "stable"

  # Finally, you can specify a bootstrap-password for your rancher instance. Minimum 48 characters long!
  # If you leave empty, one will be generated for you.
  # (Can be used by another rancher2 provider to continue setup of rancher outside this module.)
  # rancher_bootstrap_password = ""

  # Separate from the above Rancher config (only use one or the other). You can import this cluster directly on an
  # an already active Rancher install. By clicking "import cluster" choosing "generic", giving it a name and pasting
  # the cluster registration url below. However, you can also ignore that and apply the url via kubectl as instructed
  # by Rancher in the wizard, and that would register your cluster too.
  # More information about the registration can be found here https://rancher.com/docs/rancher/v2.6/en/cluster-provisioning/registered-clusters/
  # rancher_registration_manifest_url = "https://rancher.xyz.dev/v3/import/xxxxxxxxxxxxxxxxxxYYYYYYYYYYYYYYYYYYYzzzzzzzzzzzzzzzzzzzzz.yaml"

  # Extra values that will be passed to the `extra-manifests/kustomization.yaml.tpl` if its present.
  # extra_kustomize_parameters={}

  # It is best practice to turn this off, but for backwards compatibility it is set to "true" by default.
  # See https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/349
  # When "false". The kubeconfig file can instead be created by executing: "terraform output --raw kubeconfig > cluster_kubeconfig.yaml"
  # Always be careful to not commit this file!
  # create_kubeconfig = false

  ### ADVANCED - Custom helm values for packages above (search _values if you want to located where those are mentioned upper in this file)
  # ⚠️ Inside the _values variable below are examples, up to you to find out the best helm values possible, we do not provide support for customized helm values.
  # Please understand that the indentation is very important, inside the EOTs, as those are proper yaml helm values.
  # We advise you to use the default values, and only change them if you know what you are doing!

  # Cilium, all Cilium helm values can be found at https://github.com/cilium/cilium/blob/master/install/kubernetes/cilium/values.yaml
  # The following is an example, please note that the current indentation inside the EOT is important.
  /*   cilium_values = <<EOT
ipam:
  mode: kubernetes
devices: "eth1"
k8s:
  requireIPv4PodCIDR: true
kubeProxyReplacement: strict
l7Proxy: false
encryption:
  enabled: true
  type: wireguard
  EOT */

  # Cert manager, all cert-manager helm values can be found at https://github.com/cert-manager/cert-manager/blob/master/deploy/charts/cert-manager/values.yaml
  # The following is an example, please note that the current indentation inside the EOT is important.
  /*   cert_manager_values = <<EOT
installCRDs: true
replicaCount: 3
webhook:
  replicaCount: 3
cainjector:
  replicaCount: 3
  EOT */

  # csi-driver-smb, all csi-driver-smb helm values can be found at https://github.com/kubernetes-csi/csi-driver-smb/blob/master/charts/latest/csi-driver-smb/values.yaml
  # The following is an example, please note that the current indentation inside the EOT is important.
  /*   csi_driver_smb_values = <<EOT
controller:
  name: csi-smb-controller
  replicas: 1
  runOnMaster: false
  runOnControlPlane: false
  resources:
    csiProvisioner:
      limits:
        memory: 300Mi
      requests:
        cpu: 10m
        memory: 20Mi
    livenessProbe:
      limits:
        memory: 100Mi
      requests:
        cpu: 10m
        memory: 20Mi
    smb:
      limits:
        memory: 200Mi
      requests:
        cpu: 10m
        memory: 20Mi
  EOT */

  # Longhorn, all Longhorn helm values can be found at https://github.com/longhorn/longhorn/blob/master/chart/values.yaml
  # The following is an example, please note that the current indentation inside the EOT is important.
  /*   longhorn_values = <<EOT
defaultSettings:
  defaultDataPath: /var/longhorn
persistence:
  defaultFsType: ext4
  defaultClassReplicaCount: 3
  defaultClass: true
  EOT */

  # Traefik, all Traefik helm values can be found at https://github.com/traefik/traefik-helm-chart/blob/master/traefik/values.yaml
  # The following is an example, please note that the current indentation inside the EOT is important.
  /*   traefik_values = <<EOT
deployment:
  replicas: 1
globalArguments: []
service:
  enabled: true
  type: LoadBalancer
  annotations:
    "load-balancer.hetzner.cloud/name": "k3s"
    "load-balancer.hetzner.cloud/use-private-ip": "true"
    "load-balancer.hetzner.cloud/disable-private-ingress": "true"
    "load-balancer.hetzner.cloud/location": "nbg1"
    "load-balancer.hetzner.cloud/type": "lb11"

ports:
  web:
    redirectTo: websecure

    proxyProtocol:
      trustedIPs:
        - 127.0.0.1/32
        - 10.0.0.0/8
    forwardedHeaders:
      trustedIPs:
        - 127.0.0.1/32
        - 10.0.0.0/8
  websecure:
    proxyProtocol:
      trustedIPs:
        - 127.0.0.1/32
        - 10.0.0.0/8
    forwardedHeaders:
      trustedIPs:
        - 127.0.0.1/32
        - 10.0.0.0/8
  EOT */

  # Nginx, all Nginx helm values can be found at https://github.com/kubernetes/ingress-nginx/blob/main/charts/ingress-nginx/values.yaml
  # You can also have a look at https://kubernetes.github.io/ingress-nginx/, to understand how it works, and all the options at your disposal.
  # The following is an example, please note that the current indentation inside the EOT is important.
  /*   nginx_values = <<EOT
controller:
  watchIngressWithoutClass: "true"
  kind: "DaemonSet"
  config:
    "use-forwarded-headers": "true"
    "compute-full-forwarded-for": "true"
    "use-proxy-protocol": "true"
  service:
    annotations:
      "load-balancer.hetzner.cloud/name": "k3s"
      "load-balancer.hetzner.cloud/use-private-ip": "true"
      "load-balancer.hetzner.cloud/disable-private-ingress": "true"
      "load-balancer.hetzner.cloud/location": "nbg1"
      "load-balancer.hetzner.cloud/type": "lb11"
  EOT */

  # Rancher, all Rancher helm values can be found at https://rancher.com/docs/rancher/v2.5/en/installation/install-rancher-on-k8s/chart-options/
  # The following is an example, please note that the current indentation inside the EOT is important.
  /*   rancher_values = <<EOT
ingress:
  tls:
    source: "rancher"
hostname: "rancher.example.com"
replicas: 1
bootstrapPassword: "supermario"
  EOT */

}

provider "hcloud" {
  token = var.hcloud_token != "" ? var.hcloud_token : local.hcloud_token
}

terraform {
  required_version = ">= 1.3.3"
  required_providers {
    hcloud = {
      version = ">= 1.38.2"
    }
  }
}

output "kubeconfig" {
  sensitive = true
}

variable "hcloud_token" {
  sensitive = true
  default   = ""
}

variable "etcd_s3_endpoint" {
  sensitive = true
  default   = ""
}

variable "etcd_s3_bucket" {
  sensitive = true
  default   = ""
}

variable "etcd_s3_access_key" {
  sensitive = true
  default   = ""
}

variable "etcd_s3_secret_key" {
  sensitive = true
  default   = ""
}
```

### Screenshots

_No response_

### Platform

Linux

schlichtanders:
fixes #949 by
- ~adding documentation about how to backup~
- making k3s_token manually overwritable

### DIFF
--- a/README.md
+++ b/README.md
@@ -672,11 +672,27 @@ For backup you need to the following steps
 2. [Optional] You can also trigger a manual backup by `ssh` into your first control-plane node, and running `k3s etcd-snapshot`. This will create a manual backup. It is good to do this at least once to see that your configuration is actually working.
 3. Note down the terraform output `k3s_token` - you need it for restoration
 
-For restoration you need to initiate the cluster creation with the mentioned `k3s_token`.
-1. Before cluster creation, set the environment variable `export TF_VAR_k3s_token="..."` to your backedup token.
-2. Start the cluster as usual.
+For restoration you need to initiate the cluster creation with the mentioned `k3s_token`. For instance do the following
+1. Before cluster creation, adapt your `kube.tf` file to include
+    ```tf
+    module "kube-hetzner" {
+      k3s_token = var.k3s_token
+    }
+    ```
+    and
+    ```tf
+    variable "k3s_token" {
+      sensitive = true
+      type      = string
+      default   = null
+    }
+    ```
+    and set the environment variable `export TF_VAR_k3s_token="..."` to your backedup token. (Be careful, this token is like an admin password to the entire cluster.)
+
+2. Create the cluster as usual.
 
     Note, for the backup to work correctly, it is important that your cluster has the same cluster name, but be cautious if you do this for testing purposes to not accidentally messs with your production cluster (e.g. by having the wrong hcloud context activated and running the cleanup script you may be able to accidentally delete everything on production). Be extra cautious, so it may be good advise to first test with another cluster name.
+
 3. login into controlplane node and run the following
     ```bash
     # k3s needs to be stopped before doing the backup restoration
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ To achieve this, we built up on the shoulders of giants by choosing [openSUSE Mi
 
 **Why OpenSUSE MicroOS (and not Ubuntu)?**
 
-- Optimized container OS that is fully locked down, most of the filesystem is read-only! 
+- Optimized container OS that is fully locked down, most of the filesystem is read-only!
 - Hardened by default with an automatic ban for abusive IPs on SSH for instance.
 - Evergreen release, your node will stay valid forever, as it piggy-backs into OpenSUSE Tumbleweed's rolling release!
 - Automatic updates by default and automatic roll-backs if something breaks, thanks to its use of BTRFS snapshots.
@@ -306,7 +306,7 @@ However, some applications that e.g. provide custom CRDs (e.g. [ArgoCD](https://
 </details>
 
 <details>
-  
+
 <summary>Useful Cilium commands</summary>
 
 With Kube-Hetzner, you have the possibility to use Cilium as a CNI. It's very powerful and has great observability features. Below you will find a few useful commands.
@@ -331,7 +331,7 @@ kubectl -n kube-system exec --stdin --tty cilium-xxxx -- cilium service list
 ```
 
 _For more cilium commands, please refer to their corresponding [Documentation](https://docs.cilium.io/en/latest/cheatsheet)._
-  
+
 </details>
 
 <details>
@@ -464,9 +464,9 @@ spec:
             port:
               number: 80
 ```
-  
+
 _⚠️ In case of using Ingress-Nginx as an ingress controller if you choose to use the HTTP challenge method you need to do an additional step of adding variable `lb_hostname = "cluster.example.org"` to your kube.tf. You must set it to an FQDN that points to your LB address._
-  
+
 _This is to circumvent this known issue [cert-manager/cert-manager/issues/466](https://github.com/cert-manager/cert-manager/issues/466). Otherwise, you can just use the DNS challenge, which does not require any additional tweaks to work._
 
 </details>
@@ -489,9 +489,9 @@ To delete a snapshot, first find it with:
 ```bash
 hcloud image list
 ```
-  
+
 Then delete it with:
-  
+
 ```bash
 hcloud image delete <image-id>
 ```
@@ -660,6 +660,24 @@ This can be helpful when you setup a mixed-architecture cluster, and there are m
 
 
 </details>
+
+<details>
+
+<summary>Backup and restore a cluster</summary>
+
+K3s allows for automated etcd backups to S3.
+
+For backup you need to the following steps
+1. fill the kube.tf config `etcd_s3_backup`, it will trigger a regular automated backup to S3
+2. [Optional] You can also trigger a manual backup by `ssh` into your first control-plane node, and running `k3s etcd-snapshot`. This will create a manual backup.
+3. `ssh` into your first control-plane node, `cat /etc/rancher/k3s/config.yaml`, and copy the `token` field. This token is the secret use for encrypting sensible parts in the backup. You won't be able to restore the backup without this key. Store at a safe place, it is like an admin password.
+
+For restoration you need to initiate the cluster creation with the `token` you stored safely in backup step 3.
+1. Before cluster creation, set the environment variable `export TF_VAR_k3s_token="..."` to your backedup token.
+2. start the cluster as usual
+
+</details>
+
 
 ## Debugging
 
--- a/README.md
+++ b/README.md
@@ -661,65 +661,6 @@ This can be helpful when you setup a mixed-architecture cluster, and there are m
 
 </details>
 
-<details>
-
-<summary>Backup and restore a cluster</summary>
-
-K3s allows for automated etcd backups to S3.
-
-For backup you need to the following steps
-1. fill the kube.tf config `etcd_s3_backup`, it will trigger a regular automated backup to S3
-2. [Optional] You can also trigger a manual backup by `ssh` into your first control-plane node, and running `k3s etcd-snapshot`. This will create a manual backup. It is good to do this at least once to see that your configuration is actually working.
-3. Note down the terraform output `k3s_token` - you need it for restoration
-
-For restoration you need to initiate the cluster creation with the mentioned `k3s_token`. For instance do the following
-1. Before cluster creation, adapt your `kube.tf` file to include
-    ```tf
-    module "kube-hetzner" {
-      k3s_token = var.k3s_token
-    }
-    ```
-    and
-    ```tf
-    variable "k3s_token" {
-      sensitive = true
-      type      = string
-      default   = null
-    }
-    ```
-    and set the environment variable `export TF_VAR_k3s_token="..."` to your backedup token. (Be careful, this token is like an admin password to the entire cluster.)
-
-2. Create the cluster as usual.
-
-    Note, for the backup to work correctly, it is important that your cluster has the same cluster name, but be cautious if you do this for testing purposes to not accidentally messs with your production cluster (e.g. by having the wrong hcloud context activated and running the cleanup script you may be able to accidentally delete everything on production). Be extra cautious, so it may be good advise to first test with another cluster name.
-
-3. login into controlplane node and run the following
-    ```bash
-    # k3s needs to be stopped before doing the backup restoration
-    systemctl stop k3s
-
-    # restore a specific backup - you need to fill in all the variables
-    k3s server \
-    --cluster-init \
-    --cluster-reset \
-    --etcd-s3 \
-    --cluster-reset-restore-path=$SNAPSHOT_NAME \
-    --etcd-s3-endpoint=$ETCD_S3_ENDPOINT \
-    --etcd-s3-bucket=$ETCD_S3_BUCKET \
-    --etcd-s3-access-key=$ETCD_S3_ACCESS_KEY \
-    --etcd-s3-secret-key=$ETCD_S3_SECRET_KEY
-
-    # if everything worked out
-    systemctl start k3s
-
-    # note down the changed kubeconfig and adapt your config respectively
-    cat /etc/rancher/k3s/k3s.yaml
-    ```
-
-Awesome! You restored a whole cluster from a backup.
-
-</details>
-
 
 ## Debugging
 
--- a/README.md
+++ b/README.md
@@ -669,12 +669,38 @@ K3s allows for automated etcd backups to S3.
 
 For backup you need to the following steps
 1. fill the kube.tf config `etcd_s3_backup`, it will trigger a regular automated backup to S3
-2. [Optional] You can also trigger a manual backup by `ssh` into your first control-plane node, and running `k3s etcd-snapshot`. This will create a manual backup.
-3. `ssh` into your first control-plane node, `cat /etc/rancher/k3s/config.yaml`, and copy the `token` field. This token is the secret use for encrypting sensible parts in the backup. You won't be able to restore the backup without this key. Store at a safe place, it is like an admin password.
+2. [Optional] You can also trigger a manual backup by `ssh` into your first control-plane node, and running `k3s etcd-snapshot`. This will create a manual backup. It is good to do this at least once to see that your configuration is actually working.
+3. Note down the terraform output `k3s_token` - you need it for restoration
 
-For restoration you need to initiate the cluster creation with the `token` you stored safely in backup step 3.
+For restoration you need to initiate the cluster creation with the mentioned `k3s_token`.
 1. Before cluster creation, set the environment variable `export TF_VAR_k3s_token="..."` to your backedup token.
-2. start the cluster as usual
+2. Start the cluster as usual.
+
+    Note, for the backup to work correctly, it is important that your cluster has the same cluster name, but be cautious if you do this for testing purposes to not accidentally messs with your production cluster (e.g. by having the wrong hcloud context activated and running the cleanup script you may be able to accidentally delete everything on production). Be extra cautious, so it may be good advise to first test with another cluster name.
+3. login into controlplane node and run the following
+    ```bash
+    # k3s needs to be stopped before doing the backup restoration
+    systemctl stop k3s
+
+    # restore a specific backup - you need to fill in all the variables
+    k3s server \
+    --cluster-init \
+    --cluster-reset \
+    --etcd-s3 \
+    --cluster-reset-restore-path=$SNAPSHOT_NAME \
+    --etcd-s3-endpoint=$ETCD_S3_ENDPOINT \
+    --etcd-s3-bucket=$ETCD_S3_BUCKET \
+    --etcd-s3-access-key=$ETCD_S3_ACCESS_KEY \
+    --etcd-s3-secret-key=$ETCD_S3_SECRET_KEY
+
+    # if everything worked out
+    systemctl start k3s
+
+    # note down the changed kubeconfig and adapt your config respectively
+    cat /etc/rancher/k3s/k3s.yaml
+    ```
+
+Awesome! You restored a whole cluster from a backup.
 
 </details>
 
--- a/agents.tf
+++ b/agents.tf
@@ -59,7 +59,7 @@ resource "null_resource" "agents" {
     content = yamlencode({
       node-name     = module.agents[each.key].name
       server        = "https://${var.use_control_plane_lb ? hcloud_load_balancer_network.control_plane.*.ip[0] : module.control_planes[keys(module.control_planes)[0]].private_ipv4_address}:6443"
-      token         = random_password.k3s_token.result
+      token         = local.k3s_token
       kubelet-arg   = local.kubelet_arg
       flannel-iface = local.flannel_iface
       node-ip       = module.agents[each.key].private_ipv4_address
--- a/autoscaler-agents.tf
+++ b/autoscaler-agents.tf
@@ -77,7 +77,7 @@ data "cloudinit_config" "autoscaler-config" {
         sshAuthorizedKeys = concat([var.ssh_public_key], var.ssh_additional_public_keys)
         k3s_config = yamlencode({
           server        = "https://${var.use_control_plane_lb ? hcloud_load_balancer_network.control_plane.*.ip[0] : module.control_planes[keys(module.control_planes)[0]].private_ipv4_address}:6443"
-          token         = random_password.k3s_token.result
+          token         = local.k3s_token
           kubelet-arg   = local.kubelet_arg
           flannel-iface = local.flannel_iface
           node-label    = concat(local.default_agent_labels, var.autoscaler_labels)
--- a/control_planes.tf
+++ b/control_planes.tf
@@ -103,7 +103,7 @@ resource "null_resource" "control_planes" {
             module.control_planes[each.key].private_ipv4_address == module.control_planes[keys(module.control_planes)[0]].private_ipv4_address ?
             module.control_planes[keys(module.control_planes)[1]].private_ipv4_address :
           module.control_planes[keys(module.control_planes)[0]].private_ipv4_address}:6443"
-          token                       = random_password.k3s_token.result
+          token                       = local.k3s_token
           disable-cloud-controller    = true
           disable                     = local.disable_extras
           kubelet-arg                 = local.kubelet_arg
--- a/docs/terraform.md
+++ b/docs/terraform.md
@@ -195,6 +195,8 @@
 | <a name="output_control_planes_public_ipv4"></a> [control\_planes\_public\_ipv4](#output\_control\_planes\_public\_ipv4) | The public IPv4 addresses of the controlplane servers. |
 | <a name="output_ingress_public_ipv4"></a> [ingress\_public\_ipv4](#output\_ingress\_public\_ipv4) | The public IPv4 address of the Hetzner load balancer |
 | <a name="output_ingress_public_ipv6"></a> [ingress\_public\_ipv6](#output\_ingress\_public\_ipv6) | The public IPv6 address of the Hetzner load balancer |
+| <a name="output_k3s_endpoint"></a> [k3s\_endpoint](#output\_k3s\_endpoint) | A controller endpoint to register new nodes |
+| <a name="output_k3s_token"></a> [k3s\_token](#output\_k3s\_token) | The k3s token to register new nodes |
 | <a name="output_kubeconfig"></a> [kubeconfig](#output\_kubeconfig) | Kubeconfig file content with external IP address |
 | <a name="output_kubeconfig_data"></a> [kubeconfig\_data](#output\_kubeconfig\_data) | Structured kubeconfig data to supply to other providers |
 | <a name="output_kubeconfig_file"></a> [kubeconfig\_file](#output\_kubeconfig\_file) | Kubeconfig file content with external IP address |
--- a/init.tf
+++ b/init.tf
@@ -13,7 +13,7 @@ resource "null_resource" "first_control_plane" {
       merge(
         {
           node-name                   = module.control_planes[keys(module.control_planes)[0]].name
-          token                       = random_password.k3s_token.result
+          token                       = local.k3s_token
           cluster-init                = true
           disable-cloud-controller    = true
           disable                     = local.disable_extras
--- a/locals.tf
+++ b/locals.tf
@@ -7,6 +7,9 @@ locals {
   # Otherwise, a new one will be created by the module.
   hcloud_ssh_key_id = var.hcloud_ssh_key_id == null ? hcloud_ssh_key.k3s[0].id : var.hcloud_ssh_key_id
 
+  # if given as a variable, we want to use the given token. This is needed to restore the cluster
+  k3s_token = var.k3s_token == null ? random_password.k3s_token.result : var.k3s_token
+
   ccm_version    = var.hetzner_ccm_version != null ? var.hetzner_ccm_version : data.github_release.hetzner_ccm[0].release_tag
   csi_version    = length(data.github_release.hetzner_csi) == 0 ? var.hetzner_csi_version : data.github_release.hetzner_csi[0].release_tag
   kured_version  = var.kured_version != null ? var.kured_version : data.github_release.kured[0].release_tag
@@ -609,10 +612,10 @@ EOF
     set -euo pipefail
 
     sleep 11
-    
+
     INTERFACE=$(ip link show | awk '/^3:/{print $2}' | sed 's/://g')
     MAC=$(cat /sys/class/net/$INTERFACE/address)
-    
+
     cat <<EOF > /etc/udev/rules.d/70-persistent-net.rules
     SUBSYSTEM=="net", ACTION=="add", DRIVERS=="?*", ATTR{address}=="$MAC", NAME="eth1"
     EOF
--- a/output.tf
+++ b/output.tf
@@ -37,6 +37,14 @@ output "ingress_public_ipv6" {
   value       = (local.has_external_load_balancer || var.load_balancer_disable_ipv6) ? null : data.hcloud_load_balancer.cluster[0].ipv6
 }
 
+output "k3s_endpoint" {
+  value = "https://${var.use_control_plane_lb ? hcloud_load_balancer_network.control_plane.*.ip[0] : module.control_planes[keys(module.control_planes)[0]].private_ipv4_address}:6443"
+}
+
+output "k3s_token" {
+  value = random_password.k3s_token.result
+}
+
 # Keeping for backward compatibility
 output "kubeconfig_file" {
   value       = local.kubeconfig_external
--- a/output.tf
+++ b/output.tf
@@ -38,11 +38,13 @@ output "ingress_public_ipv6" {
 }
 
 output "k3s_endpoint" {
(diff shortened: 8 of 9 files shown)


schlichtanders:
I tried this now by changing source to this pullrequest, however it seems the original master branch this was forked from is buggy...

I am getting infinite
```
module.kube-hetzner.null_resource.kustomization: Still creating... [50s elapsed]
module.kube-hetzner.null_resource.kustomization: Still creating... [1m0s elapsed]
module.kube-hetzner.null_resource.kustomization: Still creating... [1m10s elapsed]
module.kube-hetzner.null_resource.kustomization: Still creating... [1m20s elapsed]
module.kube-hetzner.null_resource.kustomization: Still creating... [1m30s elapsed]
module.kube-hetzner.null_resource.kustomization: Still creating... [1m40s elapsed]
module.kube-hetzner.null_resource.kustomization: Still creating... [1m50s elapsed]
module.kube-hetzner.null_resource.kustomization: Still creating... [2m0s elapsed]
module.kube-hetzner.null_resource.kustomization: Still creating... [2m10s elapsed]
module.kube-hetzner.null_resource.kustomization: Still creating... [2m20s elapsed]
module.kube-hetzner.null_resource.kustomization: Still creating... [2m30s elapsed]
module.kube-hetzner.null_resource.kustomization: Still creating... [2m40s elapsed]
```
(happens also when pointing source to kube-hetzner's master branch)

@mysticaltech I am bit surprised that the master branch seems dysfunctional - what is the recommended approach for opening and testing pullrequests if not by forking from master?

mysticaltech:
@schlichtanders There shouldn't be any issues with master or staging, and both should be in sync. If you are getting an issue, maybe it's the new code, or it's something with Hetzner at the moment. I will give it a try myself too later on when this is merged and before getting released. However, the code looks good. Keep in mind though that the k3s_token probably must be alphanumeric and of a specific length.

mysticaltech:
Basically, if we can keep the token autogenerated, it would be great. So maybe just add the example, up to you, don't hesitate to push back at my suggestions ;)

schlichtanders:
> Basically, if we can keep the token autogenerated, it would be great. So maybe just add the example, up to you, don't hesitate to push back at my suggestions ;)

I tested setting k3s_token and could successfully restore configmaps and secrets from my backup 
(there where still some other issues with the restore, but I think they happened because I choose another cluster name for test purposes, which results in other node names and hence some backuped configuration cannot be applied to the new test cluster)

So k3s_token is really needed: Being able to manually set k3s_token on cluster construction is necessary to restore a failed cluster from scratch.

> I just merged #946 in staging,
Thank you for the pointer to the other pullrequest. Actually sounds like a perfect addition - I will merge it today soon.

schlichtanders:
> Basically, if we can keep the token autogenerated, it would be great. So maybe just add the example, up to you, don't hesitate to push back at my suggestions ;)

I tested setting k3s_token and could successfully restore configmaps and secrets from my backup 
(there where still some other issues with the restore, but I think they happened because I choose another cluster name for test purposes, which results in other node names and hence some backuped configuration cannot be applied to the new test cluster)

So k3s_token is really needed: Being able to manually set k3s_token on cluster construction is necessary to restore a failed cluster from scratch.

> I just merged #946 in staging,
Thank you for the pointer to the other pullrequest. Actually sounds like a perfect addition - I will merge it today soon.

schlichtanders:
It is always confusing me if someone else adapts my pullrequest branch.

Hence I now rebased multiple times to staging, first locally and then again when rebasing the upstream pullrequest branch of mine.

I don't know why github mentions there is conflict to be resolved - it should all be nicely rebased.

schlichtanders:
I finally took out the documentation because it is still not working for me to truly restore a cluster (partially it works, which confirms that this pullrequest is needed).

From my side this can be merged.

mysticaltech:
@schlichtanders Wonderful, thanks for pushing the envelope on this one, especially in the light of the other complementary PR. Will merge by the end of the day.

PR #958, opened 2023-09-04
waiting for the completion of the cloudinit's work

metanovii:
You should wait for the cloudinit's work to be completed  to avoid collisions, locks and other things.


mysticaltech (approved):
That makes sense! Thanks for that. @kube-hetzner/core FYI.

aleksasiriski (approved):

PR #967, opened 2023-09-07
adding postinstall_exec to hook restore commands

schlichtanders:
After two intensive weeks with all kinds of trial and errors with cluster restoration from etcd backup, actually only this `postinstall_exec` hook is needed to make it work.

EDIT: I am surprised by all the changes which are not from myself... I try to make a cleaner rebase now...


schlichtanders:
closing this convoluted pullrequest in favor of the one which rebases onto master (which somehow has a much cleaner diff)

PR #968, opened 2023-09-07
adding postinstall_exec to hook restore commands

schlichtanders:
After two intensive weeks with all kinds of trial and errors with cluster restoration from etcd backup, actually only this postinstall_exec hook is needed to make it work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -662,6 +662,128 @@ This can be helpful when you setup a mixed-architecture cluster, and there are m
 </details>
 
 
+
+
+<details>
+
+<summary>Backup and restore a cluster</summary>
+
+K3s allows for automated etcd backups to S3. Etcd is the default storage backend on kube-hetzner, even for a single control plane cluster, hence this should work for all cluster deployments.
+
+**For backup do:**
+1. Fill the kube.tf config `etcd_s3_backup`, it will trigger a regular automated backup to S3.
+2. Add the k3s_token as an output to your kube.tf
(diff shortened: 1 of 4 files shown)


mysticaltech (approved):
Looking good! An example of how to use it, with or without the greatest context of restauration would be awesome. We have the examples section in the readme and/or the examples folder, for such demos.

schlichtanders:
good idea! I have a longer train ride today and will add in then.

schlichtanders:
I added a documentation about how to do the restoration

unfortunately, I needed to revert one change recently done, because it broke my cluster. See https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/970

schlichtanders:
according to this https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/970#issuecomment-1710221565 it seems indeed best to comment the cloud-init command as currently done in this pullrequest

mysticaltech:
@schlichtanders  This line was not needed, I don't know why I accepted the PR in the first place, it looked like a good idea, but we already do implicit checks. So will merge this now and remove it completely. And wonderful job on the restore example, love it! 🥳 🙏

mysticaltech (approved):
Let's go  🚀

mysticaltech:
Awesome, thanks @schlichtanders. Keep up the good work, amazing contributions to this project.

PR #975, opened 2023-09-11
Change default DNS servers

jidckii:
DNS 9.9.9.9 applies filtering based on the VirusTotal database. My domain was included in this database by mistake and I spent several days debugging to understand what was wrong.

After updating the cluster and the release of the 9.9.9.9 resolver, my sites in the yucca.app domain began to sometimes return http 500, this is because nginx ingress cannot resolve the request on the incoming domain.

I suggest not using the default DNS which uses any filtering and can lead to service failure. The use of such resolvers should be a conscious choice. On its website, quad9 suggests using the address without filtering, so I replaced it in the default variable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -866,6 +866,8 @@ Code contributions are very much **welcome**.
     packer build ../kube-hetzner/packer-template/hcloud-microos-snapshots.pkr.hcl
     ```
 
+1. Update examples in `kube.tf.example` if required.
+1. Run [terraform-docs](https://github.com/terraform-docs/terraform-docs) generator `docker run --rm --volume "$(pwd):/terraform-docs" -u $(id -u) quay.io/terraform-docs/terraform-docs:0.16.0 markdown /terraform-docs > docs/terraform.md` if you changed default variables.
 1. Commit your Changes (`git commit -m 'Add some AmazingFeature')
 1. Push to the Branch (`git push origin AmazingFeature`)
 1. Open a Pull Request targeting the `staging` branch.
--- a/README.md
+++ b/README.md
@@ -867,7 +867,6 @@ Code contributions are very much **welcome**.
     ```
 
 1. Update examples in `kube.tf.example` if required.
-1. Run [terraform-docs](https://github.com/terraform-docs/terraform-docs) generator `docker run --rm --volume "$(pwd):/terraform-docs" -u $(id -u) quay.io/terraform-docs/terraform-docs:0.16.0 markdown /terraform-docs > docs/terraform.md` if you changed default variables.
 1. Commit your Changes (`git commit -m 'Add some AmazingFeature')
 1. Push to the Branch (`git push origin AmazingFeature`)
 1. Open a Pull Request targeting the `staging` branch.
--- a/variables.tf
+++ b/variables.tf
@@ -679,7 +679,7 @@ variable "control_plane_lb_enable_public_interface" {
 
 variable "dns_servers" {
   type        = list(string)
-  default     = ["1.1.1.1", "8.8.8.8", "9.9.9.9"]
+  default     = ["1.1.1.1", "8.8.8.8", "9.9.9.10"]
   description = "IP Addresses to use for the DNS Servers, set to an empty list to use the ones provided by Hetzner. The length is limited to 3 entries, more entries is not supported by kubernetes"
 
   validation {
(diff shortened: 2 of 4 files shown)


mysticaltech (approved):
Yep, makes sense!

M4t7e:
Just my 5 cents:

I was often thinking about the default DNS settings and still I do. Technically all 3 are good and fast, but they have some hidden drawbacks...

- **Cloudflare DNS:**
  They openly say that they observe, log and monitor all DNS requests made using their standard DNS resolvers with some level of anonymization. https://www.cloudflare.com/privacypolicy/
- **Google DNS:**
  They also observe, log and monitor all DNS requests like Cloudflare. https://developers.google.com/speed/public-dns/faq#privacy
  Additionally they send a so called EDNS-Client-Subnet (ECS) along with the DNS query to the authoritative servers, which simply contains a part of your source IP address.
  Example:
  ```bash
  # dig txt ecs.test.smartguard.io @8.8.8.8 +short
  "128.140.118.0/24"
  # dig txt ecs.test.smartguard.io @2001:4860:4860::8888 +short
  "2a01:4f8:c17:cd00::/56"
  
  # dig txt ecs.test.smartguard.io @1.1.1.1 +short
  "no ECS"
  # dig txt ecs.test.smartguard.io @9.9.9.9 +short
  "no ECS"
  ```
- **Quad9:**
  The only privacy friendly DNS resolver service and not located in the US (nothing against the US in general, but their laws explicitly ignore EU law, as Edward taught us). But they don't serve all DNS answers uncensored anymore. There is a lawsuit ongoing from Sony Music Entertainment. They claim that Quad9 actively supports people to download music illegally (what a bulls**t). See here for more info: https://www.quad9.net/news/blog/quad9-and-sony-music-german-injunction-update-for-july-2023

TL;DR
I think the Hetzner DNS resolvers may be a better default from a privacy and consistency perspective. Especially because the mixture between Cloudflare, Google and Quad9 can lead to toggle effects in case of some issues like this here. First something works, then it does not work, testing manually and it suddenly works again, and so on...

jidckii:
I agree with @M4t7e , maybe it makes sense to use resolvers from the hoster?

mysticaltech:
Sounds good folks, @jidckii let's default to [Hetzner DNS servers](https://docs.hetzner.com/dns-console/dns/general/recursive-name-servers/) then, and cite the current one in the kube.tf.example line to be uncommented if needed.

mysticaltech (commented):
@jidckii Just two small comments before we move on.

README.md
```diff
@@ -866,6 +866,8 @@ Code contributions are very much **welcome**.
     packer build ../kube-hetzner/packer-template/hcloud-microos-snapshots.pkr.hcl
     ```
 
+1. Update examples in `kube.tf.example` if required.
+1. Run [terraform-docs](https://github.com/terraform-docs/terraform-docs) generator `docker run --rm --volume "$(pwd):/terraform-docs" -u $(id -u) quay.io/terraform-docs/terraform-docs:0.16.0 markdown /terraform-docs > docs/terraform.md` if you changed default variables.
```
mysticaltech, line 870: @jidckii For line 870, it's not needed, because we run it automatically when merged to staging. But good initiative, appreciate it.

variables.tf
```diff
@@ -679,7 +679,7 @@ variable "control_plane_lb_enable_public_interface" {
 
 variable "dns_servers" {
   type        = list(string)
-  default     = ["1.1.1.1", "8.8.8.8", "9.9.9.9"]
+  default     = []
```
mysticaltech, line 682: No need to hardcode the hetzner ones? It gets picked up correctly from the OS, are you sure?

jidckii: No I'm not sure, I followed the comment for this variable.
But I can check. Need to check?

mysticaltech (approved):

variables.tf
jidckii: @mysticaltech? its ok
```
test-kube-hetzner-control-plane-fsn1-wdg:/ # cat /etc/resolv.conf 
# Generated by NetworkManager
nameserver 185.12.64.2
nameserver 185.12.64.1
nameserver 2a01:4ff:ff00::add:2
# NOTE: the libc resolver may not support more than 3 nameservers.
# The nameservers listed below may not be recognized.
nameserver 2a01:4ff:ff00::add:1
```

mysticaltech: Thanks for checking @jidckii, really appreciate it! Will merge and deploy tonight.

mysticaltech (approved):

mysticaltech:
Now live on v2.7.1. thanks again for this @jidckii, and for the great suggestion @M4t7e 👏

PR #984, opened 2023-09-17
do not prompt for hcloud token if it is already in env

kindlychung:


mysticaltech (approved):
@kindlychung Yes, that was needed. Thank you!

PR #985, opened 2023-09-17
Expose ingress_controller variable in example

ricristian:
There are cases where we need to disable load balancer completely. Even if the option is available in tf code it is not exposed In kube.tf example

### DIFF
--- a/kube.tf.example
+++ b/kube.tf.example
@@ -225,12 +225,15 @@ module "kube-hetzner" {
   load_balancer_type     = "lb11"
   load_balancer_location = "fsn1"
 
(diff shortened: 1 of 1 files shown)


M4t7e:
Hey @ricristian, the `ingress_controller` variable is already available in kube.tf.example. Please have a look here:
https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/1f44821d9b17036d5dff088d135063c042af0db7/kube.tf.example#L374-L379

mysticaltech:
@ricristian Indeed, @M4t7e is right. But thanks for that anyways.

PR #993, opened 2023-09-23
Update terraform provider versions

RelativeSure:


RelativeSure:
image

mysticaltech:
@RelativeSure Thanks for this, super important indeed. Please could you also check kube.tf.example, I believe it has some versions to upgrade in there too.

mysticaltech (approved):
LGTM

aleksasiriski (approved):

aleksasiriski (changes requested):
Please remove your terraform lock state

mysticaltech:
@RelativeSure , about the terraform lock file, @aleksasiriski has a point, I understand the recommendation from terraform, however this does not apply to that repo that is just a module. In the place where you use the kube-hetzner module, you can commit that state indeed.

aleksasiriski (approved):

PR #995, opened 2023-09-26
Support K8s v1.28 version

clementguillot:
Hi guys,

Very simple PR: now that K3s 1.28.2 is available, fixing a critical bug with init containers, we can safely use 1.28 as supported (and default) version.

Works like a charm on my staging cluster:

image

Please tell me if I missed something ;)


mysticaltech (approved):
@clementguillot Thanks for this! LGTM. Will merge ASAP.

valkenburg-prevue-ch:
Looking good indeed. One question: would you agree that the defaults should be 1.28.2, instead of 1.28.0? Because of the regression that was there in 1.28.0 and 1.28.1.

M4t7e:
Before upgrading to the next Kubernetes version, it's crucial to verify the compatibility of all third-party components. As noted in https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/974, Cilium is not yet compatible with version 1.28. For more details, refer to Cilium's [compatibility documentation](https://docs.cilium.io/en/stable/network/kubernetes/compatibility/). The same applies to Calico, as outlined in their [Kubernetes requirements documentation](https://docs.tigera.io/calico/latest/getting-started/kubernetes/requirements#kubernetes-requirements). It's possible that other components may also not be ready for the upgrade.


Given that third-party components often require updates to align with new Kubernetes releases, I think a checklist could help us? Then we can systematically verify the compatibility of each component. @mysticaltech, what are your thoughts on this approach?

M4t7e:
I tried to collect the relevant info. The following table outlines the minimum required versions of components to support the corresponding Kubernetes release. Component versions marked in brackets have not yet been released.

| Kubernetes  | Autoscaler | Calico   | Cilium   | HCCM | HCSI | Kured |
|-------------|------------|----------|----------|------|------|-------|
| 1.26        | 1.26       | 3.25     | 1.13     | 1.14 | 2.2  | 1.12  |
| 1.27        | 1.27       | 3.26     | 1.14     | 1.16 | 2.4  | 1.13  |
| 1.28        | 1.28       | _(3.27)_ | _(1.15)_ | 1.18 | 2.5  | 1.14  |

As of now, Calico and Cilium have not released versions compatible with Kubernetes 1.28.

mysticaltech:
@M4t7e  Fully agree with your proposed approach, it's way more rigourous and will ensure our software works with all the components we offer. Thanks for the table, it would be nice to have it in the readme somewhere or any other part.

So @clementguillot, no need to close this PR, let's wait until both Cilium and Calico support 1.28 and then we can update these versions and merge.

mysticaltech:
We had to reset the old staging and some changes were already merged in here. @clementguillot closing for now, but I will move the versions as soon as available. I just check today and calico and cilium are still not ready for v1.28. hopefully in the near future. Thanks again for this!

PR #1000, opened 2023-09-29
[Feature] Add option to use an existing private network.

valkenburg-prevue-ch:
This PR adds two variables for passing an existing private network to the kube-hetzner module. Changes are minimal. Motivation for this is addition of other non-k8s machines in the private network prior to setting up the cluster. One reason is for example the use of a proxy (using HTTP_PROXY and HTTPS_PROXY environment variables), which allows you to reserve a single ip address for outgoing requests of the entire cluster.


kube.tf.example
```diff
@@ -57,6 +57,11 @@ module "kube-hetzner" {
   # * For Hetzner locations see https://docs.hetzner.com/general/others/data-centers-and-connection/
   network_region = "eu-central" # change to `us-east` if location is ash
 
+  # If you want to create the private network before calling this module,
+  # you can do so and pass its id here. NOTE: make sure to adapt network_ipv4_cidr accordingly
```
mysticaltech, line 61: @valkenburg-prevue-ch  Having just the second variable to take in the id would be enough, and with a defautl value of null you can easily test if available or not hence giving you you the needed boolean functionality needed for the network creation phase.

main.tf
```diff
@@ -23,16 +23,21 @@ resource "hcloud_ssh_key" "k3s" {
 }
 
 resource "hcloud_network" "k3s" {
+  count    = var.use_existing_network ? 0 : 1
```
mysticaltech, line 26: Same as above

main.tf
```diff
@@ -23,16 +23,21 @@ resource "hcloud_ssh_key" "k3s" {
...
   name     = var.cluster_name
   ip_range = var.network_ipv4_cidr
   labels   = local.labels
 }
 
+data "hcloud_network" "k3s" {
+  id = var.use_existing_network ? var.existing_network_id : hcloud_network.k3s[0].id
```
mysticaltech, line 33: Same as above

mysticaltech (changes requested):
@valkenburg-prevue-ch  This is great, however, please see my comments, probabably not needed to have the boolean variable.

Also, the way we use the private network is very specific, maybe provide guidance or refer to the right part in the readme, at the level where you collect the network id in kube.tf.example, so that people know what kind of ip ranges need to be free when using this feature. And naturally make sure they understand that this is for exceptional use cases, for advanced users only.

variables.tf
```diff
@@ -74,7 +74,20 @@ variable "network_region" {
   type        = string
   default     = "eu-central"
 }
-
+variable "use_existing_network" {
+  # Unfortunately, we need this extra boolean variable. If we only check that existing_network_id is null,
```
valkenburg-prevue-ch, line 78: @mysticaltech See this comment: initially I had no boolean variable, but terraform forced me to! The problem is that if the id is output of another resource in the same project, terraform cannot be sure if it will be null or not, and hence cannot decide when the count values will be before creating any resource.

If you have literally this:
```

resource "hcloud_network" "k3s_proxied" {
  name     = "k3s-proxied"
  ip_range = "10.0.0.0/8"
}
module "kube_hetzner" {
  existing_network_id  = hcloud_network.k3s_proxied.id
 ...
}
``` 

then the line
```
resource "hcloud_network" "k3s" {
  count    = var.existing_network_id == null ? 0 : 1
  ...
}
```
will raise an error, because apparently hcloud_network.k3s_proxied.id cannot be guaranteed to be non-null.

Do you know of a workaround? I would love to avoid the extra boolean!

valkenburg-prevue-ch:
> Also, the way we use the private network is very specific, maybe provide guidance or refer to the right part in the readme, at the level where you collect the network id in kube.tf.example, so that people know what kind of ip ranges need to be free when using this feature. And naturally make sure they understand that this is for exceptional use cases, for advanced users only.

Thanks for reviewing! And yes, I will adapt the PR with more and better guidance in the docs.

Perhaps you can guess where this is going: I've been setting up a proxy server, which with this PR can be part of the private network. It starts to work flawlessly, so all my egress goes through one pre-reserved ipv4 address which is not blacklisted anywhere. I'll probably share that project soon. It's an http-proxy and a dns-proxy. I'm wondering if it makes sense to make it a NAT-router as well, so one can finally create a private-ip only network (for which a use of bastion-host parameters in all the ssh connections is needed, that's a simple PR which I'm however postponing until I find that it makes sense).

valkenburg-prevue-ch:
Ok, found a workaround. Added documentation.

How about setting the default cidr to 10.0.0.0/9 ? There's plenty of room in that range. All ips are already correctly computed using the cidr functions of tf, and the default service and cluster ranges are inside 10.0.0.0/9 already. This would mean that any user at any time can add stuff to the private network in the 10.128.0.0/9 range, without ever having to dig into the cidr settings.

variables.tf
mysticaltech: @valkenburg-prevue-ch Ok, thanks for clarifying.

mysticaltech (approved):
@valkenburg-prevue-ch Wonderful job, this is super valuable.

mysticaltech:
@valkenburg-prevue-ch LGTM, please just make sure the validate terraform (probably just formatting) passes and we're good to go! 

@M4t7e  @aleksasiriski  I think you will like that one.

kube.tf.example
```diff
@@ -57,6 +57,18 @@ module "kube-hetzner" {
...
+  # you can do so and pass its id here. For example if you want to use a proxy
+  # which only listens on your private network. Advanced use case.
+  # 
+  # NOTE1: make sure to adapt network_ipv4_cidr, cluster_ipv4_cidr, and service_ipv4_cidr accordingly.
+  #        If your network is created with 10.0.0.0/8, and you use subnet 10.128.0.0/9 for your
+  #        non-k3s business, then adapting `network_ipv4_cidr = "10.0.0.0/9"` should be all you need.
+  #
+  # NOTE2: square brackets! This must be a list of length 1.
+  #
+  # existing_network_id = [hcloud_network.your_network.id]
```
M4t7e, line 70: I don't understand this here completely. Below you mentioned, that list-based workaround for `var.existing_network_id` is because:
> Unfortunately, we need this to be a list or null. If we only use a plain
   string here, and check that existing_network_id is null, terraform will
  complain that it cannot set `count` variables based on existing_network_id
  != null, because that id is an output value from
  hcloud_network.your_network.id, which terraform will only know after its
  construction.

Why you would use `hcloud_network.your_network.id` as input for `existing_network_id`? I thought the point here is to let the user provide the existing network id, no? I found no other place where `hcloud_network.your_network.id` is used as input for it, so I guess this here is the place you mentioned in your comment?

valkenburg-prevue-ch:
Weird, Validate Terraform crashes, and its logs say "this is indicative of a bug in terraform". Not sure what I can do about it.

variables.tf
M4t7e: @valkenburg-prevue-ch I think if you make `existing_network_id` a non-nullable variable this should work. Your problem might be caused by a strange bug in Terraform. See: https://github.com/hashicorp/terraform/issues/26755

A solution could be to treat an empty string ("") as null equivalent and set "" as the default, at least until Terraform can handle null values for count parameters. Then you can get rid of the extra `use_existing_network` variable.

valkenburg-prevue-ch: That's exactly the issue, indeed, thanks for finding that discussion. I already got rid of use_existing_network by accepting only null or a list, see the commits below (from yesterday, at the time of writing). I resolve this particular conversation.

valkenburg-prevue-ch: Ah, by the way, any default value is not going to fix it, because the count condition would still depend on a computed value rather than on static information. I tried that. :)

kube.tf.example
M4t7e: Okay, now I got it. But your example from README should then be `existing_network_id = [hcloud_network.k3s_proxied.id]` and not `existing_network_id = [hcloud_network.your_network.id]`, am I right? Or the other way around, all `k3s_proxied` should be `your_network`.

See my comment above regarding the null value. Would that solve the extra bool variable and maybe also the need for this list workaround? It appears to have the same root cause.

valkenburg-prevue-ch: Ha, great catch! Thanks, will commit, it should indeed be k3s_proxied in the README. (but feel free to suggest a better name for the network, if it helps explain it better).

The root cause is that terraform cannot distinguish between (1) the situation where the user supplied a dynamic value which could an empty string or a non-empty string and (2) the user did not supply a value and the default value used. Because that is fundamentally the check that we need here. The only hack I could come up with, is instead of checking the value, checking the type (kind of), which is something terraform will know from the start before any actions are taken. A list vs null (but probably other types could work as well).

valkenburg-prevue-ch:
Terraform is broken at the moment. Forgive me the naughtiness, but I retriggered an older validation which had succeeded, and it crashes too:
https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/actions/runs/6363783912

Must have been a regression in the docker image used for the validation.

kube.tf.example
valkenburg-prevue-ch: I fooled myself: the workaround is not a workaround, it has the same problem, but now I improved on that. Default is an empty list, which *does* play nice with count.

mysticaltech (approved):

kube.tf.example
mysticaltech: @valkenburg-prevue-ch Maybe that can inspire you a little bit (straight from gpt-4):

The conversation seems to be about managing infrastructure as code with Terraform, particularly dealing with the nuances of Terraform's state management and dependency resolution.

### Context

1. **`existing_network_id` Parameter**: This is supposed to accept an ID for an already-existing network, or be left null to create a new one.
2. **Dependency Issue**: The `hcloud_network.your_network.id` is an output value that gets determined only after the resource is created. Terraform, when resolving the dependency graph, doesn't like conditional count variables that depend on such dynamic output values.
3. **List Hack**: To resolve this, a list was used. But as discussed, it also faces the same issue.
4. **Type-check Hack**: They considered checking the type instead of the value to resolve the issue.

### The Core Issue

At its core, the problem is state and dependency management. In Terraform, state is deterministic. It uses the state to figure out what it needs to do. If you give it an ID that will only exist in the future, you disrupt this determinism.

### Possible Solutions

1. **Use `depends_on` or Explicit Dependencies**: You could explicitly tell Terraform that one resource depends on the other. However, this still may not resolve the conditional count issue.

2. **Terraform `locals`**: Using local values, you could write some conditional logic to figure out what value should be passed to the `existing_network_id`. If it is null, then the local block would return a particular value (like 'create_new') which can then be checked within the resource blocks.

    ```hcl
    locals {
      existing_network_id_value = var.existing_network_id != null ? var.existing_network_id : "create_new"
    }
    ```

3. **Conditional Resource Creation**: You could opt for creating resources conditionally based on the value of `existing_network_id`.

    ```hcl
    resource "hcloud_network" "example" {
      count = var.existing_network_id != null ? 0 : 1
      // resource definition
    }
    ```

4. **Null Checks**: If you are open to rewriting parts of the module, you could utilize `null` values more effectively, creating or using resources based on whether the values are null or not.

5. **Two Separate Modules**: You could separate out the logic into two different Terraform modules. One would be used if `existing_network_id` is not null, and the other one would be used otherwise.

6. **Dynamic Blocks**: Terraform 0.12 introduced dynamic nested blocks that can assist in writing conditional logic inside your resources, which can make this kind of dependency management a bit easier.

7. **Solve it Outside Terraform**: Create a shell script, Python script, or another Terraform root module that runs two different configurations based on whether `existing_network_id` is null or not.

Given that Terraform is very declarative and is not well-suited to handle complex conditional logic, some of these solutions might require a bit of "hacky" code. It is a limitation of the tool at its current stage.

mysticaltech: My intuition is that the list hack is maybe causing the terraform validation crash and there must be a better way. I also like solution # 2 and # 3 above, that was my initial thought too, but maybe am not grasping the full issue yet.

mysticaltech: @valkenburg-prevue-ch About terraform crashing, it's generalized not just this PR, if it continues, will just remove the check for now.

So if there are no other optimization, and you are satisfied, we can merge. However, if you see a way to avoid the list hack, it would be great. Thanks again for this great feature.

valkenburg-prevue-ch:
Yes, will do. Maybe remove it from staging until then?

-------- Oorspronkelijk bericht --------
Op 11 okt. 2023 17:49, schreef Karim Naufal :

mysticaltech:
Will do.

PR #1002, opened 2023-09-30
set k3s_token output to sensitive

netzding:
To reduce the risk of accidentally exporting sensitive data that was intended to be only internal, Terraform requires that any root module output containing sensitive data be explicitly marked as sensitive, to confirm your intent.

If you do intend to export this data, annotate the output value as sensitive by adding the following argument:
    sensitive = true

### DIFF
--- a/output.tf
+++ b/output.tf
@@ -45,6 +45,7 @@ output "k3s_endpoint" {
 output "k3s_token" {
   description = "The k3s token to register new nodes"
   value       = local.k3s_token
+  sensitive   = true
 }
 
 # Keeping for backward compatibility


ifeulner:
> @aleksasiriski @ifeulner @M4t7e What do you folks think of this? I guess we had it sensitive before but somehow this was changed back to non sensitive.

There was some integration issue, yes. I think I marked it once as sensitive, but then some things break.

netzding:
The reason why I added this is because internally random_password.k3s_token.result is already considered sensitive and therefore in the current state you cannot use the module as root module (I use terragrunt to compose several terraform modules to my infrastructure). 
Terraform rightfully complains that a sensitive attribute is being passed as an non sensitive output.

As an alternative if you want to keep the k3s_token non-sensitive you should do so explicitly by wrapping it with the nonsensitive function:
`value       = nonsensitive(local.k3s_token)`

Right now as a workaround I use the terragrunt overwrite function to overwrite the outputs.tf (For anyone who is facing the same issue)

mysticaltech:
Thanks for clarifying @ifeulner, I wasn't clear on the matter myself. @netzding If you could create this `output_k3s_token_sensitive` flag, and make it use the nonsensitive function if false, it would be great. And by default let it be false.

mysticaltech:
@netzding Thanks for hanging in there. We will merge this PR to mark it as sensitive and bump the version and announce potentially breaking changes on that. Hopefully it will not affect many people. This seems like the cleanest way to proceed.

mysticaltech (approved):

Issue #539, opened 2023-01-26
Cant get IP from LB, timeout


bulnv:
Hey! I am trying to spin up new cluster on 1.8.2 module but facing this error

```module.kube-hetzner.null_resource.kustomization (remote-exec): Waiting for load-balancer to get an IP...
module.kube-hetzner.null_resource.kustomization (remote-exec): Waiting for load-balancer to get an IP...
module.kube-hetzner.null_resource.kustomization (remote-exec): Waiting for load-balancer to get an IP...

│ Error: remote-exec provisioner error
│ 
│   with module.kube-hetzner.null_resource.kustomization,
│   on .terraform/modules/kube-hetzner/init.tf line 246, in resource "null_resource" "kustomization":
│  246:   provisioner "remote-exec" {
│ 
│ error executing "/tmp/terraform_1613767178.sh": Process exited with status 124
╵
╷
│ Error: attach server to network: provided IP is not available (ip_not_available)
│ 
│   with module.kube-hetzner.module.agents["1-0-agent-large"].hcloud_server_network.server,
│   on .terraform/modules/kube-hetzner/modules/host/main.tf line 214, in resource "hcloud_server_network" "server":
│  214: resource "hcloud_server_network" "server" {```

my config is pretty simple

`
module "kube-hetzner" {
  providers = {
    hcloud = hcloud
  }
  hcloud_token = local.hcloud_token
  source = "kube-hetzner/kube-hetzner/hcloud"
  version = "1.8.2"
  ssh_public_key  = file("./id_ed25519_k8s.pub")
  ssh_private_key = file("./id_ed25519_k8s")
  network_region  = "eu-central" # change to us-east if location is ash
  control_plane_nodepools = local.env == "staging" ? local.staging_control_plane_nodepools : local.production_control_plane_nodepools
  agent_nodepools = local.env == "staging" ? local.staging_agent_nodepools : local.production_agent_nodepools
  load_balancer_type     = "lb11"
  load_balancer_location = "fsn1"
  autoscaler_nodepools = [
    {
      name        = "autoscaler"
      server_type = "cpx21" # must be same or better than the control_plane server type (regarding disk size)!
      location    = "fsn1"
      min_nodes   = 0
      max_nodes   = 5
    }
  ]
  etcd_s3_backup = {
    etcd-s3-endpoint   = "***"
    etcd-s3-access-key = "k8s-backups"
    etcd-s3-secret-key = "***"
[...]

PR #1005, opened 2023-10-03
Add validation for uniqueness of nodepool-names

mrclrchtr:
As discussed here: https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/539

I think it makes sense to do a check on uniqueness of names.

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -153,6 +153,16 @@ variable "control_plane_nodepools" {
     count       = number
   }))
   default = []
+  validation {
+    condition = length(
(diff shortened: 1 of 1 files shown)


mysticaltech (approved):
That is super useful. You tried it right? Just confirm me that it works and we can merge. Thanks.

mrclrchtr:
Yes, I tried it local with my project.

mrclrchtr:
Fixed the tf fmt errors

PR #1009, opened 2023-10-05
Remove warnings about missing EU arm server types from kube.tf.example

maggie44:
All three EU locations are now supported.


mysticaltech (approved):
Wonderful, will merge ASAP.

PR #1010, opened 2023-10-08
Add the possibility to export used values.yaml-files as local files.

mrclrchtr:
Add the possibility to export the values.yaml files used for the deployment of traefik, longhorn, cert-manager, etc.
 
This can be helpful to use them for later deployments like with ArgoCD.


mysticaltech (approved):
LGTM, could be super useful.

mrclrchtr:
I don't know what the "TERRAFORM CRASH" is about..?  😕

mysticaltech:
@mrclrchtr Don't worry about it, the action has not been updated for a long time, so just removed that check, it's not that useful anyways.

PR #1013, opened 2023-10-10
fix: dynamic k3s configuration for control planes

Silvest89:
Thanks to @mraerino for the original implementation I just did the error handling

Original PR
https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/881#issue-1795512839

### DIFF
--- a/control_planes.tf
+++ b/control_planes.tf
@@ -77,6 +77,68 @@ resource "hcloud_load_balancer_service" "control_plane" {
   listen_port      = "6443"
 }
 
+locals {
+  k3s-config = { for k, v in local.control_plane_nodes : k => merge(
+    {
+      node-name = module.control_planes[k].name
+      server = length(module.control_planes) == 1 ? null : "https://${
+        var.use_control_plane_lb ? hcloud_load_balancer_network.control_plane.*.ip[0] :
+        module.control_planes[k].private_ipv4_address == module.control_planes[keys(module.control_planes)[0]].private_ipv4_address ?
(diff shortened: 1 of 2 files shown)


control_planes.tf
```diff
@@ -144,7 +164,7 @@ resource "null_resource" "control_planes" {
     inline = [
       "systemctl start k3s 2> /dev/null",
       <<-EOT
-      timeout 360 bash <<EOF
+      timeout 120 bash <<EOF
```
mysticaltech, line 167: Sometimes it takes time to start, especially on very weak nodes, so best to keep the original timeout, it was tried and tested! (Was 120 before, but that turned out not robust enough).

mysticaltech:
@Silvest89 This is exactly what was missing, thanks! You've tested it right?

control_planes.tf
Silvest89: Alright will fix it later.

Silvest89:
@mysticaltech 
Updated the PR.
And yes it's tested. Spinned up a single node cluster to test some configs changes. And an invalid config ;]

image

I will open a PR for the workers after this is merged. And you can expect more PRs to support more config changes ;]

control_planes.tf
mysticaltech: Ah ok, no worries then! Will fix after merging.

mysticaltech:
@Silvest89 Love it! Merging ASAP. I needed the help, thanks for this.

PR #1021, opened 2023-10-12
chore: add pre-commit config

Silvest89:
To keep PR's consistent I've added the pre-commit that I currently use.

I commented out the tflint, since well this project does not adhere to all the rules yet ;]
But this should at least make sure EOL is there. And auto generates the docs as well! 

Make sure you have the following packages
```
pre-commit
terraform
tfsec
terraform-docs
```

Execute this in the root
`pre-commit install`


mysticaltech (approved):
Lovely!

PR #1024, opened 2023-10-13
fix: validation of var.existing_network_id repaired.

valkenburg-prevue-ch:
As per @mysticaltech 's comment, the validation check was broken.


mysticaltech (approved):
Wonderful

PR #1030, opened 2023-10-15
feat: dynamic k3s configuration for agents

Silvest89:
Applied the same changes for the agent nodes
Sorry for the delay ;]

Also updated the update script so it checks if the file to backup exists first.
This prevents on cluster init that you get a `cp` error (does not impact the cluster)

### DIFF
--- a/agents.tf
+++ b/agents.tf
@@ -40,6 +40,49 @@ module "agents" {
   ]
 }
 
+locals {
+  k3s-agent-config = { for k, v in local.agent_nodes : k => merge(
+    {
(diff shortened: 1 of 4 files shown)


mysticaltech (approved):
Looking good

mysticaltech:
@Silvest89  You have been bringing tons of good stuff to this project. I really want to say thank you and it would be an honor to have you be part of the team. It will give you commit access and give you will be invited to our discord for direct messaging and sharing with the team (that discord tend to be very calm, but it's there when we need it). Just let me know if you would like that!  🙏

PR #1034, opened 2023-10-17
Fix/lb upgrade

mysticaltech:
This fixes the 2.8.x upgrade identified in #1028.

### DIFF
--- a/init.tf
+++ b/init.tf
@@ -1,6 +1,6 @@
 resource "hcloud_load_balancer" "cluster" {
   count = local.has_external_load_balancer ? 0 : 1
-  name  = var.cluster_name
+  name  = local.load_balancer_name
 
   load_balancer_type = var.load_balancer_type
   location           = var.load_balancer_location
--- a/locals.tf
+++ b/locals.tf
@@ -101,7 +101,7 @@ locals {
   })
 
   apply_k3s_selinux = ["/sbin/semodule -v -i /usr/share/selinux/packages/k3s.pp"]
-  swap_node_label = ["node.kubernetes.io/server-swap=enabled"]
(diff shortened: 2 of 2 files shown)


Silvest89:
@mysticaltech 
I would add a delete_protection  = true on the loadbalancer else the hcloud controller can still delete it. This way it ignores it. 

Official docs
```
If you delete this Service in Kubernetes, the hcloud-cloud-controller-manager will delete the associated Load Balancer. If the Load Balancer is managed through Terraform, this causes problems. To disable this, you can enable deletion protection on the Load Balancer, this way hcloud-cloud-controller-manager will just skip deleting it when the associated Service is deleted.
```

mysticaltech:
> @mysticaltech I would add a delete_protection = true on the loadbalancer resource else the hcloud controller can still delete it. This way it ignores it.
> 
> Official docs
> 
> ```
> If you delete this Service in Kubernetes, the hcloud-cloud-controller-manager will delete the associated Load Balancer. If the Load Balancer is managed through Terraform, this causes problems. To disable this, you can enable deletion protection on the Load Balancer, this way hcloud-cloud-controller-manager will just skip deleting it when the associated Service is deleted.
> ```

Most definitely. Now I understand better what you were mentioning before. Will do! 🙏

mysticaltech:
@Silvest89 I tried, but it turns out that the delete protection is way too strong and just refuses to delete by any means we use now. Unless the cluster name or ingress_controller changes, there is no reason for terraform to ever delete it! So going to merge without this for now. 

Screenshot from 2023-10-17 23-55-10

Screenshot from 2023-10-17 23-54-57

mysticaltech:
@kube-hetzner/core Merging without review as it's been 24h and it's an urgent fix. Thanks for your understanding folks.

PR #1044, opened 2023-10-18
feat: kubelet args per nodepool

Silvest89:
This is especially handy when you want to set different reserved compute resources based on nodepool specs

### DIFF
--- a/agents.tf
+++ b/agents.tf
@@ -46,7 +46,7 @@ locals {
       node-name     = module.agents[k].name
       server        = "https://${var.use_control_plane_lb ? hcloud_load_balancer_network.control_plane.*.ip[0] : module.control_planes[keys(module.control_planes)[0]].private_ipv4_address}:6443"
       token         = local.k3s_token
-      kubelet-arg   = local.kubelet_arg
+      kubelet-arg   = concat(local.kubelet_arg, var.k3s_global_kubelet_args, var.k3s_agent_kubelet_args, v.kubelet_args)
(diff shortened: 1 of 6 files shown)


Silvest89:
@mysticaltech 
Persisting these args in the config.yaml instead of during the install script is the better way

k3s_exec_server_args
k3s_exec_agent_args

Are kinda obsolete. Not the recommended way to do it.

K3S Docs:
```
If you set configuration when running the install script, but do not set it again when re-running the install script, the original values will be lost.

The contents of the [configuration file](https://docs.k3s.io/installation/configuration#configuration-file) are not managed by the install script. If you want your configuration to be independent from the install script, you should use a configuration file instead of passing environment variables or arguments to the install script.
```

mysticaltech (approved):
Beautiful 🙏

mysticaltech:
> @mysticaltech 
> Persisting these args in the config.yaml instead of during the install script is the better way
> 
> k3s_exec_server_args
> k3s_exec_agent_args
> 
> Are kinda obsolete.
> 
> K3S Docs:
> ```
> If you set configuration when running the install script, but do not set it again when re-running the install script, the original values will be lost.
> 
> The contents of the [configuration file](https://docs.k3s.io/installation/configuration#configuration-file) are not managed by the install script. 
> If you want your configuration to be independent from the install script, you should use a configuration file instead of passing environment variables or arguments to the install script.
> ```

Oh wow, thanks for the info! Makes sense.

Issue #934, opened 2023-08-12
[Bug]: Creation stuck on waiting for the system-upgrade-controller


mducoli:
### Description

Hi, I'm trying to deploy a single node cluster but the creation gets stuck on waiting for the system-upgrade-controller deployment to become available (stuck on "Still creating...").
Here is the log:

```
module.kube-hetzner.null_resource.kustomization (remote-exec): + echo 'Waiting for the system-upgrade-controller deployment to become available...'
module.kube-hetzner.null_resource.kustomization (remote-exec): Waiting for the system-upgrade-controller deployment to become available...
module.kube-hetzner.null_resource.kustomization (remote-exec): + kubectl -n system-upgrade wait --for=condition=available --timeout=360s deployment/system-upgrade-controller
module.kube-hetzner.null_resource.control_planes["0-0-control-plane-fsn1"] (remote-exec): + mkdir -p /etc/rancher/k3s
module.kube-hetzner.null_resource.control_planes["0-0-control-plane-fsn1"] (remote-exec): + '[' -f /tmp/config.yaml ']'
module.kube-hetzner.null_resource.control_planes["0-0-control-plane-fsn1"] (remote-exec): + mv /tmp/config.yaml /etc/rancher/k3s/config.yaml
module.kube-hetzner.null_resource.control_planes["0-0-control-plane-fsn1"] (remote-exec): + chmod 0600 /etc/rancher/k3s/config.yaml
module.kube-hetzner.null_resource.control_planes["0-0-control-plane-fsn1"] (remote-exec): + '[' -e /etc/rancher/k3s/k3s.yaml ']'
module.kube-hetzner.null_resource.control_planes["0-0-control-plane-fsn1"] (remote-exec): + exit 0
module.kube-hetzner.null_resource.control_planes["0-0-control-plane-fsn1"]: Provisioning with 'remote-exec'...
module.kube-hetzner.null_resource.control_planes["0-0-control-plane-fsn1"] (remote-exec): Connecting to remote host via SSH...
module.kube-hetzner.null_resource.control_planes["0-0-control-plane-fsn1"] (remote-exec):   Host: 128.140.125.10
module.kube-hetzner.null_resource.control_planes["0-0-control-plane-fsn1"] (remote-exec):   User: root
module.kube-hetzner.null_resource.control_planes["0-0-control-plane-fsn1"] (remote-exec):   Password: false
[...]

PR #1053, opened 2023-10-23
Pin Traefik Chart to v24.0.0

mxmxchere:
Traefik Chart v25.0.0 was released with a breaking change (https://github.com/traefik/traefik-helm-chart/pull/934). This is a quick fix to pin to the old behaviour.

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -502,7 +502,9 @@ service:
 ports:
   web:
 %{if var.traefik_redirect_to_https~}
-    redirectTo: websecure
+    redirectTo:
+      port: websecure
+      priority: 10
 %{endif~}
 %{if !local.using_klipper_lb~}
     proxyProtocol:
--- a/templates/traefik_ingress.yaml.tpl
+++ b/templates/traefik_ingress.yaml.tpl
@@ -11,6 +11,7 @@ metadata:
   namespace: kube-system
 spec:
   chart: traefik
+  version: v24.0.0
   repo: https://traefik.github.io/charts
   targetNamespace: traefik
   bootstrap: true
--- a/templates/traefik_ingress.yaml.tpl
+++ b/templates/traefik_ingress.yaml.tpl
@@ -11,7 +11,7 @@ metadata:
   namespace: kube-system
 spec:
   chart: traefik
-  version: v24.0.0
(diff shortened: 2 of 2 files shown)


Silvest89:
The PR is merged, so is the pinning still needed?

benbehringer:
Without pinning: "Waiting for load-balancer to get an IP..."

Thanks for the quick fix!

mxmxchere:
> The PR is merged, so is the pinning still needed?

I had to be a bit more clear. The pinning (to v24) is needed because https://github.com/traefik/traefik-helm-chart/pull/934 was merged (and then released as v25, which now needs a different values syntax).

mysticaltech:
> > The PR is merged, so is the pinning still needed?
> 
> I had to be a bit more clear. The pinning (to v24) is needed _because_ https://github.com/traefik/traefik-helm-chart/pull/934 was merged (and then released as v25, which now needs a different values syntax).

Ok, that good to know, we shall move to the new traefik values syntax, that would be more future proof, then you folks will be able to upgrade.

mxmxchere:
> Without pinning: "Waiting for load-balancer to get an IP..."

I don't know if this is really related. The error i get is:

`Error: UPGRADE FAILED: execution error at (traefik/templates/_podtemplate.tpl:567:19): ERROR: Syntax of `ports.web.redirectTo` has changed to `ports.web.redirectTo.port`. Details in PR #934.`

The correct solution is to use the new values format of course. I will have a look into that now. The pinning was just a quickfix.

mysticaltech:
It seems that does not apply to the default values! Do you see something that needs changing?

mxmxchere:
I will have a look now

mysticaltech:
Ok, I will try on my end.

mxmxchere:
I now used v25.0.0 and the proposed syntax, that also worked fine

mysticaltech:
@mxmxchere Thanks for confirming. I will double check in the code if we make use of traefik redirects or not and do the necessary changes if needed. Closing this for now, but appreciate the initiative 🙏

mysticaltech:
That was fixed in v2.8.3. Thanks again for the heads up folks! 🚀

PR #1056, opened 2023-10-24
feat: taints and labels per autoscaler nodepool

Silvest89:
@mysticaltech it has backwards compatibility now :)

Also added a warning in the docs. Since it depends on an external party ;]

### DIFF
--- a/autoscaler-agents.tf
+++ b/autoscaler-agents.tf
@@ -1,28 +1,33 @@
 locals {
   cluster_prefix = var.use_cluster_name_in_node_name ? "${var.cluster_name}-" : ""
+  first_nodepool_snapshot_id = length(var.autoscaler_nodepools) == 0 ? "" : (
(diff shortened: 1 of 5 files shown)


mysticaltech (approved):
LGTM 🙏🙏

mysticaltech:
@Silvest89 While trying to upgrade an already created cluster deployed with staging. I get this error:

Screenshot from 2023-10-24 03-09-19

mysticaltech:
I checked at that line, nothing suspicious, only the values yaml, so better you check for yourself what could be going on. Note that I resolved a conflict while merging, but I chose this branch's version.

mysticaltech:
@Silvest89 Confirmed also for non-upgrade scenario, exact same issue.

Silvest89:
Let me check it

mysticaltech:
Thanks! This is my merge conflict correction, so nothing really changed: https://github.com/Silvest89/terraform-hcloud-kube-hetzner/commit/0da340f85903d1ceb2efcc5ce7f9a1644bee52b3

PR #1059, opened 2023-10-24
Fix indentation

Silvest89:


mysticaltech:
@Silvest89 Good catch!! Yaml is always tricky 😅

Silvest89:
@mysticaltech especially when there is no proper linting for these template files

mysticaltech:
Yes, good point! We should definitely add it.

mysticaltech:
@Silvest89 Still we have an issue. It turns out HCLOUD_INIT_CONFIG cannot be blank, the autoscaler pod fails to start with it blank.

Screenshot from 2023-10-24 21-34-41

Screenshot from 2023-10-24 21-34-59

mysticaltech:
@Silvest89  It does not work when HCLOUD_INIT_CONFIG is removed, (as this is necessary the node to boot correctly and get all set up). 

I'm going to remove the check on isUsingLegacyConfig. Will post the PR shortly for your analysis.

Here's the error:

Screenshot from 2023-10-24 21-34-59

Screenshot from 2023-10-24 21-34-41

mysticaltech:
Ah I see, you also include the cloud config but with index. Now I understand better. In that case, you may want to check for yourself please.

mysticaltech:
Ah we are using the old autoscaler version in the project, it's set as default to v1.27.3. But I'm seeing the new version has not been released yet in the format that we use, via manifest. Or has it? https://github.com/kubernetes/autoscaler/releases

mysticaltech:
I see no trace of new container released for the latest version in the releases of autoscaler. As we use the manifest, it calls the container directly.

However, maybe if we install via helm, it could work? I will try also search for the updated container in the helm release.

mysticaltech:
The default tag in their latest helm release even uses an older version that we use.
Screenshot from 2023-10-24 21-34-59

mysticaltech:
@Silvest89 All in all, maybe they are waiting for the next v1.28.x to release what your merged? If you can think of a workaround please let me know.

PR #1062, opened 2023-10-24
New autoscaler

mysticaltech:
@Silvest89 I have moved all the autoscaler changes in staging to here.

Continuation of #1059.


mysticaltech:
@Silvest89 I do not know if you saw my message from some time ago here https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/1030#issuecomment-1765556731, I have invited you, but feel free to join or not. I would understand if you can't join the org for various reasons and please know that your contributions are always so much appreciated.

mysticaltech:
@Silvest89 FYI, I have also reset staging to mirror master just after cherry-picking (moving) the zram changes to there.

mysticaltech:
For this branch, I wonder if we should not move to helm for the autoscaler, what do you think? In either case, it seems we are going to have to wait for your changes to show up at the container level.

Silvest89:
Yeah you cannot use the new taints and labels until their newest version is released. So you have to use the old method

mysticaltech:
> Yeah you cannot use the new taints and labels until their newest version is released. So you have to use the old method. I have built my own image for my cluster. So it's working fine here 🤣


Haha... That would explain why you did not face those issues lol. Great then, let's wait for the autoscaler version to drop! 🤞

Silvest89:
@mysticaltech They do like releases every 2-3 months or so xD
What about building our own image based off on their master with the date as tag ?
We can push it to Github

mysticaltech:
@Silvest89 We have some unused donation money, we could do that.I will look into it, good idea 🙏

Silvest89:
@mysticaltech 
We can fully use github runners are free for public repos anyway, and we can host it through Github Packages. So should not cost anything 🤭

mysticaltech:
@Silvest89 Ok perfect l, let's do this.

mysticaltech:
@Silvest89 After updating the autoscaler container uri, it now works like a charm! Wonderful work. Merging now.

Issue #1073, opened 2023-10-31
[Bug]: Randomly Job for k3s.service failed when terraform deploy


tschuering:
### Description

Hey guys, I am new to kube-hetzner project and first I want to say: Great work, we can finally now go away from spending too much amount of money on EKS! :)

I like to automate certain things, and that is why we facing currently the following issue. I am setting up a super simple k3s cluster up via your example kube.tf and I face the issue, that randomly, the deployment hangs with "Job for k3s.service failed because the control process exited with error code." after for example I destroyed the cluster and redeployed it again. Always checked that everything in the Hetzner cloud project was deleted. This happens purely random at the moment. Sometimes it works and the deploy goes through without one error, sometimes just not.

The goal is that we have our own Terraform wrapper module for internal reusage of creating clusters in Hetzner cloud in our wanted way with your module and attach later different helm stacks behind that. So we like to automate that module by actually running deploy and destroy when things were merged into that internal module in GitLab.

Also the helm stacks later will use that internal module as requirement and should automatically deploy even if we have different branches to for testing.

You see also in the screenshot the error which causes the system service not to start on the first cluster node. I am using freshest v2.9.0 release. If you have any hint where otherwise to look when this error happens, let me know, i am pretty new to k3s.

Posted the part of module "kube-hetzner", as the rest is the same standard stuff from the kube.tf

### Kube.tf file

```terraform
module "kube-hetzner" {
  providers = {
    hcloud = hcloud
  }
  hcloud_token = var.hcloud_token != "" ? var.hcloud_token : local.hcloud_token
  source = "kube-hetzner/kube-hetzner/hcloud"
  ssh_public_key = file("...")
  ssh_private_key = file("...")
  control_plane_nodepools = [
    {
      name        = "control-plane-fsn1",
      server_type = "cax21",
      location    = "fsn1",
      labels      = [],
      taints      = [],
      count       = 1
    },
    {
      name        = "control-plane-nbg1",
      server_type = "cax21",
      location    = "nbg1",
      labels      = [],
      taints      = [],
      count       = 1
    },
    {
      name        = "control-plane-hel1",
      server_type = "cax21",
      location    = "hel1",
      labels      = [],
      taints      = [],
      count       = 1
    }
  ]

  agent_nodepools = [
    {
      name        = "agent",
      server_type = "cpx11",
      location    = "fsn1",
      labels      = [],
      taints      = [],
      count       = 3
    },
  ]

[...]

PR #1077, opened 2023-11-03
Fix missing dependencies for #1073

tschuering:
@mysticaltech @Silvest89 This basically fixes all the issues I experienced in #1073 .

- [x] The problem mentioned in the last comment regarding k3s.service does not occur anymore
- [x] I am able to destroy and deploy the same cluster config again and again

I 100% sure, that all other problems I had in #1073 were basically only happening because of pure randomness. Due these changes updating k3s configs only happens when first node is provisioned with base config and has running k3s.service. Also made sure that kustomization regarding all the other stuff CNI/CSI... is happening after all control nodes are deployed and have running k3s.

### DIFF
--- a/control_planes.tf
+++ b/control_planes.tf
@@ -139,6 +139,11 @@ resource "null_resource" "control_plane_config" {
   provisioner "remote-exec" {
     inline = [local.k3s_config_update_script]
   }
+
+  depends_on = [
+    null_resource.first_control_plane,
+    hcloud_network_subnet.control_plane
+  ]
 }
 
 resource "null_resource" "control_planes" {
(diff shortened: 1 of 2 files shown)


mysticaltech (approved):
LGTM, will test and if all good will merge. Thanks.

aleksasiriski (approved):
I've tested creating a new cluster and it works as well as it did before for me. If this helps solve some edge case bugs it's a welcome fix! Also, it does make sense to wait for all control planes before applying things to the cluster.

mysticaltech:
Thanks @aleksasiriski for testing this, really appreciate it!  🙏

PR #1088, opened 2023-11-08
Fix default k3s version in kube.tf

thomasprade:
Comment in kube.tf diverged from default value in variables.tf


mysticaltech (approved):
Thanks for that!

Silvest89:
@mysticaltech yeah, I would just skip it

PR #1091, opened 2023-11-09
add option to enable or disable k3s local-storage controller

mnbro:


mysticaltech:
@mnbro  Thank you for this. Please explain the reasoning behind this proposed change. It may seem obvious, but it's not for me, thanks! 🙏

mysticaltech:
@mnbro Thanks for adding context.

So when local storage is enabled, you won't need Longhorn to use it, correct?

But what will it use? I do not think Hetzner CSI support this. Or is it host path? 

I would like to understand the exact use case, and what will be using that. Also you have to understand that our nodes use a specific partitioning scheme, with limit spaces for some directories and read only for others, so important to know where the data will be stored.

The only R-W directory that is not part of a snapshot is `var/`, important that it uses that. I do not see any config pointing to that in this PR, unless I missed it.

mysticaltech:
@kube-hetzner/core  FYI my above recommendation. Feedback welcome. Also more on the context explained here: https://chat.openai.com/share/2e9050f1-2a75-4293-86d5-bef232fa4132

mysticaltech:
Ok I see, thanks for the clarification, will merge tonight.

aleksasiriski:
@mnbro Isn't [openebs-local-pv](https://openebs.io/docs/concepts/localpv) with it's variety of types a better option to use?

aleksasiriski:
openebs-local-pv hostpath seems identical to rancher local-path-provisioner. I'm fine with the option to enable the rancher's implementation.

openebs also offer zfs, lvm, rawfile and block local-pvs as well as using any pvc to maki it an RWX with nfs server, in case someone needs this info

mysticaltech:
@mnbro Thanks for your contribution and patience. This was just released in v2.10.0  🙏

PR #1097, opened 2023-11-22
added additional documentation for the firewall_kube_api_source setting

lv2u:
As it stood, it looked like the `firewall_kube_api_source` setting would also apply firewall rules to the load balancer when the `use_control_plane_lb` setting was set to true. 

Adding additional documentation to the kube.tf.example file, so that it is clear that firewall rules cannot be applied to load balancers just yet. Hopefully this addition prevents users from having their API exposed to the world whilest they might think it is not.


mysticaltech (approved):
@lv2u  Thanks for this!

PR #1098, opened 2023-11-29
fix: ignore changes to LB's hcloud-ccm/service-uid label

mnencia:
Hetzner cloud-controller-manager automatically adds the "hcloud-ccm/service-uid" label to identify the load balancer in case of a rename.

Previously, Terraform removed the label at every application, only for it to be added back immediately by the manager.


mnencia:
@mysticaltech do you have any insight on why the trivy check is failing? Should I investigate?

PR #1099, opened 2023-11-29
Link explaining the tls setup via cert-managet and traefic proxy

vyskocilm:
Thanks for your project. As a novice in k8s ops world it helped me tremendously when deploying the cluster.

The problem is that README is apparently written by and for experts. So the sentence about how to setup TLS - while technically correct - does not make any sense to people like me :-). Added a link to Traefik blog with detailed instructions.


mysticaltech (approved):
@vyskocilm  Appreciate this! :) Thanks.

PR #1104, opened 2023-12-01
Kube 1.28

mysticaltech:
Upgraded all relevant versions and bumped k3s to v1.28.

Follows on the footsteps of #995.


mysticaltech:
@clementguillot  This brings v1.28, thanks for starting it a month and a half ago. Now live in v2.10.0. 

@kube-hetzner/core FYI!

PR #1112, opened 2023-12-06
fix: Kubelet args examples, and set the same as defaults

### LINKED ISSUE #1109: [Bug]: Kubelet args from example
### Description

Using kube.tf examples for kubelet args results in error:
```
Error: failed to run Kubelet: --system-reserved value failed to parse: failed to parse quantity "memory=200Mi" for "cpu" resource: quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'
Dec 04 13:23:00 k3s-control-plane-fsn1-mbq k3s[5016]: time="2023-12-04T13:23:00Z" level=fatal msg="kubelet exited: failed to run Kubelet: --system-reserved value failed to parse: failed to parse quantity \"memory=200Mi\" for \"cpu\" resource: quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'"
```

~~My guess is that~~ it should be:
```terraform
["kube-reserved=cpu=100m,memory=200Mi,ephemeral-storage=1Gi", "system-reserved=cpu=100m,memory=200Mi"]
```

Also, reading minimum requirement for MicroOS [here](https://microos.opensuse.org/blog/2020-11-23-requirements/) and for K3s [here](https://docs.k3s.io/reference/resource-profiling?_highlight=ram#k3s-agent) I would argue that these example values should be used:

For control planes:
```terraform
kubelet_args = ["kube-reserved=cpu=250m,memory=1500Mi,ephemeral-storage=1Gi", "system-reserved=cpu=250m,memory=300Mi"]
```

And for agents:
```terraform
kubelet_args = ["kube-reserved=cpu=50m,memory=300Mi,ephemeral-storage=1Gi", "system-reserved=cpu=250m,memory=300Mi"]
```

**If we decide that this is correct I will gladly make a PR. Also, I would like to propose that we make these example values applied by default**

### Kube.tf file

```terraform
kubelet_args = ["kube-reserved=cpu=100m,memory=200Mi,ephemeral-storage=1Gi", "system-reserved=cpu=memory=200Mi"]
```

### Screenshots

_No response_

### Platform

Linux

aleksasiriski:
Closes #1109

### DIFF
--- a/kube.tf.example
+++ b/kube.tf.example
@@ -128,7 +128,7 @@ module "kube-hetzner" {
       count       = 1
       # swap_size   = "2G" # remember to add the suffix, examples: 512M, 1G
       # zram_size   = "2G" # remember to add the suffix, examples: 512M, 1G
-      # kubelet_args = ["kube-reserved=cpu=100m,memory=200Mi,ephemeral-storage=1Gi", "system-reserved=cpu=memory=200Mi"]
+      # kubelet_args = ["kube-reserved=cpu=250m,memory=1500Mi,ephemeral-storage=1Gi", "system-reserved=cpu=250m,memory=300Mi"]
 
       # Enable automatic backups via Hetzner (default: false)
       # backups = true
@@ -167,7 +167,7 @@ module "kube-hetzner" {
       count       = 1
       # swap_size   = "2G" # remember to add the suffix, examples: 512M, 1G
       # zram_size   = "2G" # remember to add the suffix, examples: 512M, 1G
-      # kubelet_args = ["kube-reserved=cpu=100m,memory=200Mi,ephemeral-storage=1Gi", "system-reserved=cpu=memory=200Mi"]
+      # kubelet_args = ["kube-reserved=cpu=50m,memory=300Mi,ephemeral-storage=1Gi", "system-reserved=cpu=250m,memory=300Mi"]
 
       # Enable automatic backups via Hetzner (default: false)
       # backups = true
--- a/variables.tf
+++ b/variables.tf
@@ -174,7 +174,7 @@ variable "control_plane_nodepools" {
     count        = number
     swap_size    = optional(string, "")
     zram_size    = optional(string, "")
-    kubelet_args = optional(list(string), [])
+    kubelet_args = optional(list(string), ["kube-reserved=cpu=250m,memory=1500Mi,ephemeral-storage=1Gi", "system-reserved=cpu=250m,memory=300Mi"])
   }))
   default = []
   validation {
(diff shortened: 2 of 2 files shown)


aleksasiriski:
> Seems like they are ideally suited to this? If they are minimum requirements and the goal is to add them to all nodes and control planes? Maybe I have missed something.

As I've added them currently, if you don't specify anything in the node pool sections, settings are applied. If you do specify anything, it's overridden. So it is applied correctly to all of the nodes.

aleksasiriski:
> I am learning as I go, but have encountered some out of memory issues with these memory limits. Only under totally unreasonable strain and non-production, but was expecting it not to be possible to encounter a total crash but for pods to be evicted.
> 
> Going to try again with this: https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/#node-pressure-eviction-good-practices
> 
> This is the issue encountered: https://www.suse.com/support/kb/doc/?id=000018698

Hmm interesting.. We could also add those other claims but I myself am not using any swap on nodes. I will try out ZRAM soon in my cluster but swap to disk is generally bad and should be avoided especially for Kubernetes nodes (even though it was recently introduced to this script). Have you tried with ZRAM only (compressing RAM instead of swapping to disk)?

mysticaltech (approved):
Awesome, thanks @aleksasiriski. Next time please tag me to review and merge faster 🙏

aleksasiriski:
> I am learning as I go, but have encountered some out of memory issues with these memory limits. Only under totally unreasonable strain and non-production, but was expecting it not to be possible to encounter a total crash but for pods to be evicted.
> 
> Going to try again with this: https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/#node-pressure-eviction-good-practices
> 
> This is the issue encountered: https://www.suse.com/support/kb/doc/?id=000018698

One of my nodes on production randomly stopped working (even ssh isn't available). I'm investigating if this might be the cause. I guess this is related to system-reserved and not kube-reserved?

mysticaltech:
I have no idea. I had the same happen to one of my clusters recently. We need to have better monitoring of both cluster and nodes to be able to debug.

aleksasiriski:
@mysticaltech Do you have these kubelet args set there?

mysticaltech:
> @mysticaltech Do you have these kubelet args set there?

@aleksasiriski  It was an old cluster deploy a few months back. So probably not using kubelet args. I have replaced it since then, so now using latest with defaults and seems ok. Will continue to monitor.

PR #1115, opened 2023-12-08
chore(docs): k3s_upgrade label

cygnusb:
Remove namespace from syntax, which is not needed when editing node labels. Fix '_' to '-' and clarify text.


aleksasiriski (approved):

mysticaltech:
Thanks @cygnusb, @aleksasiriski.

PR #1119, opened 2023-12-13
Kured automatic unlock

### LINKED ISSUE #1117: [Bug/Issue]: kured may get stuck
I had the issue that kured got stuck because of an autoscaled node being killed and I had to unlock it manually, I checked the arguments of kured and I don't see this option being set and from what I can read out of the documentation kured highly recommends this to be set when using the cluster autoscaler. So I think this should be set as an argument by default on kured

These are the default arguments of kured:
image

This is taken from kured's documentation:
https://kured.dev/docs/operation/#automatic-unlock
image

mysticaltech:
Fixes #1117


aleksasiriski (approved):

Timeraa (approved):

mnencia (approved):

mnencia:
@mysticaltech I've noticed that the kustomization was not re-executed after the change. I wonder if we should use the local value instead of vars also for all the other values in the version field (where applicable).

mnencia (approved):

mysticaltech:
@mnencia You are right, it's better that way, so it applies the change even on upgrade. Thanks for the push!  🙏

PR #1123, opened 2023-12-13
feat: specify the version of the ingress (traefik & nginx) helm chart

mnencia:
With these two new options, you can control the helm chart version for traefik and nginx ingresses.

### DIFF
--- a/init.tf
+++ b/init.tf
@@ -131,6 +131,8 @@ resource "null_resource" "kustomization" {
       coalesce(var.kured_version, "N/A"),
       coalesce(var.calico_version, "N/A"),
       coalesce(var.cilium_version, "N/A"),
+      coalesce(var.traefik_version, "N/A"),
+      coalesce(var.nginx_version, "N/A"),
     ])
     options = join("\n", [
       for option, value in var.kured_options : "${option}=${value}"
@@ -156,6 +158,7 @@ resource "null_resource" "kustomization" {
     content = templatefile(
       "${path.module}/templates/traefik_ingress.yaml.tpl",
       {
+        version          = var.traefik_version
         values           = indent(4, trimspace(local.traefik_values))
         target_namespace = local.ingress_target_namespace
     })
@@ -167,6 +170,7 @@ resource "null_resource" "kustomization" {
     content = templatefile(
       "${path.module}/templates/nginx_ingress.yaml.tpl",
       {
+        version          = var.nginx_version
         values           = indent(4, trimspace(local.nginx_values))
         target_namespace = local.ingress_target_namespace
     })
--- a/templates/nginx_ingress.yaml.tpl
+++ b/templates/nginx_ingress.yaml.tpl
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: nginx
+  name: ${target_namespace}
 ---
 apiVersion: helm.cattle.io/v1
 kind: HelmChart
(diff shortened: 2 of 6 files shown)


init.tf
```diff
@@ -131,6 +131,8 @@ resource "null_resource" "kustomization" {
       coalesce(var.kured_version, "N/A"),
       coalesce(var.calico_version, "N/A"),
       coalesce(var.cilium_version, "N/A"),
+      coalesce(var.traefik_chart_version, "N/A"),
+      coalesce(var.nginx_chart_version, "N/A"),
```
mysticaltech, line 135: Could we remove the _chart_ so that the naming becomes consistant with the other ones like cilium_version?

templates/nginx_ingress.yaml.tpl
```diff
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: nginx
+  name: ${target_namespace}
```
mysticaltech, line 5: I do not see the logic that determines the value of `target_namespace`. Where is it?

templates/nginx_ingress.yaml.tpl
```diff
@@ -11,6 +11,7 @@ metadata:
   namespace: kube-system
 spec:
   chart: ingress-nginx
+  version: "${version}"
```
mysticaltech, line 14: Here also, I'm not seeing the value that determines the value of `version` and fetches the latest if empty. Maybe I'm missing something.

mysticaltech (changes requested):
@mnencia  So good seeing your PRs again! Just a few comments. Maybe some changes were not committed on your end.

init.tf
mnencia: We can, but what you specify here is the chart version that in general is different from the software version. If you think it's better to remove the chart word I'll do it

templates/nginx_ingress.yaml.tpl
mnencia: If you do not specify anything you revert to the previous behavior, and you select the latest at the moment of the installation

templates/nginx_ingress.yaml.tpl
mnencia: The value is already used in the chart below. It is passed to the template from the local.ingress_namespace variable. See init.tf file.

init.tf
mysticaltech: @mnencia Ah I understand now, the chart version does not correspond to the software version, so all is good. Thanks for the clarification!

templates/nginx_ingress.yaml.tpl
mysticaltech: Ah yes, my bad. Awesome!

templates/nginx_ingress.yaml.tpl
mysticaltech: Ok, wonderful!

init.tf
mysticaltech: I see you made the change, since it's not used anywhere else I think we can keep it like with without _chart_, just for consistancy. Thanks!

mysticaltech (approved):
LGTM  🚀

aleksasiriski (approved):
I've literally wanted to add these for the past couple of days, with #1126 as well

mysticaltech:
@mnencia  FYI, please join us on discord https://discord.gg/VcSYQp7D

PR #1126, opened 2023-12-14
feat: added option to pin traefik image tag with example for traefik beta

aleksasiriski:
Useful if someone wants the new v3 opentelemetry metrics and tracing.


mnencia:
That's very interesting! 

I have two questions:

- what about the CRD? Is it compatible with the v3?
- why not pass the whole `image` section instead of the only `image.tag` value?

aleksasiriski:
Everything is crosscompatible, v3 only introduces some new features like otel. I only needed image tag, but I can add more options. I think I best leave that to the v3 (kube-hetzner) refactor I'm working on.

PR #1134, opened 2023-12-20
feat: explicit ip for the control-plane LB, as Hetzner's automatic IP…

valkenburg-prevue-ch:
… assignment is broken (not guaranteed to be on the given subnet).

If you let Hetzner assign ip addresses, you get an address in the last created subnet, not in the subnet which you give as an argument. This fix sets the ip for the control plane load balancer explicitly, in line with other resources, so it actually is guaranteed to be in the subnet which it is supposed to end up in.

This is a long-standing bug in Hetzner Cloud, which they know about, and they have no solution for it.

### DIFF
--- a/control_planes.tf
+++ b/control_planes.tf
@@ -58,6 +58,7 @@ resource "hcloud_load_balancer_network" "control_plane" {
   load_balancer_id        = hcloud_load_balancer.control_plane.*.id[0]
   subnet_id               = hcloud_network_subnet.control_plane.*.id[0]
   enable_public_interface = var.control_plane_lb_enable_public_interface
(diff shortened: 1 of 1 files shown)


valkenburg-prevue-ch:
Uhm, where is that created? In my personal setup, the agents load balancer is created by an ingress helm chart with type loadbalancer (and the appropriate annotations to make hetzner deploy the right thing with the right name etc). I don't find a second `resource "hcloud_load_balancer_network"` in the code base.

mysticaltech:
@valkenburg-prevue-ch  My bad, I was confused, this is just for the LB, yes way better than to leave it pick any IPs. Merging now.

mysticaltech (approved):
Wonderful

PR #1139, opened 2023-12-29
chore(traefik): remove unused entrypoint

mnencia:
I noticed that the Traefik configuration includes the option `--entrypoints.tcp=true`,
which generates an entry point named `tcp` without configuring it.
While this option doesn’t cause any harm, it is entirely unnecessary.
Therefore, I suggest removing it.

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -581,7 +581,6 @@ podDisruptionBudget:
   maxUnavailable: 33%
 %{endif~}
 additionalArguments:
-  - "--entrypoints.tcp=true"
   - "--providers.kubernetesingress.ingressendpoint.publishedservice=${local.ingress_controller_namespace}/traefik"
 %{for option in var.traefik_additional_options~}
   - "${option}"


mysticaltech (dismissed):
Indeed, good catch!

aleksasiriski:
It is required for "additional ports" option I've added to traefik a while ago.

mnencia:
@aleksasiriski can you show me an example of usage? It doesn't feel right to have it always added to the configuration.

aleksasiriski:
I think I added to "always" because it causes no harm, but it can be changed to be added only when the additional_ports isn't an empty list. But the current state of this PR breaks additional_ports functionality which I use for load balancing port 22 for my git instance.

locals.tf
```diff
@@ -587,7 +587,6 @@ podDisruptionBudget:
   maxUnavailable: 33%
 %{endif~}
 additionalArguments:
-  - "--entrypoints.tcp=true"
```
aleksasiriski, line 590: Change it to:

```terraform
%{if var.traefik_additional_options != [] ~}
  - "--entrypoints.tcp=true"
%{endif~}
```

aleksasiriski (approved):

locals.tf
mnencia: I've tried the feature, and it works for me without that line. Can you five me a reproducer that doesn't work without that line? AFAIK of traefic configuration, that line has no effect.

aleksasiriski: Did it work successfully with a TCP protocol like SSH, not HTTP/HTTPS on non-standard ports? Traefik docs explain that TCP endpoint is needed if you want traefik to load balance non-http traffic (like SSH).

aleksasiriski: This was introduced here btw: https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/809

aleksasiriski: Also, I still haven't had the time to test it out really. So if you want, you can merge it and I will change it later if required.

otavio: For TCP it is indeed required

mnencia:
I have tried hard to reproduce a case that requires that line, but I failed. @aleksasiriski, if you have an example, please provide it to me.

I create a cluster with this branch, and the following configuration

```
traefik_additional_ports = [{name = "ssh", port = 2222, exposedPort = 2222}]
```

and after deploying the following code I'm able to ssh in without issues.

```
---
apiVersion: traefik.io/v1alpha1
kind: IngressRouteTCP
metadata:
  name: ssh
spec:
  entryPoints:
    - ssh
  routes:
  - match: HostSNI(`*`)
    services:
    - name: ssh
      namespace: default
      port: 2222
apiVersion: v1
kind: Service
metadata:
  name: ssh
  labels:
    app: ssh
spec:
  ports:
  - port: 2222
  selector:
    app: ssh
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: ssh
  labels:
    app: ssh
spec:
  replicas: 1
  selector:
    matchLabels:
      app: ssh
  template:
    metadata:
      labels:
        app: ssh
    spec:
      containers:
      - name: ssh
        image: lscr.io/linuxserver/openssh-server:latest
        ports:
        - containerPort: 2222
        env:
        - name: PUID
          value: "1000"
        - name: PGID
          value: "1000"
        - name: TZ
          value: "Europe/Rome"
        - name: USER_NAME
          value: "mnencia"
        - name: PUBLIC_KEY
          value: "ssh-ed25519 REDACTED mnencia@laptop"
```

aleksasiriski:
It's great to hear that it works without that line, I haven't had the time to test it out but now you did! Then it really isn't needed after all.

aleksasiriski (approved):
LGTM

PR #1164, opened 2024-01-09
refactor: modularize everything, separate files and new submodules

aleksasiriski:


maggie44:
Could we add pinning of all dependency versions and bumping as PRs to the list as mentioned here? https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/1206/files#r1497072303

I think pinned versions is a good default for production. Not all of the dependent packages are built like OpenSuse where it’s designed for rolling updates. It would relate to the CI which would then run tests against new minor and major bumps.

aleksasiriski:
> Could we add pinning of all dependency versions and bumping as PRs to the list as mentioned here? https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/1206/files#r1497072303
> 
> I think pinned versions is a good default for production. Not all of the dependent packages are built like OpenSuse where it’s designed for rolling updates. It would relate to the CI which would then run tests against new minor and major bumps.

I wholeheartedly agree, if you want you can make a PR to this PR with that fix or I will do it when I get the free time again.

BuddhiAbeyratneAL:
Im not a lawyer but from what I've seen in practice

MIT to MPL2 can't be a direct swap. 
one has a copy left ish clause that people have to comply with 
Also the old code needs to continue to distribute as MIT even if you pull it out as submodules

aleksasiriski:
> Im not a lawyer but from what I've seen in practice
> 
> MIT to MPL2 can't be a direct swap. one has a copy left ish clause that people have to comply with Also the old code needs to continue to distribute as MIT even if you pull it out as submodules

You are correct, this won't be merged until all contributors agree to it. We will be creating a poll to accomodate that.

aleksasiriski:
Closing this in favor of a complete rewrite of the project that will be pushed to a new branch and PR soon.

PR #1175, opened 2024-01-17
Add posibility to run create.sh unattended

vafgoettlich:
Hi all,
I want to be able to execute create.sh non-interactive (using already set env-Variables). So I added some Checks in Variables are empty or not if if not just using them instead of reading them from CLI.


mysticaltech (approved):
@vafgoettlich   LGTM, thanks for this!

Issue #1182, opened 2024-01-22
Remove lock ttl from the kured defaults


valkenburg-prevue-ch:
https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blame/8cb5c3e3054530da6f4d4af640d2200b6eb701bc/locals.tf#L624

IMHO this is a regression: the ttl makes the lock expire when a node is cordoned and fails to complete the reboot cycle. As a consequence, it is possible the one by one many nodes become unavailable, practically taking down the cluster. I had to understand this the hard way. I would prefer the default being what it was before, no lock expiration. We could expand the comment about kured_options in kube.tf.example maybe?

PR #1183, opened 2024-01-22
fix: removed the default kured lock expiration as it is dangerous.

valkenburg-prevue-ch:
Added comments in kube.tf.example to document the possible use of lock expiration (lock-ttl).

Addresses https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/1182 .


mysticaltech (approved):
LGTM

PR #1161, opened 2024-01-07
Enhancements to Placement Group Logic 

mysticaltech:
This proposes a fix for #1157.

#### Changes Made

1. **Introduction of Local Variables for Node Indices Mapping**

    - **Agent Nodes Mapping (`agent_nodes_indices`):**
      ```hcl
      locals {
        agent_nodes_indices = { for node_name, node_details in local.agent_nodes : node_name => floor(node_details.index / 10) }
      }
      ```
      This local variable creates a map where each key corresponds to an agent node, and the value is the index of the placement group it belongs to. The `floor(node_details.index / 10)` calculation ensures an even distribution of nodes across the placement groups, with each group containing no more than 10 nodes.

    - **Control Plane Nodes Mapping (`control_plane_nodes_indices`):**
      ```hcl
      locals {
        control_plane_nodes_indices = { for node_name, node_details in local.control_plane_nodes : node_name => floor(node_details.index / 10) }
      }
      ```
      Similar to agent nodes, this mapping for control plane nodes assigns each node to a placement group based on its index. This consistent indexing is crucial for maintaining the balance and order of nodes in the cluster.

2. **Updating Module Blocks to Utilize New Indices**

    - **Agent Nodes (`module "agents"` in `agents.tf`):**
      ```hcl
      module "agents" {
        # ... existing configuration ...

        placement_group_id = var.placement_group_disable ? null : hcloud_placement_group.agent[local.agent_nodes_indices[each.key]].id

        # ... remaining configuration ...
      }
      ```
      This update ensures that the placement group ID for each agent node is dynamically determined based on the new `agent_nodes_indices` mapping. It respects the `placement_group_disable` variable, allowing for flexible enablement or disablement of placement groups.

    - **Control Plane Nodes (`module "control_planes"` in `control_planes.tf`):**
      ```hcl
      module "control_planes" {
        # ... existing configuration ...

        placement_group_id = var.placement_group_disable ? null : hcloud_placement_group.control_plane[local.control_plane_nodes_indices[each.key]].id

        # ... remaining configuration ...
      }
      ```
      Similarly, each control plane node's placement group ID is assigned using the `control_plane_nodes_indices` mapping. This ensures a consistent and orderly assignment of control plane nodes to placement groups.

#### Rationale Behind the Changes

- **Scalability and Flexibility**: The introduction of local variables for node indices mapping provides a scalable and flexible approach to managing node assignments to placement groups. It allows for easy adjustments and scaling of the node count without disrupting the underlying logic.
[...]

PR #1185, opened 2024-01-22
feat: backwards compatible labelling per node per nodepool.

valkenburg-prevue-ch:
And now you can remove / replace individual nodes in a nodepool, instead of having to set the count all the way to zero to make sure that the one sick node is removed.

This is an example implementation of the discussion https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/discussions/1038 .


kube.tf.example
```diff
@@ -227,7 +227,33 @@ module "kube-hetzner" {
...
+    {
+      name        = "agent-arm-small",
+      server_type = "cax11",
+      location    = "fsn1",
+      labels      = [],
+      taints      = [],
+      nodes = {
+        "a1" : {
+          unique_index_in_node_pool = 0
+          location                  = "nbg1"
```
maggie44, line 244: A location for each key/value pair and a location for the node pool? It seems that it would override the default? So the use case for this would be if someone wanted multiple Nodes from different regions on the same subnet? 🤯. I don't see any reason to limit the level of fine-grained control I guess.

valkenburg-prevue-ch: yeah, I figured that if we want to override some of the node-pool settings per node, then why not all of them. But I haven't given any thought to the consequences of one subnet across regions.

valkenburg-prevue-ch: Thanks for raising the point about more documentation about unique_index_in_node_pool. I think a better name can also be found. Let me try to explain it here, and test if this explanation is good enough for the documentation:

"The private ip of the node inside the nodepool's ip range, is computed from this index. This is why it must be unique, and it should not change for a node as long as the node exists."

Can you propose a more aptly named parameter?

If you use the count instead of key/map, the index of a node in its counted set is what determines its private ip.

variables.tf
```diff
@@ -216,6 +230,32 @@ variable "agent_nodepools" {
...
+
+  validation {
+    condition = alltrue([for agent_nodepool in var.agent_nodepools : length(
+      coalesce(agent_nodepool.nodes, {})
+      ) == length(
+      distinct(
+        [for agent_node in coalesce(agent_nodepool.nodes, {}) : agent_node.unique_index_in_node_pool]
+      )
+    )])
+    # 154 because the private ip is derived from unique_index_in_node_pool + 101. See private_ipv4 in agents.tf
```
maggie44, line 247: If this is under the same Nodepool, would it be under the same 'Network' in Hetzner speak? If so, then it would be limited to 100?

https://docs.hetzner.com/cloud/networks/faq/#are-there-any-limits-on-how-networks-can-be-used

valkenburg-prevue-ch: Ah, as I understand it, the number of nodes is limited to 100, but their private ips can be anything between 101 and 254.

maggie44: Makes sense. 

It says:

```
for each node inside a nodepool.
```

So technically there couldn't be more than 100 nodes in a Nodepool right now anyway, because it seems like they are all created in the same 'Network'. But that is a different issue, so this does make sense, and then some other validator somewhere else in the code should do the 100 check instead (as part of a different PR and issue 😄) 👍. 

Or I completely misunderstood something on Hetzner networks, someone will know I'm sure. Every day is a school day.

valkenburg-prevue-ch: Well, actually... we could just add another validation rule counting that the number of nodes <= 100. (anyway, who has such high limits with Hetzner?)

valkenburg-prevue-ch: ok!

mysticaltech:
@valkenburg-prevue-ch Looks great! I will review ASAP. It's something that might help us too with the placement group issue we had. See comment by @mnencia here on his findings: https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/1161#issuecomment-1882681214. And the original issue here https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/1157. If you can address this too, I would be super greatful!

valkenburg-prevue-ch:
> @valkenburg-prevue-ch Looks great! I will review ASAP. It's something that might help us too with the placement group issue we had. See comment by @mnencia here on his findings: [#1161 (comment)](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/1161#issuecomment-1882681214). And the original issue here #1157. If you can address this too, I would be super greatful!

Thanks! I'll make some small modifications after @maggie44 already went through everything and proposed some improvements.

kube.tf.example
valkenburg-prevue-ch: Fixed now: using integers as keys in the nodes map, as suggested by @maggie44 .

variables.tf
valkenburg-prevue-ch: Fixed: added a size limit after all.

valkenburg-prevue-ch:
Oh by the way, added a section in the readme for migrating from count-based to map-based node pools, it's super simple in the end.

mnencia:
Thank you. I'll review the PR ASAP.

valkenburg-prevue-ch:
> Thank you. I'll review the PR ASAP.

Actually, now that I read the placement group issue, I realize that this PR is broken when someone uses node keys > 10 or simply more than 10 nodes, as the placement group logic doesn't account for this sparse use of indices. I think I'll make a separate PR which fixes the placement group bug, but I'm struggling to find an implementation which is backwards compatible.

mysticaltech:
@valkenburg-prevue-ch Good job with #1217 fixing the placement group, logic. So this current PR is ready too? 🙏

valkenburg-prevue-ch:
> @valkenburg-prevue-ch Good job with #1217 fixing the placement group logic. So this current PR is ready too? 🙏

Yes! I think both are ready. But if course to be tested live in a staging environment, I guess.

mysticaltech (approved):
That's some serious level of customization! Wow.

mysticaltech:
@kube-hetzner/core Please review if you can, super important PR. I think it adds loads of possibility to the agent nodes setup without making the code overly complex. @maggie44 You also please, you are one of the main stakeholders here ;)

locals.tf
```diff
@@ -156,6 +157,43 @@ locals {
...
+          zram_size : nodepool_obj.zram_size,
+          index : floor(tonumber(node_key)),
+        },
+        { for key, value in node_obj : key => value if value != null },
+        {
+          labels : concat(local.default_agent_labels, nodepool_obj.swap_size != "" ? local.swap_node_label : [], nodepool_obj.labels, coalesce(node_obj.labels, [])),
+          taints : concat(local.default_agent_taints, nodepool_obj.taints, coalesce(node_obj.taints, [])),
+        },
+        (
+          node_obj.append_index_to_node_name ? { node_name_override : "-${floor(tonumber(node_key))}" } : {}
```
maggie44, line 185: Nice catch, even if it didn't rebuild this would have changed things like `kubernetes.io/hostname: ...`, which would have been breaking. 

Hopefully once this is set the first time it can't be changed? Otherwise someone could be in for a surprise. Although if the name field allows renaming and triggers a rebuild then consistency in how it behaves makes sense.

variables.tf
```diff
@@ -216,6 +230,27 @@ variable "agent_nodepools" {
...
+    condition = alltrue([for agent_nodepool in var.agent_nodepools :
+      alltrue([for agent_key, agent_node in coalesce(agent_nodepool.nodes, {}) : can(tonumber(agent_key)) && tonumber(agent_key) == floor(tonumber(agent_key)) && 0 <= tonumber(agent_key) && tonumber(agent_key) < 154])
+    ])
+    # 154 because the private ip is derived from tonumber(key) + 101. See private_ipv4 in agents.tf
+    error_message = "The key for each individual node in a nodepool must be a stable integer in the range [0, 153] cast as a string."
+  }
+
+  validation {
+    condition = alltrue([for agent_nodepool in var.agent_nodepools : length(agent_nodepool) <= 100])
+    # 154 because the private ip is derived from tonumber(key) + 101. See private_ipv4 in agents.tf
```
maggie44, line 250: I think the inline note is supposed to be about 100 server limit.

variables.tf
```diff
@@ -216,6 +230,27 @@ variable "agent_nodepools" {
...
+  validation {
+    condition = alltrue([for agent_nodepool in var.agent_nodepools :
+      alltrue([for agent_key, agent_node in coalesce(agent_nodepool.nodes, {}) : can(tonumber(agent_key)) && tonumber(agent_key) == floor(tonumber(agent_key)) && 0 <= tonumber(agent_key) && tonumber(agent_key) < 154])
+    ])
+    # 154 because the private ip is derived from tonumber(key) + 101. See private_ipv4 in agents.tf
+    error_message = "The key for each individual node in a nodepool must be a stable integer in the range [0, 153] cast as a string."
+  }
+
+  validation {
+    condition = alltrue([for agent_nodepool in var.agent_nodepools : length(agent_nodepool) <= 100])
```
maggie44, line 249: I think Hetzner's limit is 100 `'servers'`. It doesn't use k8 speak, it is a limit imposed on how many Cloud servers can be attached to each network. So for kube-hetzner use case the limit is `nodes`+`control planes` must be <= 100 as they are all on the same network?

variables.tf
```diff
@@ -216,6 +230,27 @@ variable "agent_nodepools" {
...
+    # 154 because the private ip is derived from tonumber(key) + 101. See private_ipv4 in agents.tf
+    error_message = "Hetzner does not support nodepools with more than 100 nodes."
```
maggie44, line 251: `servers` rather than nodes.

maggie44 (commented):
This is such a great change! Opens up a whole new realm of control, can't wait to get it on to all the clusters.

mysticaltech:
@valkenburg-prevue-ch  In light of the new review(s), will await your signal to merge this 🙏  

@aleksasiriski @mnencia FYI

locals.tf
valkenburg-prevue-ch: I'm not sure actually..

variables.tf
valkenburg-prevue-ch: Actually this check is about the validity of individual integers, which are acting as a label. The total count can be < 100, yet individual nodes are allowed to have this value set at 153 etc.

variables.tf
valkenburg-prevue-ch: weird, this check is pretty broken and makes little sense, and I don't recall writing it. Will rewrite the check completely. But we can only check the agents size, you cannot acccess other variables in a validation check.

valkenburg-prevue-ch:
Rebased on master, with some added code for making the placement-group logic work well with the individual node configuration. Adopted comments from @maggie44 . New review, anyone?

mysticaltech:
> Rebased on master, with some added code for making the placement-group logic work well with the individual node configuration. Adopted comments from @maggie44 . New review, anyone?

Awesome, will review ASAP 🙏

README.md
```diff
@@ -890,6 +892,61 @@ You can also continue using the previous code-base like this:
     },
   ]
 ```
+
+Finally, if you want to have a node-pool with more than 10 nodes, you have to use the map-based
```
mysticaltech, line 896: Would be great to explain the placement-group limit of 10 nodes, and that the name is arbitrary, as long as they make sure to not have more than 10 nodes per placement-group. And if they are not going to use more than 10 nodes, then do not need to bother with the placement-group attribute.

mysticaltech (approved):
If this is well tested for backward compatibility, and I guess that it is, looks great! Super powerful multi-layered logic there  🫡

@mnencia If you can find the time to review this it would be great.

mysticaltech (approved):

mysticaltech:
@kube-hetzner/core This is merging soon, if you want to review it's now or never!

mysticaltech:
@valkenburg-prevue-ch @maggie44 Shipped in v2.13.0, thanks for the wonderful work people!

PR #1191, opened 2024-01-24
Add server location labels for load balancer targeting

mixman:
Based on https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/discussions/1187.
Adds:
- agent_nodepools[].location as a hetzner server label for the hetzner load balancer to be able to target them.
- var.load_balancer_node_selector to extend default best practice traefik_values

**BEFORE:**
Traffic -> LB -> agent_nodepools[]

**AFTER:**
`load_balancer_node_selector = "location in (nbg1,fsn1)"`
Traffic -> LB -> agent_nodepools[] in matching location

**TODO:**
- [ ] Server labels set for autoscaler_nodepools
- [ ] Traefik nodeSelector/nodeAffinity to match load_balancer_location
- [ ] Disable default LB->[servers] and use var.load_balancer_node_selector when set.

### DIFF
--- a/agents.tf
+++ b/agents.tf
@@ -30,7 +30,7 @@ module "agents" {
 
   private_ipv4 = cidrhost(hcloud_network_subnet.agent[[for i, v in var.agent_nodepools : i if v.name == each.value.nodepool_name][0]].ip_range, each.value.index + 101)
 
-  labels = merge(local.labels, local.labels_agent_node)
+  labels = merge(local.labels, local.labels_agent_node, {"location": each.value.location})
 
   automatically_upgrade_os = var.automatically_upgrade_os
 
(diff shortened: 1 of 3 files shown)


mixman:
Testing this (in Hetzner UI) some roadblocks:
- Server List is on and by default targeting all servers. Adding a Label Selector isn't in effect until Server List is removed
- While `load_balancer_location` can be set, the Private IP (and thus traefik) and subnet can be in an entirely different location.
  - eg. load_balancer_location=hel1, yet LB subnet is in fsn1 causing hel1->fsn1->hel1 routing when target pod in hel1.
  
 Is there a way to bind traefik to `load_balancer_location`?
 How can Server List be made a no op for Label Selector to be in effect?

mysticaltech:
@mixman Interesting work you are attempting here. @M4t7e  @Silvest89 any ideas?

mysticaltech:
@mixman  You can also use our traefik_values variable.

mixman:
Thanks for all the help! Closing as unneeded.

PR #1202, opened 2024-02-02
open autoscaler pod port to enable monitoring via port 8085

sebboer:
Enable monitoring for cluster autoscaler by opening pod port 8085
More infos: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-can-i-monitor-cluster-autoscaler

### DIFF
--- a/templates/autoscaler.yaml.tpl
+++ b/templates/autoscaler.yaml.tpl
@@ -164,6 +164,8 @@ spec:
             requests:
               cpu: 100m
               memory: 300Mi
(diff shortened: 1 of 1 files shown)


mysticaltech:
Thanks @sebboer! @ifeulner  @Silvest89 what do you think of that please? You are the autoscaler pros.

mysticaltech:
I do not see any harm in merging this, thanks for the contribution @sebboer!

PR #1216, opened 2024-02-15
feat: allow custom configuration for agent nodes

trivoallan:
New variable : `agent_nodes_custom_config`. Works exactly like `control_planes_custom_config`.

### DIFF
--- a/agents.tf
+++ b/agents.tf
@@ -51,9 +51,9 @@ locals {
       node-ip       = module.agents[k].private_ipv4_address
       node-label    = v.labels
       node-taint    = v.taints
-      selinux       = true
     },
-    var.agent_nodes_custom_config
+    var.agent_nodes_custom_config,
+    (v.selinux == true ? { selinux = true } : {})
   ) }
 }
 
--- a/agents.tf
+++ b/agents.tf
@@ -53,6 +53,7 @@ locals {
       node-taint    = v.taints
       selinux       = true
     },
+    var.agent_nodes_custom_config
   ) }
 }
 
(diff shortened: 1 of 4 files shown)


aleksasiriski:
You didn't mention in your PR description that you almost added an variable for disabling SELinux, it's good to mention all the changes in the description.

@kube-hetzner/core Do we want the option to disable SELinux? What is the usecase which cannot be covered by changing the policy to accommodate the needs of the user?

valkenburg-prevue-ch:
@aleksasiriski I think your mean "also" instead of "almost", right? Very confusing auto-correct.. 

Very good catch, I agree with you that this is something that should not be taken lightly. I would not know what the usecase is.

aleksasiriski:
> I think your mean "also" instead of "almost", right? Very confusing auto-correct..
I don't use autocorrect, I'm a bit [weird](https://en.wikipedia.org/wiki/Speech_error) when I'm typing quickly

> Very good catch, I agree with you that this is something that should not be taken lightly. I would not know what the usecase is.

I agree that it is good to have options as much as possible but by allowing for SELinux to be disabled we allow users to have a less secure setup.. and that would be fine if the benefits weren't only "I don't have to think about improving the policy". As far as I can see, that is the only benefit right now. Maybe @trivoallan can tell me another usecase?

mysticaltech:
@trivoallan Thanks for this PR, the agent nodes config logic is appreciated. Same as what @aleksasiriski asked about, why would you need to disable selinux? If it's not working for you, you can always collect the fix with this method https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/697#issuecomment-1496607924, and we can add it to the rules below. For what it's worth I think it's way more secure. Unless you have a particular usecase?

Here are the rules that we can always expand:
https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/e911232e73951ec873711a388b58770e10b8a80f/locals.tf#L728-L800

mysticaltech:
Alright folks, merging this, more flexibility is good.

mysticaltech:
(force-merging because of style issue caused by conflict resolution, will fix in a subsequent pre-release PR)

PR #1218, opened 2024-02-21
fix(placement_groups)!: remove old idx logic and added validations

aleksasiriski:
- [x] Removes old "compatibility" mode for placement groups (since it's not really a breaking change, changing placement groups is safe from node state perspective)
- [x] Added option to disable placement groups per node, alongside the old global var
- [x] Default placement group is named after the nodepool
- [x] Added validation that forbids using a placement group in a nodepool that has count greater than 10 (this requires old setups that used the global variable to disable placement groups to also disable them on the nodepool level)
- [x] Added validation that forbids count to be negative

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -165,18 +165,18 @@ variable "load_balancer_health_check_retries" {
 variable "control_plane_nodepools" {
   description = "Number of control plane nodes."
   type = list(object({
-    name                       = string
-    server_type                = string
-    location                   = string
-    backups                    = optional(bool)
-    labels                     = list(string)
-    taints                     = list(string)
-    count                      = number
-    swap_size                  = optional(string, "")
-    zram_size                  = optional(string, "")
-    kubelet_args               = optional(list(string), ["kube-reserved=cpu=250m,memory=1500Mi,ephemeral-storage=1Gi", "system-reserved=cpu=250m,memory=300Mi"])
-    placement_group_compat_idx = optional(number, 0)
-    placement_group            = optional(string, null)
+    name                 = string
+    server_type          = string
+    location             = string
+    backups              = optional(bool)
+    labels               = list(string)
+    taints               = list(string)
+    count                = number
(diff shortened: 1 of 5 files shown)


variables.tf
```diff
@@ -189,25 +189,37 @@ variable "control_plane_nodepools" {
...
   }
+  validation {
+    condition = alltrue(
+      [for control_plane_nodepool in var.control_plane_nodepools : control_plane_nodepool.count >= 0]
+    )
+    error_message = "Count in control_plane_nodepools must be greater than 0."
+  }
+  validation {
+    condition = alltrue(
+      [for control_plane_nodepool in var.control_plane_nodepools : !control_plane_nodepool.use_placement_group || control_plane_nodepool.count <= 10]
```
aleksasiriski, line 200: This validation requires setting use_placement_group to false even with the global disable_placement_groups variable set to true. Validations can't access other variables, but this could be circumvented by moving it to a [check](https://developer.hashicorp.com/terraform/language/checks) instead of variable validation or removing the global option (but this should be done in v3, not in this PR)

aleksasiriski: I think it should remain in the validation and force users that have nodepools.count>10 to switch to using the nodepool level option as well.

valkenburg-prevue-ch (approved):
Looks good to me, under the condition that nodes do not get replaced if their placement group is modified in place.

valkenburg-prevue-ch (approved):
👍

aleksasiriski:
Hetzner doesn't apply new placement group settings unless the node is powered off. This leaves the nodes without any placement groups if they are changed while the nodes it powered on. This will be incorporated into v3.

mysticaltech (commented):
Let's go for v3

variables.tf
mysticaltech: @aleksasiriski Yes, seems like an acceptable constraint. But do what you feel is best.

PR #1225, opened 2024-02-23
add variable to enable Hubble relay and ui in Cilium

tobiasehlert:
Adding `cilium_hubble_enabled` variable with default to `false`.
When setting to `true`, Cilium Hubble relay and UI is enabled by adding parameters to default `cilium_values` variable.

### DIFF
--- a/kube.tf.example
+++ b/kube.tf.example
@@ -651,6 +651,9 @@ module "kube-hetzner" {
   # Enables egress gateway to redirect and SNAT the traffic that leaves the cluster. Default: false
   # cilium_egress_gateway_enabled = true
 
+  # Enables cilium's Hubble observability tool. Default: false
+  # cilium_hubble_enabled = true
+
   # You can choose the version of Calico that you want. By default, the latest is used.
   # More info on available versions can be found at https://github.com/projectcalico/calico/releases
   # Please note that if you are getting 403s from Github, it's also useful to set the version manually. However there is rarely a need for that!
--- a/variables.tf
+++ b/variables.tf
@@ -556,6 +556,12 @@ variable "cilium_egress_gateway_enabled" {
   description = "Enables egress gateway to redirect and SNAT the traffic that leaves the cluster."
 }
 
+variable "cilium_hubble_enabled" {
+  type        = bool
+  default     = false
+  description = "Enables Hubble observability in HelmChart values."
+}
+
 variable "cilium_ipv4_native_routing_cidr" {
   type        = string
   default     = null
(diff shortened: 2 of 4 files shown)


mysticaltech (approved):
Looking good, thanks @tobiasehlert!  

@M4t7e @Silvest89 Please review if you can, small but valuable PR.

tobiasehlert:
I didn't add fw-rule for this, so you need to run port forwards or so to access it. But I could add fw-rule similar to the ssh whitelist one for huddle if you'd like. :)

variables.tf
```diff
@@ -556,6 +556,12 @@ variable "cilium_egress_gateway_enabled" {
   description = "Enables egress gateway to redirect and SNAT the traffic that leaves the cluster."
 }
 
+variable "cilium_hubble_enabled" {
+  type        = bool
+  default     = false
+  description = "Enables Hubble observability in HelmChart values."
+}
```
M4t7e, line 563: Can you rename the setting to something like `cilium_hubble_ui_enabled`. Hubble is enabled by default, and the CLI is accessible with `kubectl exec -it -n kube-system ds/cilium -- hubble`. This setting activates the Hubble UI (and the relay required for this), not Hubble itself, which might mislead users into thinking it's required to enable Hubble. The description should also reflect this.

M4t7e:
> I didn't add fw-rule for this, so you need to run port forwards or so to access it. But I could add fw-rule similar to the ssh whitelist one for huddle if you'd like. :)

Enabling Hubble UI does not open any external ports by default. Hence, a firewall rule alone would not be adequate for exposure. Moreover, making it publicly accessible might not be advisable given the default settings, which allow unrestricted access to Hubble.

M4t7e (changes requested):

variables.tf
mysticaltech: @tobiasehlert If you could also mention how to enable Hubble via `kubectl exec -it -n kube-system ds/cilium -- hubble` and also include a link to the relevant docs in the comments, it would be great. So that users know exactly how to make this work. Along with any kubectl port-forwarding command that could be used please. We need to present a cohesive workflow and use-case 🙏

kube.tf.example
```diff
@@ -651,6 +651,9 @@ module "kube-hetzner" {
   # Enables egress gateway to redirect and SNAT the traffic that leaves the cluster. Default: false
   # cilium_egress_gateway_enabled = true
 
+  # Enables cilium's Hubble observability tool. Default: false
```
mysticaltech, line 654: @tobiasehlert If you could also mention how to enable Hubble via kubectl exec -it -n kube-system ds/cilium -- hubble and also include a link to the relevant docs in the comments, it would be great. So that users know exactly how to make this work. Along with any kubectl port-forwarding command that could be used please. We need to present a cohesive workflow and use-case 🙏

mysticaltech (changes requested):

tobiasehlert:
Thanks for the feedback @mysticaltech, well yes you can do this my writing your own `cilium_values`, but I thought it would be nicer to have some support for enabling cilium features similar to `cilium_routing_mode` or `cilium_egress_gateway_enabled`.

But sure, we can skip this and I'll maintain my other `cilium_values` instead :)

mysticaltech:
@tobiasehlert Yes. It's the UI exposure matter that complicated it and made it sensitive. Thanks for your understanding.

mysticaltech:
@tobiasehlert FYI this was merged today and will be released soon. #1305

tobiasehlert:
Ah nice @mysticaltech 👌
But what has the difference to my pull request one month ago where you decided no? Good to have the functionality in place, so thanks! 😁

mysticaltech:
Just the docs and narrative around the feature.

PR #1229, opened 2024-02-24
Allow submission of the new generated terraform docs via PR

mysticaltech:
Subsequent to #1227


mysticaltech:
Merging without review to test if it works.

Issue #1231, opened 2024-02-24
Autoscaler does not scale up. predicate checking error: Insufficient ephemeral-storage.


schlichtanders:
### Description

I just realized that my kube-hetzner cluster does not scale anylonger. As far as I remember it was actually working for a long time (could be that I just thought it was working, but never really saw it).

Inspecting journalctl logs I found that the key error seems to be inside systemd `k3s-agent.service`. It logs:
```
k3s-agent.service: Failed to locate executable /usr/local/bin/k3s: Permission denied
```

I made a `ls -l` to inspect the directory, showing
```
# ls -l /usr/local/bin
total 58068
lrwxrwxrwx. 1 root root        3 Feb 24 09:32 crictl -> k3s
lrwxrwxrwx. 1 root root        3 Feb 24 09:32 ctr -> k3s
-rwxr-xr-x. 1 root root 59441152 Feb 24 09:32 k3s
-rwxr-xr-x. 1 root root     1565 Feb 24 09:32 k3s-agent-uninstall.sh
-rwxr-xr-x. 1 root root     2274 Feb 24 09:32 k3s-killall.sh
lrwxrwxrwx. 1 root root        3 Feb 24 09:32 kubectl -> k3s
```
Which looks fine, doesn't it?

Googling shows this issue https://github.com/k3s-io/k3s/issues/5903  which links to a workaround `ln -s /data/k3s /var/lib/rancher/k3s`. I have no clue how this relates.

---------------

I am really buffled to see such critical bugs of the cluster because I thought I fixed almost any version I could find.
Hence two questions:
- how to fix this? ANSWER: upgrade to most recent kube-hetzner version
- which version do I need to fix where in order to prevent such problems appearing spontanuously? ANSWER: set `automatically_upgrade_k3s = false` in your kube.tf

### EDIT: New Problem, autoscaler still not working

After updating everything to the most recent kube-hetzner (including microos snapshots), the above error disappears, however a new one still prevents the autoscaling from working.
The autoscaler logs show something like:
```
fails because of `k3s-agent.service: Failed to locate executable /usr/local/bin/k3s: Permission denied
```

Which prevents the scale up.

<details>
<summary>

### Kube.tf file

</summary>

```terraform
locals {
  # You have the choice of setting your Hetzner API token here or define the TF_VAR_hcloud_token env
  # within your shell, such as such: export TF_VAR_hcloud_token=xxxxxxxxxxx
[...]

PR #1239, opened 2024-02-26
Support for v1.29

mysticaltech:
Made possible by #1231 and a fix to the autoscaler by @Silvest89  🚀

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -232,7 +232,7 @@ variable "cluster_autoscaler_image" {
 
 variable "cluster_autoscaler_version" {
   type        = string
-  default     = "20231027"
+  default     = "20240226"
   description = "Version of Kubernetes Cluster Autoscaler for Hetzner Cloud. Should be aligned with Kubernetes version"
 }
 
@@ -454,7 +454,7 @@ variable "enable_metrics_server" {
 
 variable "initial_k3s_channel" {
   type        = string
-  default     = "v1.28" # Please update kube.tf.example too when changing this variable
+  default     = "v1.29" # Please update kube.tf.example too when changing this variable
   description = "Allows you to specify an initial k3s channel."
 
   validation {
--- a/kube.tf.example
+++ b/kube.tf.example
@@ -108,7 +108,7 @@ module "kube-hetzner" {
   # Also, before decreasing the count of any nodepools to 0, it's essential to drain and cordon the nodes in question. Otherwise, it will leave your cluster in a bad state.
 
   # Before initializing the cluster, you can change all parameters and add or remove any nodepools. You need at least one nodepool of each kind, control plane, and agent.
-  # The nodepool names are entirely arbitrary, you can choose whatever you want, but no special characters or underscore, and they must be unique; only alphanumeric characters and dashes are allowed.
+  # ⚠️ The nodepool names are entirely arbitrary, but all lowercase, no special characters or underscore (dashes are allowed), and they must be unique.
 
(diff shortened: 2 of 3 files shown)


variables.tf
```diff
@@ -454,7 +454,7 @@ variable "enable_metrics_server" {
 
 variable "initial_k3s_channel" {
   type        = string
-  default     = "v1.28" # Please update kube.tf.example too when changing this variable
+  default     = "v1.29" # Please update kube.tf.example too when changing this variable
```
aleksasiriski, line 457: Please update kube.tf.example too when changing this variable

aleksasiriski (approved):

mysticaltech:
@Silvest89 Could you please take care of updating the autoscaler mother repo so that our fork does not diverge 🙏 Hopefully soon enough we can go back to using that and not have to maintain our own.

M4t7e:
Hey @mysticaltech, I want to make sure we maintain compatibility with our 3rd party packages. The developers behind these packages are diligently working to keep up with new Kubernetes releases, which often means addressing incompatibilities/bugs or implementing new features.

I collected the relevant minimum versions:
| K3s  | Autoscaler | Calico | Cilium | HCCM | HCSI | Kured |
|------|------------|--------|--------|------|------|-------|
| 1.26 | 1.26       | 3.25   | 1.13   | 1.14 | 2.2  | 1.12  |
| 1.27 | 1.27       | 3.26   | 1.14   | 1.16 | 2.4  | 1.13  |
| 1.28 | 1.28       | 3.25   | 1.15   | 1.18 | 2.5  | 1.14  |
| 1.29 | 1.29       | 3.27   | 1.15   | 1.19 | 2.6  | 1.15  |

Versions after this PR:
- **Autoscaler:** 20240226 -> 1.29 (OK)
- **Calico:** 3.26.4 (Not OK)
- **Cilium:** 1.15.1 (OK)
- **HCCM:** latest (OK)
- **HCSI:** latest (OK)
- **Kured:** latest (OK)

It seems only Calico needs an upgrade. Fortunately, version 3.27 is already available.
I'm a bit confused about our Autoscaler versioning, but I understood this version reflects 1.29.

For the packages where we currently use the latest version, I would propose (maybe at some point in future) to adopt a more stringent versioning strategy to guarantee compatibility without being the first ones hitting day 1 issues of fresh releases. I prefer stability over being the early adopter (or software tester ^^). Over time, we might also face backward compatibility issues, for instance, if a package makes a major upgrade that requires a strict minimum Kubernetes version. Currently, our setup appears to install it without considering these constraints.

mysticaltech:
@M4t7e  Thanks for the info, super important indeed! Good to hear that Calico 3.27 is available, I will bump it. 

Definitely, same here I prefer we use stable, but this issue #1231 initially led me to believe that we needed to move to v1.19 to fix the autoscaler issue. I later learned that autoscaler v1.29.x is in fact compatible with previous versions of kube too. But what's done is done.

About the autoscaler thing, what happened is that @Silvest89 contributed vital fixes and support for labelling Hetzner nodes if I remember correctly, it was merged on the mother repo kubernetes/autoscaler, but their release schedule is super long, so it just shipped a few weeks ago, in the meantime we had to fork our own (@Silvest89 did it) for users to benefit from a working version with Hetzner. Now same as before, a fix needs to be synced to the mother repo and it will take a long time to be released, so we are stuck with our own fork for a while longer. And yes, we always have the master of our fork updated from upstream via a bot.

Silvest89:
@mysticaltech I've already opened a PR

mysticaltech:
Thank you @Silvest89, really appreciate it man 🙏 Please keep up posted when merged  🤞

PR #1241, opened 2024-02-27
feat: add support for node-to-node encryption in cilium

tobiasehlert:
WireGuard-based encryption only encrypts traffic between Cilium-managed pods by default.

With this change Cilium also enables node-to-node encryption, which additionally also encrypts node-to-node, pod-to-node and node-to-pod traffic.

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -420,6 +420,7 @@ bpf:
 %{if var.enable_wireguard}
 encryption:
   enabled: true
+  nodeEncryption: true
   type: wireguard
 %{endif~}
 %{if var.cilium_egress_gateway_enabled}


mysticaltech:
Thanks for this @tobiasehlert, however as a rule thumb, we want to keep advanced use cases for the cilium_values variable.

tobiasehlert:
> Thanks for this @tobiasehlert, however as a rule thumb, we want to keep advanced use cases for the cilium_values variable.

I don't think this is something advanced as we have a flag for turning on encryption with wireguard, but that is only partial encryption. With the `nodeEncryption` we only extend the encryption slightly more without adding some advanced stuff :)

Silvest89:
To be honest, I like flexibility, and choice. What if I don't need/want node encryption ;]

tobiasehlert:
> To be honest, I like flexibility, and choice. What if I don't need/want node encryption ;]

Well, for flexibility and choice you have to write your own `cilium_values` value as @mysticaltech mentioned in some other PR.

I think that the readme saying "Optional Wireguard encryption of the Kube network for added security." should be reformulated to reflect that encryption only encrypts traffic between Cilium-managed pods and nothing else.

Anyhow, lets what @M4t7e thinks about this ^_^

M4t7e:
Thx for the PR, @tobiasehlert!

Generally, I see it as the better default for the `enable_wireguard` setting and likely more in line with the user expectation. When enabled, it ensures that all possible communications are encrypted via WireGuard. However, the addition of the `nodeEncryption` setting still selectively excludes certain types of traffic like some control plane scenarios to avoid deadlock situations.

My primary concern against this setting as the default for now is its beta status: [v1.15 Node-to-Node Encryption (beta)](https://docs.cilium.io/en/v1.15/security/network/encryption-wireguard/#node-to-node-encryption-beta).
I prefer our configuration to include only stable and officially supported settings. As soon as this leaves the beta status, I would tend to consider it as a nice default setting :slightly_smiling_face:

mysticaltech:
@M4t7e Thanks for the clarity on the matter. 

@tobiasehlert As we are working on v3, that PR will get hard to manage (while waiting for the feature to get out of beta) when the master branch changes, and it will change a lot. 

Please don't hesitate do add the FR to the [Ideas discussion](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/discussions/categories/ideas) section, so we do not lose track of it. Closing this for now. Thanks again!

PR #1263, opened 2024-03-05
Mention of KH Assistant Custom GPT

mysticaltech:
Added the mention to get this into the wild, it's too useful to not share.


mysticaltech:
@maggie44 It's not training per se, it knows about the project via prompting and RAG. It has all the docs and kube.tf, so you can ask it to gen one to you liking or paste one for mods, whatever works for you. Feedback welcome, it's still brand new.

PR #1269, opened 2024-03-11
Add instructions on running from Terraform Cloud

bverhagen:
I managed to make kube-hetzner work on Terraform Cloud. Since the ssh_private_key and ssh_public_key variables expect the file contents rather than the path to the files (as was the case in the past), the existing instructions are more convoluted then they need to be (provided users are ok with nog password protecting their ssh private key). I assume Terraform Cloud encrypts their sensitive variables in a different way already.


mysticaltech (approved):
Thanks @bverhagen !

PR #1270, opened 2024-03-12
Strip carriage return from ssh_client_identity

dviersel:
There are situations where the ssh key contains carriage return chars, for example when generating the key using 1password.

The change ensures that the carriage return character (`\r`) is removed from the ssh_client_identity variable using sed. This will prevent authentication issues during `terraform apply`.


mysticaltech (approved):
@dviersel Thanks for this.

PR #1273, opened 2024-03-12
Fix missing SELinux rules

mysticaltech:
Subsequent to #1272 and #1266


mysticaltech:
@patope Tested this and it works, you were right though, it wasn't applying, but it was because of the missing class definitions.

mysticaltech:
@kube-hetzner/core Merging as hotfix for #1266

Issue #1278, opened 2024-03-15
[Bug]: `cluster-autoscaler` does not wait long enough for new server to become available


UBaggeler:
### Description

Since midnight our autoscaler is having issues launching new nodes (from 1 --> 3). The creation of the new servers seems to take longer than the expected (default) timeout of 5min:

```
failed to create error: failed to start server hcloud-autoscaled-xyz error: timeout waiting for server hcloud-autoscaled-xyz
```

In this case `cluster-autoscaler` removes/deletes the nodes and repeatedly tries to spin up new servers, without success.

Manually increasing the timeout (for example to 15min) by setting the env variable `HCLOUD_SERVER_CREATION_TIMEOUT` on the autoscaler deployment resolves the issue. 

Unfortunately the `autoscaler.yaml.tpl` file does not allow to set this environment variable.

### Kube.tf file

```terraform
n/a
```

### Platform

Linux

mysticaltech:
@UBaggeler Your fix was just released as part of v2.13.4 🚀

PR #1279, opened 2024-03-15
autoscaler: Add support to set HCLOUD_SERVER_CREATION_TIMEOUT 

UBaggeler:
This PR fixes https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/1278 by exposing a variable to increase server creation timeout of `cluster-autoscaler`.


mysticaltech (approved):
Thanks @UBaggeler, which timeout would you put as default? 10 minutes?

mysticaltech:
I will add a default of 10, as I think 15 seems overkill.

PR #1280, opened 2024-03-15
Fix LoadBalancer service using both public and private IP of LB's 

ricristian:
When a new service of type LoadBalancer is requested k3s will use besides the public IPv4 & 6 of the Hetzner LB also the private ip of that LB which is wrong. 

This also fixes the external-dns updating A records with the private IP of the LB

With `HCLOUD_LOAD_BALANCERS_DISABLE_PRIVATE_INGRESS: true` ( this is the default value ) 
Screenshot 2024-03-15 at 21 04 19

external-dns will detect external load balancer IP's and will create following A records: ( cloudflare, but with other dns providers will be the same issue )
Screenshot 2024-03-15 at 21 07 40

With this new change:

With `HCLOUD_LOAD_BALANCERS_DISABLE_PRIVATE_INGRESS: false`
Screenshot 2024-03-15 at 21 05 12

With this new change external dns won't create the A record for the private iP

### DIFF
--- a/templates/ccm.yaml.tpl
+++ b/templates/ccm.yaml.tpl
@@ -27,3 +27,6 @@ spec:
               value: "true"
             - name: "HCLOUD_LOAD_BALANCERS_ENABLED"
               value: "${!using_klipper_lb}"
+            - name: HCLOUD_LOAD_BALANCERS_DISABLE_PRIVATE_INGRESS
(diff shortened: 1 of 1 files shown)


mysticaltech (approved):
Makes sense, thanks for the addition @ricristian.

Issue #1281, opened 2024-03-15
[Bug]: autoscaler nodes do not (allow to) set kubelet-args like kube-reserved and system-reserved


UBaggeler:
### Description

Nodes of autoscaler pools do not have `kube-reserved` / `system-reserved` reservations set. We lost a node today due to memory pressure (as a Pod was able to allocate so much memory that the system started to OOM other processes). 

It is currently only possible to set `kubelet-arg` for control plane and agent nodes but not for autoscaler nodes. 

### Kube.tf file

```terraform
n/a
```

### Platform

Linux

PR #1282, opened 2024-03-15
autoscaler: allow to set kubelet-arg for autoscaler nodes

UBaggeler:
This PR allows to set `kubelet` arguments (through `kubelet_args`) for each autoscaler pool individually or globally (`k3s_autoscaler_kubelet_args`).

Fixes https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/1281


mysticaltech (approved):
Awesome!

PR #1284, opened 2024-03-16
Tweak autoscaler creation timeout

mysticaltech:
Subsequent to #1282 and #1278. 

@UBaggeler FYI, small tweaks, especially moved from type string for the var to number. Thanks again for the help and contributions! Never hesitate 🙏


mysticaltech:
@kube-hetzner/core Merging as hotfix, as autoscaler are not deploying because of a timeout due to hetzner being slow today.

PR #1288, opened 2024-03-19
sepolicy changes for fluentbit

patope:
There policy changes are need to make fluentbit to work on cluster

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -790,7 +790,7 @@ EOF
     module kube_hetzner_selinux 1.0;
 
     require {
-      type kernel_t, bin_t, kernel_generic_helper_t, iscsid_t, iscsid_exec_t, var_run_t,
+      type kernel_t, bin_t, kernel_generic_helper_t, iscsid_t, iscsid_exec_t, var_run_t, var_lib_t,
       init_t, unlabeled_t, systemd_logind_t, systemd_hostnamed_t, container_t,
       cert_t, container_var_lib_t, etc_t, usr_t, container_file_t, container_log_t,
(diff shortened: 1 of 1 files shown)


locals.tf
```diff
@@ -860,10 +860,10 @@ EOF
...
     allow container_t container_log_t:file { open read watch };
     allow container_t container_log_t:lnk_file read;
-    allow container_t var_log_t:dir { add_name write };
-    allow container_t var_log_t:file { create lock open read setattr write };
-    allow container_t var_log_t:dir remove_name;
-    allow container_t var_log_t:file unlink;
+    allow container_t var_log_t:dir { add_name write watch read remove_name };
+    allow container_t var_log_t:file { create lock open read setattr write getattr unlink };
+    allow container_t var_lib_t:dir { add_name write read remove_name };
+    allow container_t var_lib_t:file { create lock open read setattr write getattr unlink };
```
mysticaltech, line 866: @patope var_log_t should not be removed.

mysticaltech (changes requested):

locals.tf
patope: Now there are 3 separate rows for `var_log_t:dir` and `var_log_t:file` permissions

aleksasiriski: Why not combine all the perms of the same class into single line? I see that wasn't done in the master branch, but maybe it should? Correct me if I'm wrong.

mysticaltech: I want to do that too, will fix.

mysticaltech (approved):
Thanks for this @patope, and sorry for the delay, was a busy week. Merging now.

PR #1300, opened 2024-03-27
feat: add delete protection to some hcloud resources

tobiasehlert:
This PR adds option to set delete protection flag on resources.

This can be helpful when wanting to protect resources to be deleted by misstake in the Hetzner Cloud Console or API. It can be used to protect resources from deletion by putting a "lock" on them.

Please note, that this does not protect deletion from Terraform itself, as the Provider will lift the lock in that case. The resources will only be protected from deletion via the Hetzner Cloud Console or API.

The three resources that have support in the Hetzner Cloud Provider that can be useful to start of are:

- Floating IPs
- Load Balancers
- Volumes (used by Longhorn)


mysticaltech (approved):
@tobiasehlert Awesome, thanks!

PR #1302, opened 2024-03-29
Various fixes

mysticaltech:
- Consolidated SELinux rules.
- Fix for system System upgrade controller via tagging the image explicitely and not relying on "latest"


mysticaltech:
@kube-hetzner/core Merging as hotfix, all tested.

PR #1305, opened 2024-03-30
feat: Add option to enable Hubble

DjVinnii:
This PR adds option to enable Hubble for the Cilium CNI without using `cilium_values`.


mysticaltech (changes requested):
@DjVinnii  Interesting, thanks for this. So we would access it via kubectl port-forward correct? If so, could you please create a tiny example in the examples section of the readme? That would make that feature complete.

DjVinnii:
> @DjVinnii  So we would access it via kubectl port-forward correct? If so, could you please create a tiny example in the examples section of the readme?

That is correct indeed! I added a small section to the readme to mention this.

mysticaltech (approved):
@DjVinnii Awesome, thanks!

PR #1330, opened 2024-04-28
Automatically set ssh_public_key and ssh_private_key from id_ed25519 local keypair

jaumebarber:
# Automatically set ssh_public_key and ssh_private_key from id_ed25519 local keypair

Provided that the user has previously generated ssh id_ed25519 keys within `ssh-keygen -t ed25519`, these will automatically be added to the cluster so that it is accessible right away from the local machine.


jaumebarber:
Mixed two PRs in here.

PR #1331, opened 2024-04-29
Automatically set ssh_public_key and ssh_private_key from local id_ed25519 keypair

jaumebarber:
# Automatically set ssh_public_key and ssh_private_key from local id_ed25519 keypair

Provided that the user has previously generated ssh id_ed25519 keys within `ssh-keygen -t ed25519`, these will automatically be added to the cluster so that it is accessible right away from the local machine.

### DIFF
--- a/modules/host/locals.tf
+++ b/modules/host/locals.tf
@@ -1,13 +1,15 @@
 locals {
+  ssh_public_key  = file("~/.ssh/id_ed25519.pub")
+  ssh_private_key = file("~/.ssh/id_ed25519")
   # ssh_agent_identity is not set if the private key is passed directly, but if ssh agent is used, the public key tells ssh agent which private key to use.
   # For terraforms provisioner.connection.agent_identity, we need the public key as a string.
-  ssh_agent_identity = var.ssh_private_key == null ? var.ssh_public_key : null
+  ssh_agent_identity = var.ssh_private_key == null ? local.ssh_private_key : var.ssh_public_key != null ? var.ssh_public_key : null
   # shared flags for ssh to ignore host keys for all connections during provisioning.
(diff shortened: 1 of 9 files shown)


mysticaltech (commented):
@jaumebarber  Appreciate this, it's an interesting idea. However, I'm not sure if it's best practice. @kube-hetzner/core  @aleksasiriski  @M4t7e @valkenburg-prevue-ch What's your thought on this folks?

modules/host/locals.tf
```diff
@@ -1,13 +1,15 @@
 locals {
+  ssh_public_key  = file("~/.ssh/id_ed25519.pub")
+  ssh_private_key = file("~/.ssh/id_ed25519")
   # ssh_agent_identity is not set if the private key is passed directly, but if ssh agent is used, the public key tells ssh agent which private key to use.
   # For terraforms provisioner.connection.agent_identity, we need the public key as a string.
-  ssh_agent_identity = var.ssh_private_key == null ? var.ssh_public_key : null
+  ssh_agent_identity = var.ssh_private_key == null ? local.ssh_private_key : var.ssh_public_key != null ? var.ssh_public_key : null
```
mysticaltech, line 6: @jaumebarber No private keys should be passed here under any circumstance. It's either null or the public key.

valkenburg-prevue-ch:
I think this should, if at all, be "opt in". I would not want terraform to start reading everyone's `~/.ssh/id_ed25519` and storing it in the plain-text state, without them explicitly choosing to do so. For me this is in fact the reason to just generate a key for each cluster, so if it gets leaked, only one cluster is compromised, not my entire internet presence.

Could it be that this desire comes from wanting to have a solution where you do not have to pass the explicit key to use for each ssh command? If so, please have a look at `~/.ssh/config`, where you can define which keys to use for which host, and much much more.

PR #1338, opened 2024-05-06
adds system_upgrade_enable_eviction flag

lennart:
on small clusters eviction of pods during system-upgrade of k3s might hang since pods cannot be scheduled on a different node during upgrade.

see the example plans of [system-upgrade-controller](https://github.com/rancher/system-upgrade-controller?tab=readme-ov-file#example-plans) as well


mysticaltech (approved):
Very useful indeed 🙏

PR #1347, opened 2024-05-14
Keep OS disk size

WebSpider:
Add variable to support keeping OS disk size the same when resizing nodes as discussed in https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/discussions/1344


mysticaltech (approved):
Awesome 👍

PR #1375, opened 2024-06-07
Update Default Server Types to reflect new pricing and server types

skyride:
Hetzner announced their new Intel cloud servers yesterday, CX22-CX52. They are much better value for money than the old Intel and even current AMD types while being almost exactly the same price. This PR updates the default templates to use CX22 and CX32 instances where CPX11 and CPX21 were used. This both reduces the cost and increases the performance of the default configuration.

https://www.hetzner.com/news/new-cx-plans/

This has been tested by following through the tutorial and creating a new cluster from scratch. CX22 is used to create the x86 snapshot, and also as the default type in the generated `kube.tf`.

Hetzner will discontinue the old Intel server types in 6 months.

### DIFF
--- a/README.md
+++ b/README.md
@@ -366,7 +366,7 @@ Example nodepool configuration:
 ```tf
 {
   name        = "egress",
-  server_type = "cpx11",
+  server_type = "cx22",
   location    = "fsn1",
   labels = [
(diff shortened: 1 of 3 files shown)


mysticaltech (approved):
Awesome! Thanks for that 🚀

PR #1395, opened 2024-06-29
Make kubeconfig_server_address configurable

### LINKED ISSUE #1390: [Feature Request]: Make kubeconfig_server_address configurable
### Description

Making [`kubeconfig_server_address`](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/eb39273101356a8c428e7c00684511ced4cffa24/kubeconfig.tf#L15C3-L15C17) configurable would be helpful in a setup without a control plane loadbalancer but with a common hostname pointing to the IP addresses of all control plane nodes, as described [here](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/eb39273101356a8c428e7c00684511ced4cffa24/kube.tf.example#L748-L750).

When using the IP address of the first control plane node as `kubeconfig_server_address`, as currently implemented, the cluster cannot be accessed using that kubeconfig in case the first node is down. However, the cluster would still be accessible if the kubeconfig contained the hostname pointing to the IP addresses of all control plane nodes.

kranich:
Closes #1390

### DIFF
--- a/kube.tf.example
+++ b/kube.tf.example
@@ -750,6 +750,11 @@ module "kube-hetzner" {
   # In which case, you need to define that hostname in the k3s TLS-SANs config to allow connection through it. It can be hostnames or IP addresses.
   # additional_tls_sans = ["cp.cluster.my.org"]
 
+  # If you create a hostname with multiple A records pointing to all of your
+  # control-plane nodes ips, you may want to use that hostname in the generated
+  # kubeconfig.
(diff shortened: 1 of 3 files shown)


mysticaltech:
@kranich Thanks for this, appreciate it! Did you test it and can confirm that it works properly?

kranich:
Yes, I have tested it successfully.

mysticaltech (approved):
Awesome!

Issue #1413, opened 2024-07-20
[Feature Request]: Configure all HelmChart deployments with version


schlichtanders:
### Description

Currently the version for some HelmCharts cannot be set

e.g.
- longhorn
- cert-manager
- ... probably more

For recreating the very same cluster it is essential to be able to fix these versions

mysticaltech:
@schlichtanders Very good you took this on, it's was a pain point I knew about but did not have time to focus on. Thanks!

PR #1415, opened 2024-07-20
Helmchart version and bootstrap

### LINKED ISSUE #1414: [Feature Request]: Enable configuration of "bootstrap" argument for HelmCharts
### Description

it might be a specific case, however I'd like to deploy kube-hetzner completely on the control-plane to safe resources on hetzner.

This is currently not possible because HelmCharts get only deployed on control plane nodes if the `bootstrap` flag is set. While this is true for `traefik`, it is not true for `cert-manager` (and `longhorn` in case it is activated, and probably some other as well).

It would be great to have this option configurable everywhere.

schlichtanders:
fixes #1414 and #1413 

builds upon https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/1411

### DIFF
--- a/init.tf
+++ b/init.tf
@@ -245,6 +245,8 @@ resource "null_resource" "kustomization" {
       {
         longhorn_namespace  = var.longhorn_namespace
         longhorn_repository = var.longhorn_repository
+        version             = local.longhorn_version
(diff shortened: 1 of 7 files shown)


mysticaltech (approved):
Wonderful

PR #1420, opened 2024-07-24
30m lock-ttl default added for kured_options

asafbennatan:
#1362 
added default value for kured_options.lock-ttl= 30m

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -756,7 +756,8 @@ kured_options = merge({
   "pre-reboot-node-labels" : "kured=rebooting",
   "post-reboot-node-labels" : "kured=done",
   "period" : "5m",
-  "reboot-sentinel" : "/sentinel/reboot-required"
+  "reboot-sentinel" : "/sentinel/reboot-required",
+  "lock-ttl" : "30m"
 }, var.kured_options)
 
 k3s_registries_update_script = <<EOF


mysticaltech (approved):
Awesome!

mysticaltech:
@asafbennatan Please if you could add the default values here, it would be better. Also please make sure to use terraform format to make sure it's fixes any spacing or formatting issues automatically. Thanks!

https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/master/locals.tf#L755-L761

asafbennatan:
done @mysticaltech

mysticaltech (approved):
Awesome

jr-dimedis:
@mysticaltech , @asafbennatan 

This is a very dangerous change and can lead to a faulty upgrade completely paralyzing the cluster.

It has been made before and was withdrawn for this very reason (see this commit: https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/commit/d1e5b49a3294eb3c43fa10b5e582a797f7f21dc5). I think the right “fix” here would be to explicitly set the value to 0 and add a comment as to why it should not be changed.

mysticaltech:
@jr-dimedis Appreciate the heads up, you are right, it's dangerous, will revert.

asafbennatan:
hi @jr-dimedis ,
point understood , but the problem is that currently this setup causes sporadic update failures without any reason.
regardless of the default value if not lock-ttl what should be the solution for this ?
this is the solution kured team suggested

jr-dimedis:
Hi @asafbennatan,

I think the solution for you is setting this value just in your local kube-hetzner tf file in the `kured_options` variable.

Changing the default value for _all_ kube-hetzner users is the culprit here :)

PR #1436, opened 2024-07-30
Allows specifying k3s version

gmautner:
Sometimes a cluster operator might prefer to specify the k3s version (e.g. `v1.29.6+k3s2`) instead of the channel (e.g. `v1.29`, `stable` etc.)

Reasons for that could be not wanting the system upgrade controller to run at unpredictable times, or having more fine-grained control over the running version.

This PR introduces a new variable `install_k3s_version` in `kube.tf`. If undefined, the present behavior is preserved such that current installations are not affected. However, when defined, this variable takes precedence over `initial_k3s_channel`.

The mechanism for upgrades in either case continues to be the system upgrade controller.

### DIFF
--- a/init.tf
+++ b/init.tf
@@ -128,6 +128,7 @@ resource "null_resource" "kustomization" {
     # Redeploy when versions of addons need to be updated
     versions = join("\n", [
       coalesce(var.initial_k3s_channel, "N/A"),
+      coalesce(var.install_k3s_version, "N/A"),
       coalesce(var.cluster_autoscaler_version, "N/A"),
       coalesce(var.hetzner_ccm_version, "N/A"),
       coalesce(var.hetzner_csi_version, "N/A"),
@@ -233,6 +234,7 @@ resource "null_resource" "kustomization" {
       "${path.module}/templates/plans.yaml.tpl",
       {
         channel          = var.initial_k3s_channel
+        version          = var.install_k3s_version
         disable_eviction = !var.system_upgrade_enable_eviction
     })
     destination = "/var/post_install/plans.yaml"
--- a/kube.tf.example
+++ b/kube.tf.example
@@ -590,6 +590,10 @@ module "kube-hetzner" {
   # The default is "v1.29".
   # initial_k3s_channel = "stable"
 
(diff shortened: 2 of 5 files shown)


gmautner:
By the way, the following scenarios were tested and worked according to expected:
- Created a new cluster with default settings and initial k3s channel was used
- Configured `install k3s version` on above cluster and plan was changed to refer to k3s version and executed properly
- Created a new cluster with `install_k3s_version` and it was installed with the desired k3s version
- Deleted `install_k3s_version` variable on above cluster, plan changed to use initial k3s channel instead and executed properly

mysticaltech (approved):
Looking good!

mysticaltech:
Sorry @gmautner, I had to revert the PR as it did not did not deploy on my test. If you manage to fix it with or without channel and with or without version, please don't hesitate to submit a new one.

Screenshot from 2024-08-09 03-36-09

gmautner:
I think this was caused by a network connectivity issue unrelated to my PR.

When the server https://update.k3s.io/v1-release/channels is down, the behaviour of the k3s installation script is to return the input channel instead of the release, because `curl` follows all redirects and uses the called URL when there are no redirects.

I manually tested and was able to reproduce the problem with the following commands:

```
version_url='https://update.k3s.io/v1-release/channels/v1.29'
curl -w '%{url_effective}' -L -s -S ${version_url} -o /dev/null | sed -e 's|.*/||'
v1.29.7+k3s1
```

Turning off the network:

```
curl -w '%{url_effective}' -L -s -S ${version_url} -o /dev/null | sed -e 's|.*/||'
curl: (6) Could not resolve host: update.k3s.io
v1.29
```

This behaviour has been documented by other users of kube-hetzner, like:

https://github.com/k3s-io/k3s/issues/10578#issuecomment-2257389559

So probably it has no relationship to the PR. Could you please test it again?

mysticaltech:
@gmautner Thanks for the info. This is a crazy coincidence because I've ever seen this very error before.

mysticaltech:
Ok, I see it's a new feature to allow specifying the version. Please let's wait a bit until it's more stable then we can try again.

PR #1463, opened 2024-08-27
autoscaler formatting documentation fix

Seesbastian:
There was an error in the documentatio about the example file.
Resulting in error when using autoscaler:
Error: Invalid expression                                                                                                                                                │                                                                                                                                                                         
│   on kube.tf line 333, in module "kube-hetzner":                                                                                                                        
│  333:       taints      =                                                                                                                                               
│  334:       [{                                                                                                                                                          
│                                                                                                                                                                         
│ Expected the start of an expression, but found an invalid expression token.
Solution: 
1. change position of "[" and "]" bracket
2. change format of listed taints: key="value", not key:"value"

### DIFF
--- a/kube.tf.example
+++ b/kube.tf.example
@@ -321,23 +321,24 @@ module "kube-hetzner" {
   #
   # * Example below:
   # autoscaler_nodepools = [
-  #   {
-  #     name        = "autoscaled-small"
-  #     server_type = "cx32"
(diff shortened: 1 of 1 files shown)


mysticaltech (approved):
Thanks for this  @Seesbastian

mysticaltech (approved):
Wonderful, thanks

Issue #1470, opened 2024-09-06
[Bug]: Terraform failing if csi is disabled (disable_hetzner_csi) and `hetzner_csi_version` not set


UBaggeler:
### Description

We are not using csi (disable_hetzner_csi = true) and therefore `hetzner_csi_version` is not set. Since a few weeks, Terraform apply is failing with the following error.

```
╷
│ Error: Error in function call
│ 
│   on .terraform/modules/kube-hetzner/init.tf line 258, in resource "null_resource" "kustomization":
│  258:     content = templatefile(
│  259:       "${path.module}/templates/hcloud-csi.yaml.tpl",
│  260:       {
│  261:         version = local.csi_version
│  262:         values  = indent(4, trimspace(var.hetzner_csi_values))
│  263:     })
│     ├────────────────
│     │ while calling templatefile(path, vars)
│     │ local.csi_version is null
│     │ path.module is ".terraform/modules/kube-hetzner"
│     │ var.hetzner_csi_values is ""
│ 
│ Call to function "templatefile" failed:
│ .terraform/modules/kube-hetzner/templates/hcloud-csi.yaml.tpl:10,15-22:
│ Invalid template interpolation value; The expression result is null. Cannot
│ include a null value in a string template..
```

### Kube.tf file

```terraform
[...]

  # To disable Hetzner CSI storage, you can set the following to "true", default is "false".
  disable_hetzner_csi = true

  # If you want to use a specific Hetzner CCM and CSI version, set them below; otherwise, leave them as-is for the latest versions.
  # hetzner_ccm_version = ""
  # hetzner_csi_version = ""

[...]
```

### Platform

Linux

PR #1417, opened 2024-07-20
Helmchart hetzner csi

schlichtanders:
fixes #1416 

I am not sure whether this works, because I just realised I never used the hetzner csi. 
but I think it should

builds upon #1415

PR #1471, opened 2024-09-06
fix hcloud-csi.yaml if csi is disabled (disable_hetzner_csi)

UBaggeler:
This PR fixes https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/1470 which was introduced with https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/1417. 

- If `disable_hetzner_csi = false` then `csi_version` will always be set (either from `hetzner_csi_version` or using the latest github release).
- If `disable_hetzner_csi = true` then `csi_version` will not be automatically be inferred from the latest github release. This is why rendering the `templatefile` fails, even though the file is not needed and ignored. This PR fixes rendering the template using `coalesce`. This way the file will be generated but is still not used since csi is disabled.

### DIFF
--- a/init.tf
+++ b/init.tf
@@ -253,12 +253,12 @@ resource "null_resource" "kustomization" {
     destination = "/var/post_install/longhorn.yaml"
   }
 
-  # Upload the csi-driver-smb config
+  # Upload the csi-driver config (ignored if csi is disabled)
   provisioner "file" {
     content = templatefile(
       "${path.module}/templates/hcloud-csi.yaml.tpl",
       {
-        version = local.csi_version
+        version = coalesce(local.csi_version, "N/A")
         values  = indent(4, trimspace(var.hetzner_csi_values))
     })
     destination = "/var/post_install/hcloud-csi.yaml"


olexiyb:
I came across the same issue
I did next just did not have time to create pull
https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/commit/1d5a6721d157ac436aa1b2bb2f38d23ae9900f85

init.tf
```diff
@@ -253,12 +253,12 @@ resource "null_resource" "kustomization" {
...
   }
 
-  # Upload the csi-driver-smb config
+  # Upload the csi-driver config (ignored if csi is disabled)
   provisioner "file" {
     content = templatefile(
       "${path.module}/templates/hcloud-csi.yaml.tpl",
       {
-        version = local.csi_version
+        version = coalesce(local.csi_version, "N/A")
```
mysticaltech, line 261: @UBaggeler Thanks for this. Please let do the edit at the locals level to local.csi_version directly.

mysticaltech (changes requested):

init.tf
mysticaltech: I will merge and do the adjustment myself. As this is an important fix.

lwd8cmd: @mysticaltech Has the fallback value for `csi_version` been added in another commit? I can confirm your suggested change works, but it's not included in the current release (2.15.4), as the issue still persists. If you're busy with other tasks, I can open a pull request with your proposed changes.

PR #1478, opened 2024-09-11
fix(hccm): custom args are ignored

### LINKED ISSUE #1477: [Bug]: Network routes are not cleaned up by HCCM
### Description

Network Routes are not properly cleaned up by hcloud-cloud-controller-manager since HCCM v1.20.0.

This causes issues when users cycle through their nodes (with autoscaler) and they have a lot of stale routes laying around. Adding new nodes then fails because they have already reached the routes limit (100).

HCCM only cleans up routes that are inside its configured `--cluster-cidr`.

In HCCM v1.20.0 we changed our pod spec from using `command` to `args`. Your patches (`templates/ccm.yaml.tpl`) still try to override the `command`, and are then ignored because the `ccm-networks.yaml` has the default `--cluster-cidr` in the `args`.

This was originally reported in https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/734

### Kube.tf file

```terraform
The issue can be reproduced with the default `kube.tf`.
```

### Screenshots

_No response_

### Platform

Linux

apricote:
Closes #1477

Fix the issue described by overriding `args` instead of `command`. In addition, I also added an argument (`--route-reconciliation-period=30s`) that was added in HCCM v1.14.2.

### DIFF
--- a/templates/ccm.yaml.tpl
+++ b/templates/ccm.yaml.tpl
@@ -9,11 +9,11 @@ spec:
     spec:
       containers:
         - name: hcloud-cloud-controller-manager
-          command:
-            - "/bin/hcloud-cloud-controller-manager"
+          args:
             - "--cloud-provider=hcloud"
             - "--leader-elect=false"
             - "--allow-untagged-cloud"
+            - "--route-reconciliation-period=30s"
             - "--allocate-node-cidrs=true"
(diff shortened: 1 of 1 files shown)


mysticaltech (approved):
Wonderful 🙏

PR #1480, opened 2024-09-12
Fix regeneration and rendering of hetzner_csi_values

WebSpider:
Using `hetzner_csi_values` didn't actually do anything. Now it does.


mysticaltech (approved):
@WebSpider  Good catch 🙏

Issue #1465, opened 2024-08-30
[Bug]: Cluster autoscaler will break after 6/9


WebSpider:
### Description

I've just received an email from Hetzner.
This relates to https://github.com/kubernetes/autoscaler/issues/7210

We should decide if we want to wait for this issue to be merged and bump the cluster-autoscaler version, or use the Hetzner provided temporary cluster-autoscaler version as mentioned here: https://docs.hetzner.cloud/changelog#2024-08-30-bug-cx11-removal-will-break-certain-versions-of-cluster-autoscaler

### Kube.tf file

```terraform
-
```

### Platform

Linux

apricote:
Hello :wave:,

existing users of the cluster-autoscaler should be save.

New users will get the error mentioned in the autoscaler issue. You can switch to the temporary images, or at least point new users at the possibility of using those to get cluster-autoscaler working during the weeks inbetween.

mysticaltech:
@Silvest89  Any recommendation on the best course of action for us here, please?

Silvest89:
> @Silvest89 Any recommendation on the best course of action for us here, please?
> 
> I think I will switch the the Hetzner provided images right away.

I think using the hetzner provided images would be the best course of action for now. Saves us from using our repo for it :P.

mysticaltech:
Fully agree! Will integrate it today.

mysticaltech:
Fixed in [v2.14.5](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/releases/tag/v2.14.5)

PR #1487, opened 2024-09-23
Switched the autoscaler image to hetzner's own

mysticaltech:
Emergency fix for #1465

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -264,13 +264,13 @@ variable "agent_nodepools" {
 
 variable "cluster_autoscaler_image" {
   type        = string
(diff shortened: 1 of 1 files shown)


asemarian:
I just got an email from Hetzner:

> If you were using the alternative container images we provided ([docker.io/hetznercloud/cluster-autoscaler](http://docker.io/hetznercloud/cluster-autoscaler)), please switch back to the images built by the Kubernetes community ([registry.k8s.io/autoscaling/cluster-autoscaler](http://registry.k8s.io/autoscaling/cluster-autoscaler)). We will remove the alternative container images on 19 November 2024.

So these alternative images will be removed soon.

PR #1502, opened 2024-10-14
Fix: Output public ipv6 on single-node-clusters

mattanja:
For testing purposes I started the implementation of my cluster setup with a single-node cluster and added Hetzner DNS Api to update a domain to the cluster ingress ip. However for the single node setup, the IPv6 was not set, because the conditions were not met.

This PR fixes setting the IPv6 Addresses in the module output for single-node clusters (that do not have an external loadbalancer).

### DIFF
--- a/output.tf
+++ b/output.tf
@@ -28,13 +28,13 @@ output "agents_public_ipv4" {
 }
 
 output "ingress_public_ipv4" {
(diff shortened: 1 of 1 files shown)


aleksasiriski (approved):
LGTM

mysticaltech (approved):
Wonderful

PR #1516, opened 2024-11-05
chore: bump system-upgrade-controller version to 0.14.2

pat-s:
Should this be updated automatically in the future by dependabot/renovate?


mysticaltech (changes requested):
Thanks @pat-s, please also update the url for the manifest to use the latest. It was set in the past as there was an issue, but was never touched again.

https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/a5a1f7e3129d212afc53031810b1112039d1e93a/locals.tf#L79

mysticaltech (approved):
It's good, will do the url change myself. Thanks for this!

PR #1512, opened 2024-11-05
Add s3 region to docs & move system-upgrade-controller to variables

ricristian:

PR #1506, opened 2024-10-23
Fix: Revert back to Kubernetes community autoscaler

BrammyS:
Closes #1505

PR #1534, opened 2024-11-07
Update doc, default values and kube.tf with recent changes to cluster_autoscaler_version, initial_k3s_channel and var sys_upgrade_controller_version

janfrederik:
#1512 introduced a new var `sys_upgrade_controller_version`.
- Corrected the description, which was an incorrect copy-paste from another var.
- Added the var to kube.tf.example

#1506 reverted cluster-autoscaler back to the community version. However, the default version for var `cluster_autoscaler_version` (1.31.1), does not correspond to the new default `initial_k3s_channel` ("stable", which is currently still 1.30) and the doc says these versions *should* correspond. So I put version 1.30.3 as default. This is in line with the initial_k3s_channel and includes the patch for the Hetzner deprecation of CX11.

Added `csi_driver_smb_version` to kube.tf.example

### DIFF
--- a/kube.tf.example
+++ b/kube.tf.example
@@ -423,6 +423,8 @@ module "kube-hetzner" {
 
   # To enable Hetzner Storage Box support, you can enable csi-driver-smb, default is "false".
   # enable_csi_driver_smb = true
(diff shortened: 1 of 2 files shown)


mysticaltech (approved):
Thanks @janfrederik, appreciate this very much.

PR #1540, opened 2024-11-09
Updated maximum volume size to 10240

fbrettnich:
The maximum size for volumes is 10,240 GB
https://docs.hetzner.cloud/#volumes-create-a-volume


mysticaltech (approved):
Thanks @fbrettnich

PR #1543, opened 2024-11-12
Update versions for microos snapshot

pat-s:
@mysticaltech Can these be taken care of by dependabot? Mainly the `selinux` one, the other one is only due every two years.

### DIFF
--- a/packer-template/hcloud-microos-snapshots.pkr.hcl
+++ b/packer-template/hcloud-microos-snapshots.pkr.hcl
@@ -56,7 +56,7 @@ locals {
     transactional-update --continue shell <<- EOF
     setenforce 0
     rpm --import https://rpm.rancher.io/public.key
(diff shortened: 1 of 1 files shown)


mysticaltech (approved):

mysticaltech:
Looking good! It changes rarely. If dependabot can do it why not, but not sure it supports that kind of updates. It's already active in the repo normally.

pat-s:
I am saying this in the hope that there is motivation to keep *all* dependencies of this module up-to-date. I myself don't use dependabot but renovate, which can be configured to also check versions in files living in subdirs in any file.

I understand that it is hard to keep everything up-to-date if not all versions are located in a single file. However, I think it is important for such projects as otherwise this is an (easy) entrypoint for CVE or other failures related to leaving certain deps behind.

mysticaltech:
You are totally right @pat-s. Will looking into Renovate.

janfrederik:
Thanks, @pat-s! A few days ago I just made the same changes in my fork to propose a pull request ;-)

PR #1546, opened 2024-11-15
Allow cluster with empty agent pool

janfrederik:
I only want autoscaler nodes, no fixed agent pools.

Currently, the count logic in the validation for agent_nodepools errors if agent_nodepools is empty (which is the default). The validation doesn't _fail_, it gives an _error_. The logic for calculating local.agent_count errors the same way.

However, there is not reason why empty agent pools wouldn't work. I used this patch a while in different combinations (single and multiple control planes, with our without autoscaler_nodepools) and everything looks great.


RRosalia:
Also voting for this one!

PR #1552, opened 2024-11-15
Expose controle plane load balancer ipv4 and ipv6

cedric-lovit:
`ingress_public_ipv4` and `ingress_public_ipv6` are exposed, but the controle plane load balancer ipv4 and ipv6 are not.

### DIFF
--- a/output.tf
+++ b/output.tf
@@ -37,6 +37,17 @@ output "ingress_public_ipv6" {
   value       = local.has_external_load_balancer ? module.control_planes[keys(module.control_planes)[0]].ipv6_address : (var.load_balancer_disable_ipv6 ? null : hcloud_load_balancer.cluster[0].ipv6)
 }
 
+output "lb_controle_plane_ipv4" {
+  description = "The public IPv4 address of the Hetzner controle plane load balancer"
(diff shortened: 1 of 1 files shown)


output.tf
```diff
@@ -37,6 +37,17 @@ output "ingress_public_ipv6" {
   value       = local.has_external_load_balancer ? module.control_planes[keys(module.control_planes)[0]].ipv6_address : (var.load_balancer_disable_ipv6 ? null : hcloud_load_balancer.cluster[0].ipv6)
 }
 
+output "lb_control_plane_ipv4" {
+  description = "The public IPv4 address of the Hetzner control plane load balancer"
+  value       = var.use_control_plane_lb ? hcloud_load_balancer.control_plane[0].ipv4 : null
```
aleksasiriski, line 42: You can make this nicer with
```terraform
value = one(hcloud_load_balancer.control_plane[*].ipv4)
```

Same thing for ipv6

aleksasiriski (dismissed):
LGTM but I would prefer it simpler with `one()`

output.tf
cedric-lovit: Updated!

aleksasiriski (approved):

PR #1564, opened 2024-11-20
Fix rename interface

schlichtanders:
works around #1560 

This makes the interface renaming failure tolerant. Worked for me with most recent autoscaler.
Still it would be even better if the network interface eth0 wouldn't be disconnecting and reconnecting initially

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -845,34 +845,35 @@ cloudinit_write_files_common = <<EOT
     ip link set $INTERFACE name eth1
     ip link set eth1 up
 
-    # wait till network interface becomes available
     myrepeat () {
-        local END_SECONDS
-        local tmp
-        END_SECONDS=$((SECONDS + 300))  # Current time + 300 seconds (5 minutes)
+        # Current time + 300 seconds (5 minutes)
+        local END_SECONDS=$((SECONDS + 300))
         while true; do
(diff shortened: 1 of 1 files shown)


schlichtanders:
after some test runs, this unfortunatly is not enough - sometimes even after 3 minutes the network is not available... there is hopefully a solution to this somewhere

M4t7e:
I haven't encountered this personally, but I've read about others facing similar issues with the internal network interface connections to Hetzner's network. Are the issues occurring with both eth0 and eth1, or just one?

To ensure the configuration is as fault-tolerant as possible, I suggest waiting for the network interfaces to be fully up before proceeding with nmcli after this: https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/d2d4423e8734ad236b9d13d78163745d43408f1c/locals.tf#L848

Example snippet to wait for `up` state:
```bash
until [ "$(cat /sys/class/net/eth0/operstate)" = "up" ]; do
    echo  "Waiting for eth0 to be up"
    sleep 2
done

until [ "$(cat /sys/class/net/eth1/operstate)" = "up" ]; do
    echo  "Waiting for eth1 to be up"
    sleep 2
done
```

If the problem persists after this, there may be an underlying issue or race condition with nmcli.

PR #1567, opened 2024-11-22
Feature: allow the use of private networks

xavierleune:
Hi there,

First, thank you for this awesome project! :raised_hands:
This feature allows you to deploy k3s using only private IPs on your servers. It requires a VPN and NAT configuration, which are out of scope for this deployment.

This is my very first time working with Terraform, so I'm open to any feedback on this pull request as it can probably be improved.

Have a nice day!

CF #282 #1255


autoscaler-agents.tf
```diff
@@ -165,7 +165,7 @@ resource "null_resource" "autoscaled_nodes_registries" {
     user           = "root"
     private_key    = var.ssh_private_key
     agent_identity = local.ssh_agent_identity
-    host           = each.value.ipv4_address
+    host           = coalesce(each.value.ipv4_address, each.value.ipv6_address, one(each.value.network).ip)
```
cedric-lovit, line 168: `one` seems to return null if there is no network value defined, so it will complain about `null.ip` right?

autoscaler-agents.tf
```diff
@@ -62,7 +62,7 @@ resource "null_resource" "configure_autoscaler" {
     user           = "root"
     private_key    = var.ssh_private_key
     agent_identity = local.ssh_agent_identity
-    host           = module.control_planes[keys(module.control_planes)[0]].ipv4_address
+    host           = coalesce(module.control_planes[keys(module.control_planes)[0]].ipv4_address, module.control_planes[keys(module.control_planes)[0]].ipv6_address, module.control_planes[keys(module.control_planes)[0]].private_ipv4_address)
```
cedric-lovit, line 65: can't we use the `local.first_control_plane_ip ` here?

modules/host/main.tf
```diff
@@ -30,6 +39,19 @@ resource "hcloud_server" "server" {
   backups            = var.backups
   user_data          = data.cloudinit_config.config.rendered
   keep_disk          = var.keep_disk_size
+  public_net {
+    ipv4_enabled = var.disable_ipv4 == true ? false : true
```
cedric-lovit, line 43: Not 100% sure, but can't you put `!var.disable_ipv4` here instead of the tenary operator?

cedric-lovit (commented):
I'm super excited for this feature. So thank you very much for putting effort in this!
You mentioned this was the first time you used terraform... You did super well!

Welcome to Terraform world! :D

autoscaler-agents.tf
xavierleune: I don't really know how to be sure,  I don't think your network can be null, as it will contains the private ip, which cannot be removed

modules/host/main.tf
xavierleune: it works ! Much more concise :+1:

nicolaracco:
I attempted to switch to this implementation in an existing cookbook without making any other change to the kube.tf file. I’ve posted the current configuration [here](https://gist.github.com/nicolaracco/36f3582c070569f7c1c79fc9ec7b879b). However, I encountered several instances of the following error:
```
Error: Attempt to get attribute from null value
│ 
│   on .terraform/modules/kube-hetzner/modules/host/out.tf line 10, in output "private_ipv4_address":
│   10:   value = one(hcloud_server.server.network).ip
│     ├────────────────
│     │ hcloud_server.server.network is empty set of object
│ 
│ This value is null, so it does not have any attributes.
╵
```

Maybe is it expecting to always find the network referenced as an input? If I have some time later today, I’ll try to debug the issue and add more info

KristijanL:
could this be expanded, so you could define DNS IP on the nodes, as my router VM that have external IP and nat configured, has dns resolver, for my internal stuff?

xavierleune:
@nicolaracco I confirm that the network should always be defined. It must be created before your cluster, because you have to be connected to this network through a non-k8s server.

xavierleune:
@mysticaltech do you have an opinion on this PR ? :pray:

mysticaltech:
@xavierleune Thanks for this, will review it tonight

4erdenko:
> @xavierleune Thanks for this, will review it tonight

Hey, any updates?
Disabling public addresses on nodes its a important security thing.

autoscaler-agents.tf
cedric-lovit: Ok!

mkajander:
Tested this today (along with NAT + Wireguard) and was able to successfully deploy a cluster with public ips disabled for the nodes.

xavierleune:
@mysticaltech thanks for your feedback, I'll have a look !

README.md
```diff
@@ -993,6 +993,30 @@ enable_delete_protection = {
...
+4. A route in your network, destination: `0.0.0.0/0` through the private ip of your machine with NAT.
+5. Make sure the connexion to your vpn is established before launching terraform.
+
+Recommended values:
+- Network range: `10.0.0.0/8`
+- Subnet for your wireguard and NAT machine: `10.128.0.0/16`
+
+If you follow this values, in your kube.tf, please set:
+- `existing_network_id = [YOURID]` (with the brackets)
+- `network_ipv4_cidr = "10.0.0.0/9"`
```
lefterisALEX, line 1013: small typo here? should be "10.0.0.0/8"

xavierleune: Actually you need to leave a subnet to machines out of k8s, like your wireguard instance. So you can give `10.0.0.0/9` to k8s and use `10.128.0.0/16` for your non-k8s instances

Mikopet:
This is something I am also really looking forward to. 

I have never done k8s cluster with publicly available nodes, and made me concerned. In fact, I was thinking about creating an own module because of that. But having this here would be way better :-)

**Big thanks for the implementation!**

xavierleune:
@mysticaltech I think that should do the trick, your test file now works as expected

devops-fd:
Super excited to get this live, when it's expected to merge this?

devops-fd:
Hello @xavierleune I have tried to run your branch and got some issues with post creation script as shown in the console
```
module.kube-hetzner.null_resource.first_control_plane: Provisioning with 'remote-exec'...
module.kube-hetzner.null_resource.first_control_plane (remote-exec): Connecting to remote host via SSH...
module.kube-hetzner.null_resource.first_control_plane (remote-exec):   Host: 10.127.128.101
module.kube-hetzner.null_resource.first_control_plane (remote-exec):   User: root
module.kube-hetzner.null_resource.first_control_plane (remote-exec):   Password: false
module.kube-hetzner.null_resource.first_control_plane (remote-exec):   Private key: true
module.kube-hetzner.null_resource.first_control_plane (remote-exec):   Certificate: false
module.kube-hetzner.null_resource.first_control_plane (remote-exec):   SSH Agent: false
module.kube-hetzner.null_resource.first_control_plane (remote-exec):   Checking Host Key: false
module.kube-hetzner.null_resource.first_control_plane (remote-exec):   Target Platform: unix
module.kube-hetzner.null_resource.first_control_plane (remote-exec): Connected!
module.kube-hetzner.null_resource.first_control_plane (remote-exec): + /etc/cloud/rename_interface.sh
module.kube-hetzner.null_resource.first_control_plane: Still creating... [10s elapsed]
module.kube-hetzner.null_resource.first_control_plane (remote-exec): loop
module.kube-hetzner.null_resource.first_control_plane (remote-exec): Error: Device 'eth0' not found.
module.kube-hetzner.null_resource.first_control_plane (remote-exec): Error: unknown connection ''.
module.kube-hetzner.null_resource.first_control_plane (remote-exec): got failure exit code, repeating
module.kube-hetzner.null_resource.first_control_plane (remote-exec): loop
```
Since disabling the public interface will result in a missing eth0
image

So changes must be made to locals.tf line 844 script to take this case into consideration

devops-fd:
I bypassed the error by removing the renaming of eth0, after I encountered another issue with k3s-agent not starting on agents nodes with this log
```
"2025-01-27T18:56:28Z" level=info msg="Network policy controller waiting for removal of node.cloudprovider.kubernetes.io/uninitialized taint"
"2025-01-27T18:56:30Z" level=info msg="Network policy controller waiting for removal of node.cloudprovider.kubernetes.io/uninitialized taint"
"2025-01-27T18:56:32Z" level=info msg="Network policy controller waiting for removal of node.cloudprovider.kubernetes.io/uninitialized taint"
"2025-01-27T18:56:34Z" level=info msg="Network policy controller waiting for removal of node.cloudprovider.kubernetes.io/uninitialized taint"
"2025-01-27T18:56:36Z" level=info msg="Network policy controller waiting for removal of node.cloudprovider.kubernetes.io/uninitialized taint"
```

locals.tf
```diff
@@ -873,11 +891,12 @@ cloudinit_write_files_common = <<EOT
...
-        connection.interface-name $eth
+        local eth="$1"
+        local eth_connection
+        eth_connection=$(nmcli -g GENERAL.CONNECTION device show "$eth" || echo '')
+        nmcli connection modify "$eth_connection" \
+          con-name "$eth" \
+          connection.interface-name "$eth"
     }
 
     myrepeat myrename eth0
```
gePower, line 902: I agree with @lefterisALEX but the condition should be here because the line 866 at 870 is a duplication of the myrename function without input varaible

```suggestion
    # In case of a private-only network, eth0 may not exist
    if ip link show eth0 &>/dev/null; then
      myrepeat myrename eth0
    fi
```

gePower (commented):
thanks for this code

kube.tf.example
```diff
@@ -150,6 +154,10 @@ module "kube-hetzner" {
 
       # Enable automatic backups via Hetzner (default: false)
       # backups = true
+      
+      # To disable public ips (default: false)
+      # disable_ipv4 = true
+      # disable_ipv6 = true
     },
```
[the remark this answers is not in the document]
xavierleune: Yes I think that should do it 👍

modules/host/main.tf
```diff
@@ -63,13 +85,11 @@ resource "hcloud_server" "server" {
...
-          echo "Waiting for MicroOS to become available..."
-          sleep 3
-        done
-      EOF
+      until ssh ${local.ssh_args} -i /tmp/${random_string.identity_file.id} -o ConnectTimeout=2 -p ${var.ssh_port} root@${coalesce(self.ipv4_address, self.ipv6_address, try(one(self.network).ip, null))} true 2> /dev/null
+      do
+        echo "Waiting for MicroOS to become available..."
+        sleep 3
+      done
     EOT
```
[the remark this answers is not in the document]
xavierleune: fixed, thanks ; I also added a note in README to install `coreutils` on macos to get timeout

README.md
```diff
@@ -82,6 +82,7 @@ brew install hashicorp/tap/terraform # OR brew install opentofu
 brew install hashicorp/tap/packer
 brew install kubectl
 brew install hcloud
+brew install coreutils # For macos
```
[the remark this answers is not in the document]
xavierleune: Ok it was already here on master

xavierleune:
This one might be the one 🤞

amedeopalopoli:
> Thanks @xavierleune, on it, will do one final review and we can merge this.

Hi all,

in Hetzner Private Networks I think we should add the below command in the init
`ip route add default via ${network-gateway}`

As well as we did [here](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/1567#issuecomment-2617826120) I think we should add a condition when we get Network Manager up.

```
cloudinit_runcmd_common = <<EOT
....
# Make sure the network is up
- [systemctl, restart, NetworkManager]
- [systemctl, status, NetworkManager]
- [ip, route, add, default, via, '172.31.1.1', dev, 'eth0']
....
EOT
```

In this case I think we need to add default network gateway if public ip address is not enabled

xavierleune:
@amedeopalopoli thanks for the suggestion, however I'm not sure we should do that, actually without public ip we should add a route for 0.0.0.0/0 in hetzner to the private ip of your gateway ; I followed [this tutorial](https://community.hetzner.com/tutorials/how-to-set-up-nat-for-cloud-networks) as I pointed out in the readme. I think that trying to use hetzner gateway without public ip would fail.
The setup I propose in the readme works for me since 3 months and I recreated a control plane yesterday that has access to net. So I'm not sure how to deal with your suggestion ^^

lefterisALEX:
An example i have it working in my case.
Subnet : 192.168.100.0/24
NAT_GW IP: 192.168.100.2/24 (The IP of the instance i use as NAT Gateway)

the on the instance (running ubuntu) that has only private interface i do this:
```
    cat <<'EOF' >> /etc/systemd/network/10-enp7s0.network
    [Match]
    Name=enp7s0
    
    [Network]
    DHCP=yes
    Gateway=192.168.100.1
    EOF
    echo "DNS=9.9.9.9" | tee -a /etc/systemd/resolved.conf
    systemctl reload systemd-networkd
    systemctl restart systemd-resolved
```

Not 100% sure if i missed something , but i am happy to help if needed

xavierleune:
@lefterisALEX thanks for the feedback, seems a lot like the approach I was talking about, BTW I think you might get rid of those steps by adding the route in hetzner routes for your subnet ; it should be automatic then, like this but with my gateway ^^
Capture d’écran 2025-02-27 à 10 47 08

aru-amedeo:
> use hetzner gateway without public ip w

I'm sorry @xavierleune but if we have nodes with only private IPs they cannot reach out internet (0.0.0.0) unless you add  
`- [ip, route, add, default, via, '$gateway-network', dev, 'private-interface']`

am I wrong?

xavierleune:
> > use hetzner gateway without public ip w
> 
> I'm sorry @xavierleune but if we have nodes with only private IPs they cannot reach out internet (0.0.0.0) unless you add `- [ip, route, add, default, via, '$gateway-network', dev, 'private-interface']`
> 
> am I wrong?

You can, if you add your gateway in the routes for 0.0.0.0/0, like in my previous screenshot ; that will leverage nat masquerade

lefterisALEX:
i think you need both @xavierleune . If you see in the documentation you shared in [Step 3 - Configuring NAT](https://community.hetzner.com/tutorials/how-to-set-up-nat-for-cloud-networks#step-3---configuring-nat) and  [Step 4 - Achieving a persistent configuration](https://community.hetzner.com/tutorials/how-to-set-up-nat-for-cloud-networks#step-4---achieving-a-persistent-configuration) it says that configuration is needed on client side as well.
For example for ubuntu:

screenshot -2025-02-27 -- 12 34 14@2x

screenshot -2025-02-27 -- 12 35 56@2x

xavierleune:
@lefterisALEX ok I can see that but from my understanding 10.0.0.1 is already declared because it's the subnet gateway for hetzner. I think cloud-init already handles that part.

Here is an example for `ip route show` for a control plane created yesterday:
```
default via 10.0.0.1 dev eth1 proto dhcp src 10.127.128.101 metric 100
10.0.0.0/8 via 10.0.0.1 dev eth1 proto dhcp src 10.127.128.101 metric 100
10.0.0.1 dev eth1 proto dhcp scope link src 10.127.128.101 metric 100
10.42.0.0/24 dev cni0 proto kernel scope link src 10.42.0.1
10.42.2.0/24 via 10.42.2.0 dev flannel.1 onlink
10.42.3.0/24 via 10.42.3.0 dev flannel.1 onlink
10.42.4.0/24 via 10.42.4.0 dev flannel.1 onlink
10.42.5.0/24 via 10.42.5.0 dev flannel.1 onlink
10.42.6.0/24 via 10.42.6.0 dev flannel.1 onlink
10.42.7.0/24 via 10.42.7.0 dev flannel.1 onlink
10.42.8.0/24 via 10.42.8.0 dev flannel.1 onlink
10.42.9.0/24 via 10.42.9.0 dev flannel.1 onlink
169.254.169.254 via 10.0.0.1 dev eth1 proto dhcp src 10.127.128.101 metric 100
```

aru-amedeo:
> @lefterisALEX ok I can see that but from my understanding 10.0.0.1 is already declared because it's the subnet gateway for hetzner. I think cloud-init already handles that part.
> 
> Here is an example for `ip route show` for a control plane created yesterday:
> 
> ```
> default via 10.0.0.1 dev eth1 proto dhcp src 10.127.128.101 metric 100
> 10.0.0.0/8 via 10.0.0.1 dev eth1 proto dhcp src 10.127.128.101 metric 100
> 10.0.0.1 dev eth1 proto dhcp scope link src 10.127.128.101 metric 100
> 10.42.0.0/24 dev cni0 proto kernel scope link src 10.42.0.1
> 10.42.2.0/24 via 10.42.2.0 dev flannel.1 onlink
> 10.42.3.0/24 via 10.42.3.0 dev flannel.1 onlink
> 10.42.4.0/24 via 10.42.4.0 dev flannel.1 onlink
> 10.42.5.0/24 via 10.42.5.0 dev flannel.1 onlink
> 10.42.6.0/24 via 10.42.6.0 dev flannel.1 onlink
> 10.42.7.0/24 via 10.42.7.0 dev flannel.1 onlink
> 10.42.8.0/24 via 10.42.8.0 dev flannel.1 onlink
> 10.42.9.0/24 via 10.42.9.0 dev flannel.1 onlink
> 169.254.169.254 via 10.0.0.1 dev eth1 proto dhcp src 10.127.128.101 metric 100
> ```

Hi @xavierleune as far as I understood, when you create a private network, you should add [this](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/1567#issuecomment-2687587183). This allows you to use the network route when the destination address is part of the internal network and when it does not (e.g. public ip like 8.8.8.8). This is why I said you this PR is not enough to achieve what you aim to do. It's missing the cloudinit stuff to configure the default gw when a public ip should be routable from nodes with no public interface attached to.

amedeopalopoli:
> @lefterisALEX ok I can see that but from my understanding 10.0.0.1 is already declared because it's the subnet gateway for hetzner. I think cloud-init already handles that part.
> 
> Here is an example for `ip route show` for a control plane created yesterday:
> 
> ```
> default via 10.0.0.1 dev eth1 proto dhcp src 10.127.128.101 metric 100
> 10.0.0.0/8 via 10.0.0.1 dev eth1 proto dhcp src 10.127.128.101 metric 100
> 10.0.0.1 dev eth1 proto dhcp scope link src 10.127.128.101 metric 100
> 10.42.0.0/24 dev cni0 proto kernel scope link src 10.42.0.1
> 10.42.2.0/24 via 10.42.2.0 dev flannel.1 onlink
> 10.42.3.0/24 via 10.42.3.0 dev flannel.1 onlink
> 10.42.4.0/24 via 10.42.4.0 dev flannel.1 onlink
> 10.42.5.0/24 via 10.42.5.0 dev flannel.1 onlink
> 10.42.6.0/24 via 10.42.6.0 dev flannel.1 onlink
> 10.42.7.0/24 via 10.42.7.0 dev flannel.1 onlink
> 10.42.8.0/24 via 10.42.8.0 dev flannel.1 onlink
> 10.42.9.0/24 via 10.42.9.0 dev flannel.1 onlink
> 169.254.169.254 via 10.0.0.1 dev eth1 proto dhcp src 10.127.128.101 metric 100
> ```

Hi @xavierleune as far as I understood, when you create a private network, you should add https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/1567#issuecomment-2687587183. This allows you to use the network route when the destination address is part of the internal network and when it does not (e.g. public ip like 8.8.8.8). This is why I said you this PR is not enough to achieve what you aim to do. It's missing the cloudinit stuff to configure the default gw when a public ip should be routable from nodes with no public interface attached to.

xavierleune:
Hi, just to let you know as I was unsure about the next steps I opened a ticket at hetzner and they confirmed what @aru-amedeo said about the public gateway. I don't really understand how things was working without that (gosh I need to work more on network :D). I'll take some time to do that in the coming days 👋

mysticaltech:
@xavierleune I really want to merge this, it's been going on for way too long and it's a needed feature, let me know when you think it's good to go, the second I get your final green light I will merge.

xavierleune:
green-light-dank

@mysticaltech this is not an exercice, you're good to go !
I've rebased my branch and added the default route as 10.0.0.1 when no public ip is available

Thank you all for your help !

mysticaltech:
Thanks @xavierleune , awesome work! 🙏

mysticaltech (approved):

autoscaler-agents.tf
```diff
@@ -171,7 +171,7 @@ resource "null_resource" "autoscaled_nodes_registries" {
     user           = "root"
     private_key    = var.ssh_private_key
     agent_identity = local.ssh_agent_identity
-    host           = each.value.ipv4_address
+    host           = coalesce(each.value.ipv4_address, each.value.ipv6_address, try(one(each.value.network).ip, null))
```
utay, line 174: Hey @mysticaltech and @xavierleune, thank you for this project and this PR, it's awesome!

I'm using 2.17.0 and wanted to achieve the same and realized it's already implemented on master. I think there's a bug here though: `autoscaled_nodes` is a list of [data.hcloud_server](https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs/data-sources/server) which doesn't expose a `network` attribute. Am I missing something?

xavierleune: Hi @utay, I've never used cluster autoscaling so I'm not quite sure how it works internally. I'll try it in the coming minutes, may you please tell me how to reproduce the bug you may have ? I think the network id is missing, therefore your autoscaled server won't be included in the private network and will have public ips.

utay: If you want to repro I used this commit, it's based on 2.17.0, not on master: https://github.com/utay/terraform-hcloud-kube-hetzner/commit/f2f2de31e1f98df46704ed6220de521eb6df705b

I think the issue is not that autoscaled servers don't have private IPs but rather that the hcloud terraform provider doesn't expose a private IP/network through the [data.hcloud_server](https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs/data-sources/server). Here's the error:
```
│ Error: Unsupported attribute
│
│   on .terraform/modules/env.kube/autoscaler-agents.tf line 174, in resource "null_resource" "autoscaled_nodes_registries":
│  174:     host           = one(each.value.network).ip
│     ├────────────────
│     │ each.value is object with 20 attributes
│
│ This object does not have an attribute named "network".
```

xavierleune: Network id should be fine, but the nodes definitely have public ips, according to [default configuration](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/hetzner/README.md#configuration)

xavierleune: Thanks for the reproducer, I'll try it. Considering the network node, it's on the resource when the network id is defined https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs/resources/server#server-creation-with-network

xavierleune: @utay I could not reproduce the error you reported, I:
- Created a new cluster with autoscaling enabled
- Forced the autoscaler to create a new node
- Edited k3s_registries in my kube.tf to trigger this resource

Everything went fine, the registry has been updated and I got "k3s service or k3s-agent service restarted successfully".
I'm not sure which version of the code you tested, but actually I can tell:
- There is a line number mismatch between your error and the actual autoscaler-agents.tf
- The code provided in the error code does not match the actual code, there is a `try` around

Can you please double check your code ?

xavierleune: oh yes sorry I thought v.2.17.0 included the private network. This time I got it and that's a good catch, looks [fixable](https://github.com/hetznercloud/terraform-provider-hcloud/pull/1115).

modules/host/templates/cloudinit.yaml.tpl
```diff
@@ -34,6 +34,13 @@ runcmd:
 
 ${cloudinit_runcmd_common}
 
+# Configure default route based on public ip availability
+%{if private_network_only~}
+- [ip, route, add, default, via, '10.0.0.1', dev, 'eth0']
```
utay, line 39: @xavierleune Hi again! I'm currently migrating to this commit with private IPs only + a NAT gateway. Should this `10.0.0.1` depend on `network_ipv4_cidr`? For example I use `10.32.0.0/12` as my network CIDR and I guess this should be `10.32.0.1` instead

xavierleune: You're right; the gateway is the first ip in the network range - see [reserved ips](https://docs.hetzner.com/cloud/networks/faq/#are-any-ip-addresses-reserved)

ahachmann:
Hi @xavierleune, hi everyone,
@xavierleune first of all thank you for all your effort. I appreciate your work very much. 

I am not sure if I was missing some parts in this mergerequest dialog. I have successfully set up my Cluster with private ip adresses and am very happy about that. Only thing that is bothering me now is the internet access via NAT server. That works pretty well, as the default gateway is being set arround line 143. Complete installation finishes and my workload also does work. 

But when a node restarts, it is getting problematic with containers not having been run on those nodes yet. The default route via the Hetzner network to the NAT server is not being persisted during cloud init, and as cloud init is not being run after completing machine initialization, that route is lost uppon reboot.

Am I missing a piece, or is this maybe really an open point?

Regards,
Alex

xavierleune:
Hi @ahachmann 
Are you up-to-date with v2.18.0 or v2.18.1 ? On august 11 Hetzner made a change on [their DHCP infrastructure](https://docs.hetzner.com/cloud/networks/changes/dhcp-bugfix/). @mysticaltech fixed it in #1845 in july and it's been released in v2.18.0 (but upgrade to this version is discouraged, you should skip to v2.18.1).

ahachmann:
Hi @xavierleune ,
thanks for that Heads up. I have been basically on the v2.18.1.

With that fix #1845 I believe I would still face the problem, that the route would be gone after a reboot as the 
[ip, route, add, default, via, '10.0.0.1', dev, 'eth0', metric, '100'] 
seems to not be persistend.

I now managed to get this running by changing the part where cloudinit sets the route as follows:

```
%{if private_network_only~}
# Private-only setup: eth0 is the private interface
# - [ip, route, add, default, via, '10.0.0.1', dev, 'eth0', metric, '100']

# Configure persistent default route via systemd

- |
  cat <<EOF | tee /etc/systemd/system/default-route.service
  [Unit]
  Description=Set default route
  After=network-online.target
  Wants=network-online.target

  [Service]
  Type=oneshot
  ExecStart=/sbin/ip route add default via '${network_gw_ipv4} dev eth1
  RemainAfterExit=yes

  [Install]
  WantedBy=multi-user.target
  EOF

- systemctl daemon-reload
- systemctl enable default-route.service
- systemctl start default-route.service

%{else~}
````
That basically did the trick for me. Am I missing something with the fix #1845 or is my change maybe worth a pull request?

Cheers,
Alex

mysticaltech:
Interesting, thanks for sharing @ahachmann. I'm not sure if this is needed though, I see no renewed complaints, but I will keep it in mind.

PR #1575, opened 2024-11-30
add var `disable_hetzner_ccm`

pat-s:
This is needed to allow deploying HCCM standalone which is required when one wants to include robot servers, which is only possible when deploying HCCM using the helm chart due to some helm chart magic.


pat-s:
Strange. I am using it since some time in my fork in exactly this variant without issues.

pat-s:
@mysticaltech I just tried again and was unable to reproduce the failure. I'll open another PR and maybe others can give it a try, too. Or maybe you could add how you tested it?

PR #1580, opened 2024-12-04
Allow traefik requests and limits customization

xavierleune:
This pull requests adds a new variable to hold traefik requests and limits values. Defaults are the current values used on helm but can be customized by the end user.


xavierleune:
@mysticaltech got another one for you, easier to review though ;)

PR #1596, opened 2024-12-24
Revert "add var `disable_hetzner_ccm`"

mysticaltech:
Reverts kube-hetzner/terraform-hcloud-kube-hetzner#1575

### DIFF
--- a/data.tf
+++ b/data.tf
@@ -1,5 +1,5 @@
 data "github_release" "hetzner_ccm" {
-  count       = var.hetzner_ccm_version == null && !var.disable_hetzner_ccm ? 1 : 0
+  count       = var.hetzner_ccm_version == null ? 1 : 0
   repository  = "hcloud-cloud-controller-manager"
   owner       = "hetznercloud"
   retrieve_by = "latest"
--- a/kube.tf.example
+++ b/kube.tf.example
@@ -465,8 +465,6 @@ module "kube-hetzner" {
 
   # To disable Hetzner CSI storage, you can set the following to "true", default is "false".
   # disable_hetzner_csi = true
-  # To disable Hetzner Cloud Controller Manager, you can set the following to "true", default is "false".
-  # disable_hetzner_ccm = true
 
   # If you want to use a specific Hetzner CCM and CSI version, set them below; otherwise, leave them as-is for the latest versions.
   # See https://github.com/hetznercloud/hcloud-cloud-controller-manager/releases for the available versions.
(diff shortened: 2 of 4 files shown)


pat-s:
@mysticaltech Just saw this now. Thanks for the extensive description.

The patch works for my use cases and adapting it to the structure of this module including the kustomize logic is probably not worth the effort. At least it would cost me a lot of time.

However, overall I'd argue that the option is pretty important as without this one cannot deploy HCCCM via helm (which is required when full compatibility is sought for robot servers). The helm variant does some magic behind the scenes which the kustomize one can't.

maaft:
@mysticaltech  can we discuss this further? 

I wrote a wrapper to your module that (in theory) should allow adding custom nodes (by specifying IPs and ssh credentials).

This is done via a gateway approach: I add two gateway nodes to the private network (+ floating IP for HA) and configure a wireguard mesh across the gateways and external nodes with routing into the private network. Backwards connectivity (hetzner nodes -> external nodes) is solved using a daemonset that setups the routing into the wireguard mesh.

Connectivity seems to work and the agent joins the cluster successfully.

The only thing that holds me currently back is that I can't disable ccm entirely. 

```
  Warning  FailedToCreateRoute      32m                 route_controller  Could not create route af63eba1-2bce-4da9-90eb-9f9a5285e6a1 10.42.2.0/24 for node agent-3 after 236.681619ms: hcloud/CreateRoute: server agent-3: network with id 11148319 not attached to this server
  Warning  FailedToCreateRoute      32m                 route_controller  Could not create route af63eba1-2bce-4da9-90eb-9f9a5285e6a1 10.42.2.0/24 for node agent-3 after 288.209278ms: hcloud/CreateRoute: server agent-3: network with id 11148319 not attached to this server
  Warning  FailedToCreateRoute      23m (x18 over 31m)  route_controller  (combined from similar events): Could not create route af63eba1-2bce-4da9-90eb-9f9a5285e6a1 10.42.2.0/24 for node agent-3 after 276.131005ms: hcloud/CreateRoute: server agent-3: network with id 11148319 not attached to this server
```

Hetzner CCM is always tainting my external node with "network unavailable: NoSchedule".  I even tried to disabled routes with HCLOUD_NETWORK_ROUTES_ENABLED=false, but ccm is still tainting my node.

If you have any idea on how to continue, that would be greatly appreciated.

mysticaltech:
@maaft I have no problem with this flag per se, but the PR needs to work and not break things, that's all. Please check the latest version, now HCCM supports robots (as you probably know, haven't used it myself yet), not sure if it's related to what you are doing, maybe it could work for you as is. @pat-s has contributed to a nice PR on the matter, mainly docs and missing variables to make it work.

PR #1607, opened 2025-01-07
Add fuse to base image

jidckii:
I use CSI plugin for s3 and in some modes it is necessary to have a **fuse** installed on the host.

https://github.com/yandex-cloud/k8s-csi-s3
https://github.com/yandex-cloud/geesefs


gerwim:
I assume this means we need to create a new Packer image so we have fuse available? Else some workloads might work on one host and not on another (and debugging this in one year time will be pretty hard when we have forgotten this PR).

PR #1628, opened 2025-02-04
LeapMicro & Mixed OS Support

stani-o:
First, I want to thank 🙏 the maintainers and contributors of this project for creating such a valuable project. While working professionally with terraform-hcloud-kube-hetzner, I developed these enhancements per request and would like to contribute them back to benefit the broader community. I'm happy to engage in discussions about these changes, consider alternative approaches, or explore additional enhancements that might better serve the project's goals.

This update expands support by introducing **LeapMicro OS** as an alternative to MicroOS and adding a **new Packer template** to build LeapMicro snapshots with **Fail2ban integration** for enhanced security.

---

### **LeapMicro Support**  
Users can now select between **MicroOS** and **LeapMicro**, offering greater flexibility. Both are immutable, support transactional updates, and prioritize security.  

- **MicroOS** follows a **rolling-release model**, ensuring continuous updates.  
- **LeapMicro** uses a **snapshot-based approach**, allowing users to pin kernel and package versions for controlled updates.

### **Mixed OS Deployment**  
A new `os` parameter per node pool (default: `"microos"`) enables mixed OS types within the same cluster (e.g., some nodes on MicroOS, others on LeapMicro).  

While Kubernetes clusters typically use uniform OS images, this feature:
- **Prepares for future specialized workloads** (e.g., GPU/AI-accelerated nodes needing different OS configurations).
- **Enhances troubleshooting** by enabling kernel version control in LeapMicro, allowing comparison with MicroOS.

### **New Packer Template**  
A new template (`packer-template/hcloud-leapmicro-snapshots.pkr.hcl`) builds **LeapMicro snapshots on Hetzner Cloud**, integrating:
- Official LeapMicro image setup  
- Required package installation  
- **Fail2ban** for improved security

---

### **A Non-Rolling Release Alternative**  
Some workloads benefit from a **fixed kernel version** rather than rolling updates.  
- **MicroOS** updates continuously, making it difficult to ensure consistency across nodes.  
- **LeapMicro** provides **snapshot-based updates**, offering controlled versioning.

### **Future-Proofing for Custom OS Support**  
Per-node OS configuration is a step toward supporting **specialized OS images**, preparing for diverse workloads like:
- GPU-accelerated nodes  
- AI/ML workloads requiring custom OS configurations  

### **Enhanced Security**  
The **new LeapMicro Packer template** includes **Fail2ban**, protecting nodes from brute-force attacks.

---

### **Module & Variable Updates**
- Replaced `microos_snapshot_id` with **generic `os_snapshot_id`** across modules.  
- Introduced optional `os` parameter in node pools (`control_plane_nodepools`, `agent_nodepools`, and `autoscaler_nodepools`), defaulting to `"microos"`.  
- Cloud-init settings now **adapt dynamically based on OS** (`cloudinit_write_files_common_by_os` & `cloudinit_runcmd_common_by_os`).  

### **Data Sources & Snapshot Mapping**
- Added new **LeapMicro snapshot sources**:  
  - `hcloud_image.leapmicro_x86_snapshot`  
  - `hcloud_image.leapmicro_arm_snapshot`  
- Implemented `snapshot_id_by_os` mapping to select snapshots based on OS & architecture.  

### **New Packer Template**
- Added `packer-template/hcloud-leapmicro-snapshots.pkr.hcl`  
- Builds LeapMicro snapshots, installs required packages, and **integrates Fail2ban**.

---

## Example Configuration

```json
{
  "agent_nodepools": [
    {
      "name": "agent-large",
      "server_type": "cpx51",
      "location": "hel1",
      "taints": [],
      "os": "leapmicro",
      "nodes": {
        "0": {
          "labels": ["shop-app"],
          "taints": [],
          "server_type": "cpx31",
          "placement_group": "pool1",
          "os": "leapmicro"
        }
      }
    }
  ]
}

### CODE REVIEW

**Comment by stani-o** [comment]
> So @stani-o, if you can make this PR a full blown replacement of microos, it would be wonderful please.

Thanks for the feedback, @mysticaltech! I definitely see the benefits of going all-in on LeapMicro, and I’m on board with making it the default.

The only thing I’m thinking about is existing users who are still on MicroOS. Maybe we could:

- Default to LeapMicro for all new deployments.
- Keep MicroOS as an option for those who still need it.
- Plan a phased transition with a clear deprecation timeline and migration guidance.

This way, we move forward without leaving anyone behind. Happy to adjust the PR however makes the most sense. 
Let me know what you think!

**Comment by valkenburg-prevue-ch** [comment]
Hi all, completely agree with the assessments made here; microos has bitten us in the *!@ a few times, so Leap Micro makes more sense. If we can default to Leap Micro in a way such that existing deployments will simply throw an error if the user updates the kube-hetzner module, that's perfect.

**Comment by mysticaltech** [comment]
@stani-o I see you've done awesome progress, let me know when you think we're good to go please. This will be released with our new v3.

**Review by mysticaltech — DISMISSED** [review]
Thanks for this @stani-o

**Comment by mysticaltech** [comment]
@stani-o I appreciate your contributions to the project. @pat-s is right, migration is not needed. We can have cluster running with old MicroOS nodes, it should work well. Just when they need to launch a new node, it will launch as LeapMicro. Just one question and I take it away from there, do you think this is more or less ready?

**Comment by xavierleune** [comment]
it seems that this pull request just got a lot more traction with the util-linux gate :D I agree with what has been said, the migration can be leaved to end-user as it would be pretty easy to manage for us :) thanks for this pull request !

**Comment by loomsen** [comment]
I totally agree, that this can be left to the user, as long as there is some sort of upgrade path. The bump in the major version number should be enough, imo. 
Is there a timeline for this or something? If @stani-o doesn't have a lot of time, I'd be happy to help with testing.

**Review by pat-s — COMMENTED** [review]
I was just testing this PR. When checking the plan diff (vs HEAD of the upstream project being applied last), I see that all autoscale-groups would be recreated. This is an issue because it will lead to the removal of all nodes at the same time.

If Longhorn is used, this will probably result in the failure of all volumes.
I think we should aim to prevent this before merging this PR.

@stani-o Can you replicate this in your own env?

```tf
OpenTofu will perform the following actions:

  # module.kube-hetzner.null_resource.autoscaled_nodes_registries["auto-large-arm64-longhorn"] will be created
  + resource "null_resource" "autoscaled_nodes_registries" {
      + id       = (known after apply)
      + triggers = {
          + "registries" = " "
        }
    }

  # module.kube-hetzner.null_resource.autoscaled_nodes_registries["auto-large-arm64-longhorn-26a0d09a7c4a692e"] will be destroyed
  # (because key ["auto-large-arm64-longhorn-26a0d09a7c4a692e"] is not in for_each map)
  - resource "null_resource" "autoscaled_nodes_registries" {
      - id       = "4804018249627641194" -> null
      - triggers = {
          - "registries" = " "
        } -> null
    }

  # module.kube-hetzner.null_resource.autoscaled_nodes_registries["auto-large-arm64-longhorn-59bc926c79af7f32"] will be destroyed
  # (because key ["auto-large-arm64-longhorn-59bc926c79af7f32"] is not in for_each map)
  - resource "null_resource" "autoscaled_nodes_registries" {
      - id       = "4986102651673619041" -> null
      - triggers = {
          - "registries" = " "
        } -> null
    }

  # module.kube-hetzner.null_resource.autoscaled_nodes_registries["auto-large-arm64-longhorn-795122fbdbdc2073"] will be destroyed
  # (because key ["auto-large-arm64-longhorn-795122fbdbdc2073"] is not in for_each map)
  - resource "null_resource" "autoscaled_nodes_registries" {
      - id       = "7270600058124929574" -> null
      - triggers = {
          - "registries" = " "
        } -> null
    }

  # module.kube-hetzner.null_resource.autoscaled_nodes_registries["auto-xlarge-amd64-longhorn"] will be created
  + resource "null_resource" "autoscaled_nodes_registries" {
      + id       = (known after apply)
      + triggers = {
          + "registries" = " "
        }
    }

  # module.kube-hetzner.null_resource.autoscaled_nodes_registries["auto-xlarge-amd64-longhorn-1c56348bee158478"] will be destroyed
  # (because key ["auto-xlarge-amd64-longhorn-1c56348bee158478"] is not in for_each map)
  - resource "null_resource" "autoscaled_nodes_registries" {
      - id       = "7200577285116993983" -> null
      - triggers = {
          - "registries" = " "
        } -> null
    }

  # module.kube-hetzner.null_resource.configure_autoscaler[0] must be replaced
```

pat-s:
@stani-o I've issued a PR to your branch which avoids the nodepool recreation.

I will now continue to test the `leapmicro` option.

pat-s:
I get an error in `cloud-init` after the `leapmicro` node finished booting:

```
+ nmcli con mod eth0 ipv4.dns 1.1.1.1,8.8.8.8   
Error: unknown connection 'eth0'.
```

This also results in the `k3s-agent` not being installed etc. This might be specific to a recent change or to `leapmicro`, not sure. Yet this should be easily fixable.

packer-template/hcloud-leapmicro-snapshots.pkr.hcl
```diff
@@ -0,0 +1,245 @@
...
+    "udica"
+  ], var.packages_to_install))
+
+  # leap_oss_packages = join(" ", [
+  #   #"bind-utils",
+  #   "bzip2",
+  #   "wireguard-tools",
+  #   "mtr",
+  #   "tcpdump"
+  # ])
```
pat-s: remove if unneeded?

packer-template/hcloud-leapmicro-snapshots.pkr.hcl
```diff
@@ -0,0 +1,245 @@
...
+
+    restorecon -Rv /etc/selinux/targeted/policy
+    restorecon -Rv /var/lib
+    fixfiles restore
+    touch /.autorelabel
+
+    echo so what do we have here
+    rpm -qa |grep container
+    zypper search-packages -s container-selinux
+
```
pat-s, line 173:
```suggestion
    zypper search-packages -s container-selinux
    zypper update -y

```

Otherwise the snapshot is already super outdated during creation. FWIW, I tested the update interactively and it will download packages in size of 1GB (as we already 5 months after the initial release).

packer-template/hcloud-leapmicro-snapshots.pkr.hcl
```diff
@@ -0,0 +1,245 @@
...
+    sed -i '/keys_to_console/s/^/#/' /etc/cloud/cloud.cfg
+    sed -i '/^#*PermitRootLogin/s/.*/PermitRootLogin yes/' /etc/ssh/ssh_config.d/50-suse.conf
+    setenforce 1
+    EOF
+    sleep 10 && udevadm settle && reboot
+  EOT
+
+  install_fail2ban = <<-EOT
+    transactional-update shell <<- EOF
+    git clone --branch 1.1.0 --depth 1 https://github.com/fail2ban/fail2ban.git
```
pat-s: I'd avoid hard-coding the version in here but move it somewhere where it can be dynamically updated by dependabot or similar.

packer-template/hcloud-leapmicro-snapshots.pkr.hcl
```diff
@@ -0,0 +1,245 @@
...
+    rm -rf /etc/ssh/ssh_host_*
+    echo "Make sure to use NetworkManager"
+    touch /etc/NetworkManager/NetworkManager.conf
+    sleep 1 && udevadm settle
+  EOT
+}
+
+# Source for the Leap Micro x86 snapshot
+source "hcloud" "leapmicro-x86-snapshot" {
+  image       = "ubuntu-22.04"
```
pat-s:
```suggestion
  image       = "ubuntu-24.04"
```

packer-template/hcloud-leapmicro-snapshots.pkr.hcl
```diff
@@ -0,0 +1,245 @@
...
+    creator          = "kube-hetzner"
+  }
+  snapshot_name = "OpenSUSE Leap Micro x86 by Kube-Hetzner"
+  ssh_username  = "root"
+  token         = var.hcloud_token
+}
+
+# Source for the Leap Micro ARM snapshot
+source "hcloud" "leapmicro-arm-snapshot" {
+  image       = "ubuntu-22.04"
```
pat-s:
```suggestion
  image       = "ubuntu-24.04"
```

packer-template/hcloud-leapmicro-snapshots.pkr.hcl
```diff
@@ -0,0 +1,245 @@
...
+    echo 'done. Rebooting...'
+    sleep 1 && udevadm settle && reboot
+  EOT
+
+  install_packages = <<-EOT
+    set -ex
+
+    # echo "First reboot successful, installing needed packages..."
+    ARCH=$(uname -m)
+    REPO_URL="https://ftp.rz.uni-wuerzburg.de/opensuse/distribution/leap-micro/6.1/product/repo/openSUSE-Leap-Micro-6.1-$ARCH"
```
pat-s: Would also be good to let this be versionized by dependabot or similar to forget catching new versions.

packer-template/hcloud-leapmicro-snapshots.pkr.hcl
```diff
@@ -0,0 +1,245 @@
...
+    # thereforce, we prioritize the mirrors, as they seem to have the proper versions
+
+    zypper ar "$REPO_URL" uni-wuerzburg-mirror
+    zypper --non-interactive --gpg-auto-import-keys refresh
+
+    zypper --verbose --non-interactive install --allow-vendor-change ${local.needed_packages}
+
+    rpm --import https://rpm.rancher.io/public.key
+    zypper --verbose --non-interactive install -y https://github.com/k3s-io/k3s-selinux/releases/download/v1.6.stable.1/k3s-selinux-1.6-1.slemicro.noarch.rpm
+    zypper addlock k3s-selinux
```
pat-s: Is there a reason to use a static `k3s-selinux` version here and lock it?

pat-s:
There's also a notable different between static agent pools and cluster-autoscaler ones: for the latter, one [can currently not specify image ids per nodepool](https://github.com/kubernetes/autoscaler/issues/7537#issuecomment-2801958894). 

There's only one global config `imagesForArch` which sets the architecture-specific images for ALL nodepools. Hence, setting `os = "microos"` in cluster-autoscaler pools doesn't have an effect.

stani-o:
Hey @pat-s, thanks for your review and testing so far! I'll address your comments and follow up in the next few days.

mysticaltech:
Thanks @stani-o for your hard work, and @pat-s for the in-depth review.

pat-s:
@stani-o Are you planning to rebase & continue working on this? Otherwise I'd fork and continue with it in a fresh PR.

I think this is a really important one and because I am giving a talk about this project in November, I'd like to have this merged & released by then (+ ideally some time for real-world testing).

stani-o:
> @stani-o Are you planning to rebase & continue working on this? Otherwise I'd fork and continue with it in a fresh PR.
> 
> I think this is a really important one and because I am giving a talk about this project in November, I'd like to have this merged & released by then (+ ideally some time for real-world testing).

Merged your PR. I’ll be continuing work on this, but feel free to push to this branch as well if you want to move things forward faster.

pat-s:
I can't push to this branch as I am not a maintainer. 

Can you give some ETA? Right now, it is very unclear what that means, as previously you only did very small changes/comments with many weeks in between each action. Which is why I was asking for a clarification on this ;)

stani-o:
> I can't push to this branch as I am not a maintainer. 
> 
> Can you give some ETA? Right now, it is very unclear what that means, as previously you only did very small changes/comments with many weeks in between each action. Which is why I was asking for a clarification on this ;) 

Hi @pat-s, thanks for your patience. I needed a short break recently, which unfortunately slowed things down. Please feel free to keep making PRs like before to help move things forward.

I understand the urgency from your side, but since I'm working currently on this in my free time, I can't commit to a firm ETA just yet. I'll keep you posted as I make progress and try to update more regularly. Appreciate your support!

packer-template/hcloud-leapmicro-snapshots.pkr.hcl
stani-o: thx. resolved.

packer-template/hcloud-leapmicro-snapshots.pkr.hcl
stani-o: suggestion applied. thx.

packer-template/hcloud-leapmicro-snapshots.pkr.hcl
stani-o: suggestion applied. thx.

packer-template/hcloud-leapmicro-snapshots.pkr.hcl
stani-o: suggestion applied. thx.

mysticaltech:
Thank you both for moving this forward, @stani-o and @pat-s.

The objective is seemless replacement of microOS for new nodes.

We will get there soon. Hopefully before your presentation @pat-s. I will ship in ASAP and finalize this.

pat-s:
> There's also a notable different between static agent pools and cluster-autoscaler ones: for the latter, one https://github.com/kubernetes/autoscaler/issues/7537#issuecomment-2801958894.

https://github.com/kubernetes/autoscaler/pull/8297 was merged some weeks ago.

pat-s:
@stani-o https://github.com/stani-o/terraform-hcloud-kube-hetzner/pull/2 contains some fixes which (at least for me) avoid the recreation of the control planes due to a changed approach (dynamic vs static) for private networks.

I also outsourced the static version definitions so they can be managed in a central place and possibly be updated by a dependency bot.

pat-s:
Another PR which adds the used OS as a Hetzner server label - making it easier to spot which node is running which OS (in a mixed cluster): https://github.com/stani-o/terraform-hcloud-kube-hetzner/pull/3

pat-s:
@stani-o @mysticaltech 

I am fine with the current state of this considering all the pending PRs I've made. I am using a merged version of [all the pending changes](https://github.com/stani-o/terraform-hcloud-kube-hetzner/pulls?q=sort%3Aupdated-desc+is%3Apr+is%3Aopen) for my prod cluster right now.

I've found that the packer template had various issues, which is why I made a lot of changes there (see the individual PR).

Moving forward: How do we want to proceed? This PR is open for half a year now with spurious updates/comments every month. This way, it will be hard to get it merged and align it with other incoming PRs and changes. Similarly, merging this will likely conflict other open and pending PRs. 

I suggest to finish/merge here in the near future and then let others rebase on top of it, so the overall development isn't blocked too heavily.

As this depends on whether @stani-o wants to merge my suggestions in the first place (and if so, when he does), I am also happy to open another PR which supersedes this one to move forward. I don't wanna take this PR away from @stani-o, but at the same time find/offer a way to make effective progress.
@mysticaltech let me know what your preference is!

mysticaltech:
@stani-o If you could merge @pat-s's suggestions in here, it would be great, if you agree with them of course. That way I have you both in the commit history. That would be the best scenario.

@pat-s I appreciate your inputs and push very much. This week-end I'll work on this and review/merge your PRs, and this one, whatever the state, even if in a new `staging` branch, that you can later PR to, if @stani-o is too busy or something.

pat-s:
FYI: I am still working on selinux changes to the packer template. Currently, there are many issues for control plane nodes, i.e. `metrics-server` doesn't work as port 10250 is blocked, same for prometheus exporter and others.

Custom policies are required and must be applied during cloud-init and node startup.
I already have these working when *applied on running nodes* but it is required to have these in effect without any manual user interaction right from the start.

LeapMicro seems to be more restrictive here than microOS by default.

stani-o:
> FYI: I am still working on selinux changes to the packer template. Currently, there are many issues for control plane nodes, i.e. `metrics-server` doesn't work as port 10250 is blocked, same for prometheus exporter and others.
> 
> 
> 
> Custom policies are required and must be applied during cloud-init and node startup.
> 
> I already have these working when *applied on running nodes* but it is required to have these in effect without any manual user interaction right from the start.
> 
> 
> 
> LeapMicro seems to be more restrictive here than microOS by default. 

I've run into this in the past, try to disable the dontaudit rules. There are some denies which usually doesn't appear when dontaudit is enabled.

mysticaltech:
> FYI: I am still working on selinux changes to the packer template. Currently, there are many issues for control plane nodes, i.e. `metrics-server` doesn't work as port 10250 is blocked, same for prometheus exporter and others.
> 
> Custom policies are required and must be applied during cloud-init and node startup. I already have these working when _applied on running nodes_ but it is required to have these in effect without any manual user interaction right from the start.
> 
> LeapMicro seems to be more restrictive here than microOS by default.

It make sense as it's based on SLE Micro, which is enterprise geared.

Asked grok-4-fast, it has some good ideas, pasting the slop here for future reference.

---

Ah, the classic SELinux tango—strict by design, but it loves to step on toes during K3s bootstrap. Pat-s nails it: Leap Micro's tighter defaults (inherited from SLE Micro's enterprise paranoia) mean those control-plane denials hit harder than on MicroOS, blocking kubelet metrics (port 10250), Prometheus node exporter (often 9100), and similar without tweaks. No manual fiddling allowed? We're talking automated policy injection via cloud-init, leveraging transactional updates to keep it immutable and reboot-resilient.

Here's the crisp path to tame it, tuned for Leap Micro 6.x (as of mid-2025). This assumes you're provisioning via Ignition/Combustion or direct cloud-init (e.g., for VMs on Proxmox/KVM or cloud). I'll break it down: enable base support, craft/deploy custom modules, and wire it into startup. Tested patterns from K3s docs, openSUSE forums, and RHEL-adjacent issues (since Leap mirrors SLE/RHEL SELinux).

### 1. **Enable K3s SELinux Baseline**
   K3s ships with a custom policy module (`k3s.pp`) that covers core types (e.g., `k3s_t` for binaries, `k3s_var_lib_t` for state). It auto-installs on install if SELinux is detected, but Leap Micro needs explicit flagging to avoid skips.

   - In `/etc/rancher/k3s/config.yaml` (or env vars):
     ```
     selinux: true
     ```
   - Install command (server/control-plane):
     ```
     curl -sfL https://get.k3s.io | sh -s - server --config /etc/rancher/k3s/config.yaml
     ```
     This pulls the policy from Rancher's RPM repo and loads it via `semodule`. For air-gapped? Pre-download `k3s-selinux` RPM and bundle it.

   Pro tip: If using Packer templates (as pat-s mentions), bake this config into the image build—avoids runtime surprises.

### 2. **Diagnose & Craft Custom Policies**
   Run a test node (enforcing mode) to capture denials, then generate modules. SELinux logs to `/var/log/audit/audit.log`; grep for "denied" on ports like 10250 (kubelet read-only metrics) or 9100 (Prometheus exporter).

   - **Quick Audit2Allow Workflow** (one-shot on a dev node):
     ```
     # Tail denials during K3s start (e.g., after deploying metrics-server)
     ausearch -m avc -ts recent | audit2allow -M k3s_metrics_custom

     # This spits out k3s_metrics_custom.te and .pp
     semodule -i k3s_metrics_custom.pp
     ```
     Verify: `semodule -l | grep k3s`.

   - **Hand-Roll .te for Port Binds** (if audit2allow misses nuances; common for unreserved ports like 10250):
     For kubelet/metrics-server (container_t binding TCP 10250):
     ```
     module k3s_metrics_port 1.0;

     require {
         type container_t;
         type unreserved_port_t;
         class tcp_socket { name_bind };
     }

     # Allow container processes to bind to 10250 for metrics scraping
     allow container_t unreserved_port_t:tcp_socket name_bind;
     allow container_t unreserved_port_t:tcp_socket name_connect;  # For inbound scrapes
     ```

     For Prometheus node exporter (spc_t or container_t on port 9100):
     ```
     module k3s_prometheus_port 1.0;

     require {
         type container_t;
         type unreserved_port_t;
         class tcp_socket { name_bind name_connect };
     }

     allow container_t unreserved_port_t:tcp_socket { name_bind name_connect };
     # Add if exporter needs sysfs reads: allow container_t sysfs_t:dir read;
     ```

     Compile:
     ```
     checkmodule -M -m -o k3s_metrics_port.mod k3s_metrics_port.te
     semodule_package -o k3s_metrics_port.pp -m k3s_metrics_port.mod
     ```

     These extend K3s's base policy without overriding. For more (e.g., API timeouts in metrics-server), audit might flag `httpd_t` or `kube_apiserver_t` needs—common in RHEL10/K3s HA setups where SELinux chokes on cluster-init API calls.

### 3. **Automate via Cloud-Init (Zero-Touch Startup)**
   Leap Micro's cloud-init (via Butane/Ignition) runs early boot. Use it to drop .pp files and load before K3s service. This sidesteps pat-s's packer pain by scripting the "apply on startup" bit.

   - **Sample Cloud-Config** (user-data YAML; generate Ignition JSON via `butane` tool):
     ```yaml
     #cloud-config
     # Install SELinux tools if missing (Leap Micro has them, but ensure)
     packages:
       - selinux-policy-devel
       - checkpolicy
       - policycoreutils-python-utils

     # Write custom .pp files (base64-encode your compiled .pp binaries)
     write_files:
       - encoding: b64
         content: |  # Base64 of k3s_metrics_port.pp (use `base64 -w0 file.pp`)
           [your_base64_here_for_metrics.pp]
         path: /tmp/k3s_metrics_port.pp
         owner: root:root
         permissions: '0600'
       - encoding: b64
         content: |  # Base64 of k3s_prometheus_port.pp
           [your_base64_here_for_prom.pp]
         path: /tmp/k3s_prometheus_port.pp
         owner: root:root
         permissions: '0600'

     # Load policies post-files, pre-K3s (runcmd is first-boot only)
     runcmd:
       - semodule -i /tmp/k3s_metrics_port.pp
       - semodule -i /tmp/k3s_prometheus_port.pp
       - rm -f /tmp/k3s_*.pp  # Cleanup
       # Optional: Set enforcing if permissive: setenforce 1
       # Then kick K3s: systemctl enable --now k3s-server

     # Write K3s config with SELinux flag
     write_files:
       - content: |
           selinux: true
           # Your other flags (e.g., token, node-ip)
         path: /etc/rancher/k3s/config.yaml
         owner: root:root
         permissions: '0600'

     # Ensure K3s service starts after policies (systemd override if needed)
     systemd:
       units:
         - name: k3s-server.service
           enabled: true
           dropins:
             - name: 10-selinux.conf
               content: |
                 [Unit]
                 After=selinux-policy-load.service  # Ensures policies loaded
     ```

   - **Deployment Flow**:
     1. Compile .pp on your build host, base64-encode.
     2. Embed in cloud-init YAML → Convert to Ignition JSON (`butane -o config.ign input.yaml`).
     3. Boot Leap Micro image with `--append "ignition.config.url=<your_config_url>"` or embed in ISO.
     4. On first boot: cloud-init drops/loads policies → K3s starts enforcing → No blocks on 10250/9100.

   This keeps it transactional: Reboot/rollback via `transactional-update` won't nuke policies (semodule persists in /usr/share/selinux).

### Edge Cases & Tweaks
- **If Denials Persist**: Tail `audit.log` post-boot, regenerate .te. For Prometheus-specific, set pod securityContext: `seLinuxOptions: { type: spc_t }` in DaemonSet YAML—fixes exporter binds in enforcing mode.
- **HA Control Plane**: Mirror policies across etcd nodes; use `--cluster-init` flag but audit API timeouts (seen in RHEL10 K3s bugs).
- **Testing**: Spin a Leap Micro VM, apply config, `kubectl get --raw=/metrics` on kubelet to verify 10250 flows.
- **Vs. MicroOS**: Leap's restrictiveness shines for prod (e.g., MLS support), but if you're prototyping, MicroOS's looser defaults let you skip customs initially.

pat-s:
That's what I'm trying to get right since some weeks. Already at 10-20 iso recreations. 

The custom policy I have works locally when the node is running but always fails during cloud-init. It's a bit of a pain. But this is a blocker. Right now, the iso doesn't work for control planes and we can't merge it like this.

mysticaltech:
@pat-s What about udica, that we have integrated already? https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/eafe00313448c02c8af07dd422fe45119102b2ac/README.md?plain=1#L1069-L1126

mysticaltech:
@codex please review and propose solutions.

mysticaltech:
Thanks @stani-o, you started this great initiative!

PR #1653, opened 2025-02-23
add kubelet configuration

Atrox:
Example:
```tf
# https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/
k3s_kubelet_config = <<-EOT
  apiVersion: kubelet.config.k8s.io/v1beta1
  kind: KubeletConfiguration
  imageGCLowThresholdPercent: 40
  imageGCHighThresholdPercent: 50
  imageMaximumGCAge: 24h
EOT
```


mysticaltech:
Thanks @Atrox, that's a better way of configuring it than what we have now with the command line options `k3s_server_kubelet_args` and `k3s_agent_kubelet_args`.

autoscaler-agents.tf
```diff
@@ -184,3 +184,27 @@ resource "null_resource" "autoscaled_nodes_registries" {
...
+    user           = "root"
+    private_key    = var.ssh_private_key
+    agent_identity = local.ssh_agent_identity
+    host           = each.value.ipv4_address
+    port           = var.ssh_port
+  }
+
+  provisioner "file" {
+    content     = var.k3s_kubelet_config
+    destination = "/tmp/kubelet.yaml"
```
mysticaltech, line 204: Inconsistent filename: the file is written to `/tmp/kubelet.yaml` but your update script checks `/tmp/kubelet-config.yaml`. Please align these so the script can properly detect changes.

locals.tf
```diff
@@ -799,6 +799,26 @@ else
 fi
 EOF
 
+k3s_kubelet_config_update_script = <<EOF
```
mysticaltech, line 802: In `k3s_kubelet_config_update_script`, consider gracefully handling the scenario if `/etc/rancher/k3s/kubelet-config.yaml` doesn't exist yet so the `cp` command doesn't fail.

mysticaltech (changes requested):
Overall, this PR is close to completion but there are a couple of issues that need to be addressed:

1. **Mismatch in filenames** for the kubelet config: in `autoscaler-agents.tf`, you write the file to `/tmp/kubelet.yaml`, but the update script compares `/tmp/kubelet-config.yaml` to `/etc/rancher/k3s/kubelet-config.yaml`. That mismatch means the config update script will never find the newly written file. They must match.
2. **Initial missing kubelet-config.yaml** case: In the `k3s_kubelet_config_update_script`, there's an unconditional backup step (`cp /etc/rancher/k3s/kubelet-config.yaml …`). If it's the very first time and that file doesn't exist yet, `cp` will fail. Consider gracefully handling the "file does not exist" case or only backing up if it exists.

With those fixes, this should be good to go. Thanks for your work!

Atrox:
Fixed the issues. Thanks for the catch regarding the autoscaler. I don't use those myself so I didn't notice that typo...

mysticaltech:
@Atrox Again thanks for this. There's one thing though, maybe you are on windows and using a different kind of line ending. Because the diffs change whole blocks and files, it's hard to pin point what actually really changed. Please set your file editor to edit in unix mode, that way I can better review the PR.
Screenshot From 2025-05-18 08-23-48

Atrox:
@mysticaltech i don't think the issue in that specific block is line endings. I formatted the file, so that all the lines match up with the '='. The same way it currently is, just tried to follow the project style. 

Should I remove the formatting?

mysticaltech:
Ah I see @Atrox . In that case, please just run `terraform fmt`

Atrox:
@mysticaltech fmt doesn't change any file for me. `terraform fmt -check` successfully exits on my machine.

Atrox:
@mysticaltech did a rebase with master + terraform fmt, alignment with `=` stays the same.

mysticaltech:
No worries, thanks for trying, I will test it ASAP and merge!

modules/host/main.tf
```diff
@@ -143,6 +143,31 @@ resource "null_resource" "registries" {
...
+resource "null_resource" "kubelet_config" {
+  triggers = {
+    kubelet_config = var.k3s_kubelet_config
+  }
+
+  connection {
+    user           = "root"
+    private_key    = var.ssh_private_key
+    agent_identity = local.ssh_agent_identity
+    host           = hcloud_server.server.ipv4_address
```
sando38, line 155: I believe the `host` should be 
```
host = coalesce(hcloud_server.server.ipv4_address, hcloud_server.server.ipv6_address, try(one(hcloud_server.server.network).ip, null))
```

to make sure, also private Kubernetes nodes with public IPv4 disabled can be accessed? I derived it from the `resource "null_resource" "registries"` part.

mysticaltech:
Merged into fix/staging-review-2026-01-11 branch (rebased and integrated with other changes, including fix for bastion connection in kubelet_config)

PR #1660, opened 2025-03-11
SELinux: allow RW access in /var/lib

pschiffe:
This is needed for example for DaemonSets that require hostPath volume mount.


mysticaltech (approved):
Awesome, thanks @pschiffe

PR #1671, opened 2025-03-23
Add Udica for easy SELinux

mysticaltech:
- Added detailed instructions in README.md for using the 'udica' tool to create custom SELinux policies, promoting better security practices.
- Updated kube.tf.example to include a note on SELinux enforcement and the use of 'udica' for targeted policies.
- Included 'udica' in the list of needed packages in packer-template/hcloud-microos-snapshots.pkr.hcl to ensure availability during image creation.

Part of #697 

Thanks for the hint @carolosf.

### DIFF
--- a/README.md
+++ b/README.md
@@ -995,6 +995,67 @@ enable_delete_protection = {
 
 </details>
 
+<details>
+
+<summary>Fix SELinux issues with udica</summary>
+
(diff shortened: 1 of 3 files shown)


carolosf:
Excellent PR thanks for the credit :) 

Maybe worth having a note about `--full-network-access` because it allows the workload to both connect over TCP but also create ports. 

Might be better to investigate the /usr/share/udica/templates/net_container.cil that `--full-network-access` uses and include individual rules from there in your main file.

Even if people are lazy and allow `--full-network-access` for a single workload they are probably better off security wise than disabling something for all workloads.

Might also be worth starting a github issue for crowd sourcing udica policies that people generate for common tools.

PR #1688, opened 2025-04-07
Make Kustomization folder configurable

ludnix2:
If you'd like to use a different folder name, you can configure it using the extra_kustomize_folder variable. By default, it is set to extra-manifests. This can be useful when working with multiple environments, allowing you to deploy different manifests for each one.


mysticaltech (approved):
Clean and efficient, thanks

PR #1693, opened 2025-04-09
feat: enable nodeEncryption in cilium_values when wireguard is enabled

tobiasehlert:
The change enables encryption for pure node to node traffic when WireGuard is in Cilium.


mysticaltech (approved):
@tobiasehlert Good tweak. Thanks

PR #1696, opened 2025-04-14
Fix(private-ips): Exclude flannel from renamed interfaces, Do not attach firewalls when public interfaces are disabled

xavierleune:


mysticaltech (approved):
I trust that this is a continuation of your previous PR, looking good

PR #1697, opened 2025-04-14
Feature(CCM): Provide a flag to install CCM using helm

xavierleune:
The `kubectl apply` to install ccm is [considered legacy](https://github.com/hetznercloud/hcloud-cloud-controller-manager/?tab=readme-ov-file#deployment) and using the helm version seems to be on the path forward supporting robots servers in this project.
This pull request adds a flag to use the helm version instead of `kubectl apply`. I'm not sure if we can provide a smooth upgrade or if that should be leaved to the end user.

### DIFF
--- a/kube.tf.example
+++ b/kube.tf.example
@@ -487,6 +487,10 @@ module "kube-hetzner" {
   # If you want to use a specific Hetzner CCM and CSI version, set them below; otherwise, leave them as-is for the latest versions.
   # See https://github.com/hetznercloud/hcloud-cloud-controller-manager/releases for the available versions.
   # hetzner_ccm_version = ""
+
+  # By default, new installationsπ use Helm to install Hetzner CCM. You can use the legacy deployment method (using `kubectl apply`) by setting `hetzner_ccm_use_helm = false`.
+  hetzner_ccm_use_helm = true
+
   # See https://github.com/hetznercloud/csi-driver/releases for the available versions.
   # hetzner_csi_version = ""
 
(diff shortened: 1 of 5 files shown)


kube.tf.example
```diff
@@ -487,6 +487,10 @@ module "kube-hetzner" {
   # If you want to use a specific Hetzner CCM and CSI version, set them below; otherwise, leave them as-is for the latest versions.
   # See https://github.com/hetznercloud/hcloud-cloud-controller-manager/releases for the available versions.
   # hetzner_ccm_version = ""
+
+  # By default, new installationsπ use Helm to install Hetzner CCM. You can use the legacy deployment method (using `kubectl apply`) by setting `hetzner_ccm_use_helm = false`.
```
pat-s, line 491:
```suggestion
  # By default, new installations use Helm to install Hetzner CCM. You can use the legacy deployment method (using `kubectl apply`) by setting `hetzner_ccm_use_helm = false`.
```

pat-s (commented):
I think overall this is very much needed due to the reasons outlined in my previous comments on this topic. Thanks for taking this on.

This change surely should be communicated clearly to users. A smooth transition should be possible 👍️

kube.tf.example
xavierleune: Good catch ^^

xavierleune:
@pat-s I think my last commit should do the trick to support upgrades ; WDYT ?

mysticaltech:
@xavierleune This was very much needed indeed. Did you try it? And what about the upgrade from an already installed cluster using the old method? Just curious, I'm for merging even if the upgrade path is not tested. The fact that you are using a boolean setting with false default makes this optional at first, then later we can make it default true.

xavierleune:
@mysticaltech yes I tried it, upgrade and downgrade works fine, I think we're good here :)

mysticaltech (approved):
Awesome!

PR #1698, opened 2025-04-14
Feat(packer): Install qemu-guest-agent

xavierleune:
This tiny PR installs qemu-guest-agent, to allow the password reset to work in hetzner GUI.

It may be useful in case of emergency (ssh key lost, using console in hetzner, ...)


pschiffe:
I think it would be better on line 39 with the rest of the packages:

https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/blob/master/packer-template/hcloud-microos-snapshots.pkr.hcl#L39

mysticaltech (approved):
That's great, thanks @xavierleune

PR #1703, opened 2025-04-16
SELinux: sync container_var_lib_t:file & var_lib_t:file

pschiffe:
And include `append` permission.


pschiffe:
@mysticaltech :pray: I know about udica, this is just improving the rules that are already present. Thank you for considering.

mysticaltech (approved):
Awesome

PR #1723, opened 2025-05-16
Add missing version variables to kustomization triggers

joeig:
## Background

I explicitly pin certain versions. When I changed `var.longhorn_version` from `v1.8.0` to `v1.8.1`, I noticed that the new version was not deployed.

## Observation

`var.longhorn_version` and other version variables are not part of the deployment triggers.

## Proposed fix

Add any missing version variables to the list of deployment triggers.


mysticaltech (approved):
Thanks

PR #1739, opened 2025-05-23
Create IPv6 reverse DNS records for dual-stack servers

joeig:
## Background

The servers are dual-stack (each server has an IPv4 address and an IPv6 address).

## Observation

If there is an IPv4 address, a reverse DNS record for IPv4 is created as expected.
If there is also an IPv6 address, the corresponding resource record for the IPv6 address is missing.

## Proposed fix

Create reverse DNS records for all enabled IP protocols.


mysticaltech (approved):
Thanks @joeig! This is important.

PR #1744, opened 2025-05-30
Do not schedule CSI pods on control-plane by default

orlovmyk:
According to example in official Hetzner repo it's a good approach to follow, WDYT?

https://github.com/hetznercloud/csi-driver/blob/018e2478d98b8f81a79f3d1a2c846decb8e4de6a/chart/example-prod.values.yaml#L81

P.S.
Tried to follow contribution guide, but did not find `staging` branch
> 7. Open a Pull Request targeting the staging branch.

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -568,6 +568,14 @@ persistence:
   EOT
 
   hetzner_csi_values = var.hetzner_csi_values != "" ? var.hetzner_csi_values : <<EOT
+node:
+  affinity:
+    nodeAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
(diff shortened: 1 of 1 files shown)


mysticaltech (changes requested):
@orlovmyk Thanks for this! You are right. However, sometimes the clusters are so that we have only control planes. So it you can actually ensure that this is only used when the number of worker nodes is 0, it would be great. In locals too, you have variables that keep track of this count.

Edit: You can make it dependant on `allow_scheduling_on_control_plane`, or `length(hcloud_network_subnet.agent)`. Please propose something.

PR #1745, opened 2025-05-30
cloud-init: Remove invalid "debug" key

MarekPikula:
Fixes issue that resulted in cloud-init not running:

>  Error: Cloud config schema errors: debug: Additional properties are not allowed ('debug' was unexpected)


mysticaltech (approved):
Thanks!

PR #1747, opened 2025-06-01
fix: Respect var disable_selinux on autoscale nodes

janfrederik:
k3s' config.yaml for control planes and for agent pool nodes contains an `selinux` key whose value depends on `var.disable_selinux`. However, nodes created by the autoscaler always had `selinux=true` in their config, even though selinux is correctly disabled for them in their cloud init.

This PR applies the value of `var.disable_selinux` to the k3s' config.yaml for autoscaler nodes as well.


mysticaltech (approved):
Thanks for this @janfrederik

PR #1748, opened 2025-06-01
Fix: Apply var.agent_nodes_custom_config to autoscale nodes

janfrederik:
`var.agent_nodes_custom_config` is applied to the k3s config.yaml for nodes created for the `agent_nodepools`. However, it was not applied to nodes created by the autoscaler. So it was impossible to configure autoscaler nodes to expose their kube-proxy metrics.

This PR applies `var.agent_nodes_custom_config` to nodes created by the autoscaler as well. I think we don't need a separate variable for that: worker nodes are worker nodes, whether they are created by static `agent_nodepools` or dynamically at runtime by `autoscaler_nodepools`.

I added `agent_nodepools` to `kube.tf.example` as well and slightly improved the doc.

### DIFF
--- a/autoscaler-agents.tf
+++ b/autoscaler-agents.tf
@@ -105,15 +105,18 @@ data "cloudinit_config" "autoscaler_config" {
         dns_servers       = var.dns_servers
         has_dns_servers   = local.has_dns_servers
         sshAuthorizedKeys = concat([var.ssh_public_key], var.ssh_additional_public_keys)
-        k3s_config = yamlencode({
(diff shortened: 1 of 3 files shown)


mysticaltech (dismissed):
Fully agree

mysticaltech:
@janfrederik Please if you can resolve the merge conflict, it would be wonderful.

janfrederik:
@mysticaltech Merged master into this branch.

mysticaltech:
Thanks!

PR #1756, opened 2025-06-04
Fix external load balancer configuration for agent-less clusters

MarekPikula:
For small but HA clusters (i.e., 3 control plane nodes with `allow_scheduling_on_control_plane=true`) it was not possible to create external load balancer.

1. The first patch fixes an issue where the external load balancer couldn't be created because of wrong subnet association. Previously it tried to take a subnet from the first agent node, but as in this configuration there are no agents, there was an error during LB declaration.

2. Second issue was with the LB node association due to wrong label selector. Previously control plane nodes were omitted from external load balancer label selector as the "role" tag was overwritten by agent node. It resulted in the following label selector:

   `cluster=...,engine=k3s,provisioner=terraform,role=agent_node`

   The patch fixes the selector to include the control plane nodes thus fixing the load balancer association in case of agent-less HA clusters. Hetzner allows "or" selector in the following form:

   `cluster=...,engine=k3s,provisioner=terraform,role in (control_plane_node,agent_node)`

### DIFF
--- a/init.tf
+++ b/init.tf
@@ -23,7 +23,11 @@ resource "hcloud_load_balancer_network" "cluster" {
   count = local.has_external_load_balancer ? 0 : 1
 
   load_balancer_id = hcloud_load_balancer.cluster.*.id[0]
-  subnet_id        = hcloud_network_subnet.agent.*.id[0]
+  subnet_id = (
+    length(hcloud_network_subnet.agent) > 0
+    ? hcloud_network_subnet.agent.*.id[0]
+    : hcloud_network_subnet.control_plane.*.id[0]
+  )
 }
 
 resource "hcloud_load_balancer_target" "cluster" {
--- a/init.tf
+++ b/init.tf
@@ -36,8 +36,21 @@ resource "hcloud_load_balancer_target" "cluster" {
   depends_on       = [hcloud_load_balancer_network.cluster]
   type             = "label_selector"
   load_balancer_id = hcloud_load_balancer.cluster.*.id[0]
-  label_selector   = join(",", [for k, v in merge(local.labels, local.labels_control_plane_node, local.labels_agent_node) : "${k}=${v}"])
-  use_private_ip   = true
+  label_selector = join(",", concat(
+    [for k, v in local.labels : "${k}=${v}"],
+    [
+      # Generic label merge from control plane and agent namespaces with "or",
+      # resulting in: role in (control_plane_node,agent_node)
+      for key in keys(merge(local.labels_control_plane_node, local.labels_agent_node)) :
+      "${key} in (${
+        join(",", compact([
+          for labels in [local.labels_control_plane_node, local.labels_agent_node] :
+          try(labels[key], "")
+        ]))
+      })"
+    ]
+  ))
+  use_private_ip = true
 }
 
 locals {


mysticaltech (approved):
Awesome

arbianshkodra:
Just saw this. Now autoscaling nodes are not being detected from Load Balancer. Can you further check into this.

arbianshkodra:
If anyone is encountering this, just delete the autoscaling nodes, cluster autoscaler will add them again, and this time is being detected by the Load Balancer.

MarekPikula:
Hi @arbianshkodra, sorry for the issue. Honestly, I didn't test it with autoscaling, as I thought it wouldn't be impacted. Is the solution you found sufficient, or should I dedicate some time to look into it?

arbianshkodra:
@MarekPikula right now, it seems that after I deleted autoscaling nodes, the new ones created by the clusterautoscaler, seems to be working. Will test this in the long run.

mysticaltech:
@arbianshkodra Please next time if you see a bug on here, tag me, or open an issue and tag me too. If we're going to fix this, we will need some logs at least.

jr-dimedis:
@MarekPikula @mysticaltech 

I am testing to upgrade from 2.17.2 to 2.18.14 and stumble over the following plan output, which relates to this PR I think:

```
  # module.kube-hetzner.hcloud_load_balancer_target.cluster[0] must be replaced
-/+ resource "hcloud_load_balancer_target" "cluster" {
      ~ id               = "lb-label-selector-tgt-3d97b48f73f83fe341df421e5ec1b053ddf88b6cb2de78d979b34e75d88176fe-1659641" -> (known after apply)
      ~ label_selector   = "cluster=k8stest1,engine=k3s,provisioner=terraform,role=agent_node" -> "cluster=k8stest1,engine=k3s,provisioner=terraform,role in (control_plane_node,agent_node)" # forces replacement
        # (3 unchanged attributes hidden)
    }
```

I do have a cluster with 3 control plane and a bunch of agent nodes and when I would apply this the load balancer will target HTTP traffic to the control plane nodes, which of course does not make much sense.

Shouldn't your change only be triggered when `allow_scheduling_on_control_plane=true` is set? I don't see this in the PR diff. Or am I missing something?

mysticaltech:
@jr-dimedis This will be fixed in v2.18.5, let me know if not.

jr-dimedis:
@mysticaltech 

I just tried with v2.18.5: no change. "terraform plan" gives me the same output as stated above.

When I apply this patch to `init.tf` the LB target is not touched in my project:

```diff
diff --git a/init.tf b/init.tf
index 1417912..cd0a29d 100644
--- a/init.tf
+++ b/init.tf
@@ -56,9 +56,8 @@ resource "hcloud_load_balancer_target" "cluster" {
   load_balancer_id = hcloud_load_balancer.cluster.*.id[0]
   label_selector = join(",", concat(
     [for k, v in local.labels : "${k}=${v}"],
-    [
-      # Generic label merge from control plane and agent namespaces with "or",
-      # resulting in: role in (control_plane_node,agent_node)
+    var.allow_scheduling_on_control_plane ? [
+      # Include both control plane and agent nodes when scheduling on control plane is allowed
       for key in keys(merge(local.labels_control_plane_node, local.labels_agent_node)) :
       "${key} in (${
         join(",", compact([
@@ -66,6 +65,9 @@ resource "hcloud_load_balancer_target" "cluster" {
           try(labels[key], "")
         ]))
       })"
+    ] : [
+      # Include only agent nodes when scheduling on control plane is disabled
+      for k, v in local.labels_agent_node : "${k}=${v}"
     ]
   ))
   use_private_ip = true
```

However, I am still wondering whether I am misunderstanding something fundamental here, because this issue seems so fundamental to me that I am surprised that no one else seems to have a problem with it.

Thanks for your support!

PR #1757, opened 2025-06-04
Enable gradual roll-out on Cilium config update

MarekPikula:
This ensures automatic pod restart with a new configuration in case of changed values, which makes the deployment easier to manage.

Noted as the way to go when modifying configuration with Helm (https://docs.cilium.io/en/stable/configuration/index.html#making-changes).


mysticaltech (approved):
Good one!

Issue #1704, opened 2025-04-17
ipv6 floating ip for agents


wrenix:
### Description

current only floating-ip in ipv4 format are supported.
i like to use an ipv6 on my klipper-lb

PR #1759, opened 2025-06-04
Add output with domain associations to IPs

MarekPikula:
It allows for easy integration with DNS terraform module to dynamically create DNS records based on the current cluster configuration.

It will be extended with IPv6 once #1704 gets addressed and #1739 gets merged.


mysticaltech (approved):
Looks good

Issue #1780, opened 2025-06-09
[Bug]: util-linux 2.41 breaks mounting subpath volumes


kossmac:
### Description

After upgrading util-linux 2.41 with [that](https://lists.opensuse.org/archives/list/microos@lists.opensuse.org/thread/FVFZNC7EE3YDP6ZL7SUILVGG466423KY/) os upgrade, i'm not able to start pods with subpaths defined.

> Jun 09 10:03:34 hostname k3s[1266]: Mounting command: mount
Jun 09 10:03:34 hostname k3s[1266]: Mounting arguments: --no-canonicalize -o bind /proc/1266/fd/365 /var/lib/kubelet/pods/20ada977-3b0e-49f5-9718-42c6c7bf4a82/volume-subpaths/pvc-3b9a95f9-ef7c-459a-b6e3-7c05f604678c/sftp/0
Jun 09 10:03:34 hostname k3s[1266]: Output: mount: /var/lib/kubelet/pods/20ada977-3b0e-49f5-9718-42c6c7bf4a82/volume-subpaths/pvc-3b9a95f9-ef7c-459a-b6e3-7c05f604678c/sftp/0: wrong fs type, bad option, bad superblock on /proc/1266/fd/365, missing codepage or helper program, or other error.
Jun 09 10:03:34 hostname k3s[1266]: dmesg(1) may have more information after failed mount system call.
Jun 09 10:03:34 hostname k3s[1266]: E0609 10:03:34.223475 1266 kubelet_pods.go:364] "Failed to prepare subPath for volumeMount of the container" err=<
Jun 09 10:03:34 hostname k3s[1266]: error mounting /var/lib/kubelet/pods/20ada977-3b0e-49f5-9718-42c6c7bf4a82/volumes/kubernetes.io~csi/pvc-3b9a95f9-ef7c-459a-b6e3-7c05f604678c/mount/wordpress: mount failed: exit status 32
Jun 09 10:03:34 hostname k3s[1266]: Mounting command: mount
Jun 09 10:03:34 hostname k3s[1266]: Mounting arguments: --no-canonicalize -o bind /proc/1266/fd/365 /var/lib/kubelet/pods/20ada977-3b0e-49f5-9718-42c6c7bf4a82/volume-subpaths/pvc-3b9a95f9-ef7c-459a-b6e3-7c05f604678c/sftp/0
[...]

PR #1786, opened 2025-06-11
docs: Add manual rollback instructions for MicroOS nodes

StefanIGit:
This weekend I  run into the problem that after automated OS upgrade and reboot Nginx failed to start again most likely because of https://github.com/kubernetes/kubernetes/issues/130999
So I needed to roll back the node's OS and did not find anything here :(
Here is how it did it, I think this might be useful to others too.

### DIFF
--- a/examples/micro_os_rollback/Readme.md
+++ b/examples/micro_os_rollback/Readme.md
@@ -17,7 +17,7 @@ For more details, refer to the [Kubernetes issue #130999](https://github.com/kub
 Run the following command to identify nodes with issues:
 
 ```bash
-kubectl get pods -o wide --all-namespaces | grep CreateContainerConfigError | awk '{printf "%s,", $8} END {print ""}'
+kubectl get pods -o wide --all-namespaces | grep CreateContainerConfigError | awk '{printf "%s%s", $8, (NR == total ? "" : ",")} {total=NR} END {print ""}'
 ```
 
 **Note:** The output may include duplicates or irrelevant entries (e.g., pods with uptime of 12h).
--- a/examples/micro_os_rollback/Readme.md
+++ b/examples/micro_os_rollback/Readme.md
@@ -1,0 +1,80 @@
+# Rollback Node MicroOS Manually
+
+How to manually rollback a MicroOS node to the last snapshot or be date.
+
+## Background
+
+Certain versions of `linux-utils` (e.g., >2.40) may cause errors such as:
+
+```
+...cannot mount subpath... file exists... unmount...
+```
+
+For more details, refer to the [Kubernetes issue #130999](https://github.com/kubernetes/kubernetes/issues/130999).
+
+## Step 1: Find Problematic Nodes
+
+Run the following command to identify nodes with issues:
+
+```bash
+kubectl get pods -o wide --all-namespaces | grep CreateContainerConfigError | awk '{printf "%s,", $8} END {print ""}'
+```
+
+**Note:** The output may include duplicates or irrelevant entries (e.g., pods with uptime of 12h).
+
+## Step 2: Manual Rollback Per Node
+
+SSH into each problematic node and execute the following command:
+
+```bash
+snapper --iso list | tail -2 | head -1 | awk '{print $1}' | xargs -I{} snapper rollback {} && reboot
+```
+
+### Explanation of the Command
+
+1. `snapper --iso list`: Lists all snapshots with ISO timestamps.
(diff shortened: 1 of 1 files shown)


examples/micro_os_rollback/Readme.md
```diff
@@ -0,0 +1,80 @@
...
+```
+
+**Note:** The output may include duplicates or irrelevant entries (e.g., pods with uptime of 12h).
+
+## Step 2: Manual Rollback Per Node
+
+SSH into each problematic node and execute the following command:
+
+```bash
+snapper --iso list | tail -2 | head -1 | awk '{print $1}' | xargs -I{} snapper rollback {} && reboot
```
sehlceris, line 30: In my case, I discovered the `util-linux` problem days later. There are many snapshots to sift through to figure out one that I am comfortable rolling back to. Thus, this rollback command might be a little too automagic. Perhaps a more involved list of steps would be better?

From https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/issues/1780#issuecomment-2961555628

```
snapper list
# look through the results and select a snapshot id number to roll back to
transactional-update rollback <snapshot id>
reboot
```

pat-s (commented):
This is appreciated, yet in the end just a very lengthy description how to use `transactional-updated rollback`.

Also it should be noted that this is just a temporary fix as it will boot into newer snapshots again during future boots.

To permanently fix it, you need to add `prefer-bundled-bin: true` to the `k3s` startup args.

examples/micro_os_rollback/Readme.md
StefanIGit: That's why I added  this section "Alternative: Select Snapshot by Date" at the end
should I switch it around?

StefanIGit:
I yes this infos can be found in the interwebs... but I would like have this last Sunday at 0100h :(
And added a little more explanations just in case. 
The Temporary Fix hint is a good point I will add it, in my head this was ... obvious.

> To permanently fix it, you need to add prefer-bundled-bin: true to the k3s startup args.
How permanent would that be? Does this prevent future updates of K3s until the parameter is removed?

Our current solution for "Not reboot/upgrade"

## Temporarily stopping Kured from rebooting machines for upgrade

If you find yourself needing to disable kured upgrades, you can edit the kured daemonset temporarily 
`kubectl edit daemonset -n kube-system kured`
 and change 
`--reboot-sentinel` 
to the following:
`- --reboot-sentinel=/sentinel/reboot-required-not-existing-so-its-disabled`

This ensure that kured won’t reboot machines that have an upgrade scheduled, when comfortable again, change back to:



`- --reboot-sentinel=/sentinel/reboot-required`

examples/micro_os_rollback/Readme.md
sehlceris: Ahh I missed that part! My eyes kind of glazed over after the bash-fu. I was also frantically searching for a way to roll back. Thanks!

pat-s:
> How permanent would that be? Does this prevent future updates of K3s until the parameter is removed?

Well, do your research and you'll find enough information about it (sorry, I kinda dislike when having to be the web search engine). `prefer-bundled-bin` is an official `k3s` option.
With this enabled, you can just proceed with normal updates as the `util-linux` from the OS won't be used. 

Don't get me wrong, instructions on what to do exactly are always great and welcome. But one needs to find a middleground between a "wall of text" and the bits that matter. Otherwise it adds unnecessary information to the codebase and makes the issue itself and other "rollback" related discussions become unnecessary complicated, even though they aren't. This is why I don't think we need 80 new lines in the project because of this issue. But I am also just a random contributor, nothing else 😄️

The whole case has been a faulty update of an important underlying dep in the OS. Nothing that hasn't happened before and surely not the last time it did. Rollback, pin and continue :) 
With the upcoming switch to Leap, this might not happen again.

mysticaltech (approved):
Thanks, this is useful!

MaxLikesCode:
> This is appreciated, yet in the end just a very lengthy description how to use `transactional-updated rollback`.
> 
> Also it should be noted that this is just a temporary fix as it will boot into newer snapshots again during future boots.
> 
> To permanently fix it, you need to add `prefer-bundled-bin: true` to the `k3s` startup args.

Extremely specific issues like these scare me, because how likely is it for a package like `util-linux` to break in an OpenSUSE upgrade. I think I should disable automatic upgrades for my production cluster. Adding `prefer-bundled-bin: true` fixed it. Thank you so much.

PR #1791, opened 2025-06-12
fix: https port in single node cluster (closes #1790)

### LINKED ISSUE #1790: [Bug]: Traefik HTTPS port not work in single node cluster
### Description

When the template is configured as a single node cluster, meaning `control_plane_nodepools.count=1` and `agent_nodepools.count=0`, there is no Hetzner Load Balancer created. As a side effect, due to the way Traefik is configured, port 443 is never exposed so HTTPS does not work.

This is a regression since #1775 which aimed to fix #1721.

Ironically, the same exact problem that affected HTTP before is now affecting HTTPS - an empty `websecure` block is created, which disables the HTTPS port. I'm pretty sure that the fix you want is to move the if statement inside `websecure` one line up:

```terraform
%{if !local.using_klipper_lb~} # <-- new position
  websecure:
    # %{if !local.using_klipper_lb~} # <-- old position
    proxyProtocol:
      trustedIPs:
        - 127.0.0.1/32
        - 10.0.0.0/8
```

This is exactly how it works for HTTP/`web`:

```terraform
%{if var.traefik_redirect_to_https || !local.using_klipper_lb~}
  web:
%{if var.traefik_redirect_to_https~}
    redirections:
      entryPoint:
        to: websecure
        scheme: https
        permanent: true
%{endif~}
%{if !local.using_klipper_lb~}
    proxyProtocol:
      trustedIPs:
        - 127.0.0.1/32
        - 10.0.0.0/8
```

ViRb3:
Should fix #1790 and I don't see any side effects, but please bear in mind that I only tested this in single cluster mode.

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -719,8 +719,8 @@ ports:
 %{endfor~}
 %{endif~}
 %{endif~}
-  websecure:
 %{if !local.using_klipper_lb~}
+  websecure:
     proxyProtocol:
       trustedIPs:
         - 127.0.0.1/32


mysticaltech (approved):
Thank you!

PR #1792, opened 2025-06-15
feat: add support & docs for adding Hetzner Robot servers as nodes

pat-s:
After HCCM is now integrated via helm, there's only the ability missing to configure its values.

I've extracted the instructions from https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/discussions/1311#discussioncomment-11160315, which I am using since its existence.

As this config conflicts with my custom HCCM install right now, I haven't tested it yet - maybe somebody else will? ;)

close #1405 #1311 #283

### DIFF
--- a/docs/add-robot-server.md
+++ b/docs/add-robot-server.md
@@ -41,22 +41,25 @@ It covers configuration for both k3s and Robot nodes, including networking, conf
 
 Assumptions (change these to your values!):
 - vSwitch subnet: `10.1.0.0/24`
-- vSwitch ID: `4022`
+- vSwitch ID: `9999` # arbitrary value, replace with your vSwitch ID
 - Main interface: `enp6s0`
 
-```bash
-nmcli connection add type vlan con-name vlan4022 ifname vlan4022 vlan.parent enp6s0 vlan.id 4022
+> [!CAUTION]
+> The routes and CIDR notations depend on your local setup and may vary depending on your network configuration.
 
-nmcli connection modify vlan4022 802-3-ethernet.mtu 1400
-nmcli connection modify vlan4022 ipv4.addresses '10.1.0.2/24'
-nmcli connection modify vlan4022 ipv4.gateway '10.1.0.1'
-nmcli connection modify vlan4022 ipv4.method manual
+```bash
+nmcli connection add type vlan con-name vlan9999 ifname vlan9999 vlan.parent enp6s0 vlan.id 9999
+
+nmcli connection modify vlan9999 802-3-ethernet.mtu 1400
+nmcli connection modify vlan9999 ipv4.addresses '10.1.0.2/24'
+nmcli connection modify vlan9999 ipv4.gateway '10.1.0.1'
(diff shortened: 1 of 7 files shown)


mysticaltech:
@pat-s Excellent, will review ASAP

docs/add-robot-server.md
```diff
@@ -0,0 +1,178 @@
+# Hetzner Robot Server Integration using HCCM v1.19+
+
+This guide describes how to add Hetzner **robot servers** to a Kubernetes cluster with help of the [hcloud-cloud-controller-manager](https://github.com/hetznercloud/hcloud-cloud-controller-manager), version 1.19 or newer.
+It covers configuration for both k3s and Robot nodes, including networking, configuration, and caveats.
+
+---
+
+## Prerequisites
```
garry-t, line 8: @pat-s is not clear what type of OS we can set for robot servers?

pat-s: It doesn't matter as long as there is a `k3s` agent running.

alanaasmaa:
I got it also working using cilium in "tunnel" mode. But i have issues with MTU. Im pretty newby with kube, and also overall in networking, but what i understood all nodes and pods should use MTU 1400 as this is the linit for vSwitch. 

When i use MTU 1450 my pods that are not using nodenetworking arent stable. And when i switch to MTU 1400 manually all new nodepool servers start with 1450. So should we need option to set the MTU for new nodes also? When all are MTU 1400 except new nodes, they have then issues to connect to kube api for example. My alloy pods for example cant fetch creds.

When i finish I get my setup to work I will add here what i changed and some confs I used. I did not use this fork, but i used kustomizer to get the same result ( i think ).

olexiyb:
@alanaasmaa I have had this problem as well. I had to make next change https://github.com/olexiyb/terraform-hcloud-kube-hetzner/commit/f8903aedf4eea4c1a3b15a027219b572a72bb4e9

mysticaltech:
@olexiyb You yourself contributed a lot to this methodology, thank you for that. Do you agree that this closes your PR #1405? Please I need an honest review from you if possible.

pat-s:
I never had MTU issues in relation to that, so I can't comment on it. I don't have the bandwidth to test this through with all possible networking options and combinations that one could use.

In the cluster I am using Robot servers, I am using `enable_wireguard = true` and don't make use of `cilium`. Most of this PR are manual steps anyhow and adding information after merge for scenario X seems reasonable to me.  
The changes touching `kube.tf` are unrelated to this discussion and are needed for any networking stack.

mysticaltech:
You are correct @pat-s, I will later add the reference to the different MTUs and Cilium and we're good.

utay:
Hi @pat-s, thanks a lot for documenting that!

Regarding `HCLOUD_NETWORK_ROUTES_ENABLED=false`, did you clean up the Hetzner routes manually? It doesn't seem like the HCCM does it automatically

PR #1814, opened 2025-07-05
chore: robostify default interface detection

pat-s:
The current logic does not work for all OS and addresses the issue when testing [LeapMicro support](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/1628#issuecomment-3038598265).

Tested with

- [x] LeapMicro
- [x] microos

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -70,10 +70,18 @@ locals {
     ],
     local.has_dns_servers ? [
       join("\n", compact([
-        "IFACE=$(ip route show default | awk '{print $5; exit}')",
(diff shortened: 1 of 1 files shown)


mysticaltech:
@pat-s Thanks, that's an important one. Is it ready to go? I see it's still in draft.

pat-s:
Should be good, I am also using it since some weeks in my branch in which I use some LeapMicro nodes.

mysticaltech (approved):
Awesome, thanks!

PR #1825, opened 2025-07-24
Add audit-log feature

sando38:
Pretty similar PR like the existing kubelet-config PR. Enables audit logs on the control plane:

```terraform
  k3s_audit_policy_config = <<-EOT
apiVersion: audit.k8s.io/v1
kind: Policy
rules: 
  - level: RequestResponse
    omitStages:
      - RequestReceived
    resources:
      - group: ""
        resources: ["pods", "services"]
    namespaces: ["default", "kube-system"] 
  - level: Metadata
    omitStages:
      - RequestReceived 
  - level: None
    nonResourceURLs:
      - /api*
      - /version
      - /healthz
      - /readyz
  EOT

  k3s_audit_log_path = "/var/log/k3s-audit/audit.log"  # Path to audit log file
  k3s_audit_log_maxage = 30     # Days to retain audit logs
  k3s_audit_log_maxbackup = 10  # Number of audit log files to keep
  k3s_audit_log_maxsize = 100   # Max size in MB before rotation
```

I tested it in a private IP cluster only, And actually upgrading an existing control plane. Maybe this PR does not handle agent nodes well in the sense of that it may trigger a restart of agents as well, when a audit-log has changed, therefore, if that can be optimized, I am all ears to improve it.

### DIFF
--- a/agents.tf
+++ b/agents.tf
@@ -7,31 +7,33 @@ module "agents" {
 
   for_each = local.agent_nodes
 
-  name                         = "${var.use_cluster_name_in_node_name ? "${var.cluster_name}-" : ""}${each.value.nodepool_name}${try(each.value.node_name_suffix, "")}"
-  microos_snapshot_id          = substr(each.value.server_type, 0, 3) == "cax" ? data.hcloud_image.microos_arm_snapshot.id : data.hcloud_image.microos_x86_snapshot.id
-  base_domain                  = var.base_domain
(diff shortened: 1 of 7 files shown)


mysticaltech:
@codex review please, and let me know if backward compatible or not.

mysticaltech:
@sando38 Please, if you want want this PR merged, do address all the review comments.

PR #1894, opened 2025-09-03
Updated Kured filename reference

chrisnew:
Kured 1.20.0 no longer releases as -dockerhub.yaml but as -combined.yaml.

It’s my first PR, let me know if I can improve anything.

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -136,7 +136,7 @@ locals {
     kind       = "Kustomization"
     resources = concat(
       [
-        "https://github.com/kubereboot/kured/releases/download/${local.kured_version}/kured-${local.kured_version}-combined.yaml",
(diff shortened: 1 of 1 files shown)


ViktorSchroeder:
Same pull request as mine see https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/1888

vsalomaki:
I created a PR for the same issue before checking one already exists. I'd prefer solution in this PR, but where does the "versioncompare"-function come from? It doesn't execute in my environment.

I propose to add dependency in versions.tf:
`semvers = {
      source  = "anapsix/semvers"
      version = "0.7.1"
    }
`
and then do the comparison like `... provider::semvers::compare(local.kured_version, "1.20.0") >= 0 ? "combined" : "dockerhub"`

chrisnew:
I like #1888 seems more thought out than my attempt here!

PR #1901, opened 2025-09-09
[Bug]: Hardcoded gateway 10.0.0.1 breaks custom network_ipv4_cidr (e.g. 10.1.0.0/16)

elopsod:
### Problem  
When using a `network_ipv4_cidr` different from `10.0.0.0/8` (e.g. `10.1.0.0/16`), cluster nodes lose egress.  
In the cloud-init/script a hardcoded default route is added:  

```bash
ip route add default via 10.0.0.1 dev eth1 metric 20101 2>/dev/null
```  

In Hetzner Cloud, the correct gateway for each `/16` subnet is always the **first IP address** of that subnet (`10.X.0.1`).  
For `10.1.0.0/16` this should be `10.1.0.1`. With the hardcoded `10.0.0.1`, the next hop does not exist → `Network unreachable`.  

---

### Steps to Reproduce  
1. Set `network_ipv4_cidr = "10.1.0.0/16"`.  
2. Deploy the cluster.  
3. Check routing and connectivity:  
   ```bash
   ip route
   tracepath 8.8.8.8
   dig @8.8.8.8 google.com
   ```  
4. The default route points to `10.0.0.1` and traffic cannot leave the cluster.  

---

### Expected Behavior  
Default route should point to the **first address of the actual subnet** derived from `network_ipv4_cidr` or nodepool subnet (`cidrhost(<cidr>, 1)`).  

---

### Actual Behavior  
Default route is always set to `10.0.0.1`, which only works for `10.0.0.0/16`.

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -103,7 +103,7 @@ locals {
         "if ip link show eth1 &>/dev/null; then",
         "  # Add default route via private network with high metric",
         "  # Metric 20101 matches current DHCP-provided route for compatibility",
-        "  ip route add default via 10.0.0.1 dev eth1 metric 20101 2>/dev/null",
+        "  ip route add default via ${local.network_gw_ipv4} dev eth1 metric 20101 2>/dev/null",
         "  ",
         "  # Configure NetworkManager to ignore DHCP default route on private interface",
         "  if systemctl is-active --quiet NetworkManager; then",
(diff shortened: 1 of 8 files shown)


rp-fintopio-acc:
hi @mysticaltech 
could you pls take a look ?
thank you in advance!

utay:
Was about to open the exact same PR when I realized it was already there :)

I would also be interested in merging this!

locals.tf
```diff
@@ -308,6 +308,8 @@ locals {
   # By convention the DNS service (usually core-dns) is assigned the 10th IP address in the service CIDR block
   cluster_dns_ipv4 = var.cluster_dns_ipv4 != null ? var.cluster_dns_ipv4 : cidrhost(var.service_ipv4_cidr, 10)
 
+  network_gw_ipv4 = cidrhost(var.network_ipv4_cidr, 1)
```
utay, line 315: We could add a comment saying that the gateway IP address is always the first IP address of the network CIDR range as they mention in https://docs.hetzner.com/cloud/networks/faq/#how-do-i-setup-my-own-router

elkh510:
@utay could you pls take a look

utay:
> @utay could you pls take a look

LGTM but I'm not a maintainer, we will need @mysticaltech to approve and merge this :)

elkh510:
Hi @mysticaltech 
Any chance to get this merged ?

mysticaltech:
Yes, I'm on it today no worries.

PR #1903, opened 2025-09-11
feat: allow defining custom subnet ip ranges for static agent pools

pat-s:
Instead of relying on the default ones incrementing from `10.0.x`.

This can be useful (and even needed) if certain subnets in this range are already occupied (e.g. by a vswitch connection). 

The current approach uses a hardcoded logic to increment from 0, e.g. 10.0, 10.1, 10.2, etc.
`subnet_ip_range` allows overriding a node pool subnet on an individual basis.

### DIFF
--- a/kube.tf.example
+++ b/kube.tf.example
@@ -75,7 +75,7 @@ module "kube-hetzner" {
 
   # Using the default configuration you can only create a maximum of 42 agent-nodepools.
   # This is due to the creation of a subnet for each nodepool with CIDRs being in the shape of 10.[nodepool-index].0.0/16 which collides with k3s' cluster and service IP ranges (defaults below).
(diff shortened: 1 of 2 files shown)


mysticaltech:
@pat-s This is a breaking change. What I'm exploring now is for v3 to actually have one network per nodepool. That will allow us to have up to 100 nodes per nodepool, and not the current limited 100 nodes per cluster. It resonates a little bit with what you are trying to achieve here.

pat-s:
Why is it breaking? It is optional and only used when specified and otherwise uses the old default incrementing from 10.x per node?

mysticaltech:
@pat-s Man, subnet ranges are computed very carefully, it's a whole equilibrium. I will need to think about that one more.

mysticaltech:
@codex review please, and let me know if backward compatible or not.

PR #1910, opened 2025-09-18
Document kube.tf autoscaler min_nodes option

Giuliopime:
Spent a bunch of minutes figuring out why the autoscaler didn't have the default minimum amount of nodes provided, I think adding this simple line comment can save some minutes to a bunch of people

### DIFF
--- a/kube.tf.example
+++ b/kube.tf.example
@@ -368,6 +368,7 @@ module "kube-hetzner" {
   #    name        = "autoscaled-small"
   #    server_type = "cx32"
   #    location    = "fsn1"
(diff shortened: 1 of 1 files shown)


vsalomaki:
You have a duplicate comment # there, please separate the explanation and the code-model.

On the subject, you may want to describe the situation in more depth: the autoscaled nodes are not started if there are no pods to be placed there. 
In the configuration example, there is a taint of "node.kubernetes.io/role=peak-workloads:NoExecute", which means that if there are no pods that have a matching toleration, no autoscaled nodes are started. From the Kubernetes perspective, there is no point to start even the minimum number of nodes, because there is nothing that would be assigned to these nodes.

mysticaltech:
@vsalomaki I get your point, but still it's a good small comment to help the user.

PR #1916, opened 2025-09-25
Hetzner Robot dedicated server usage via Terraform-scripts, update instructions

vsalomaki:
### Summary
- Adds templates and configuration for connecting Cluster to Robot Nodes based on the add-robot-server.md instructions.
- Updates instructions

### Problem Analysis
When starting a new cluster, the HCCM-configration as applied in current instructions will lead to HCCM giving error as the ROBOT_USER and ROBOT_PASSWORD are required when robot.enabled = true

### Solution
Add Robot Webservice user and password handling similar to hcloud_token. 

### Test plan
- [ ] Follow instructions in docs/add-robot-server.md, use actual Robot Node
- [ ] Test with default configuration without enabling Robot-configuration, check that everything works

### Gemini

> ## Summary of Changes
> This pull request significantly extends the kube-hetzner project's capabilities by enabling the integration of Hetzner Robot dedicated servers into Kubernetes clusters. It addresses a critical issue where the hcloud-cloud-controller-manager (HCCM) previously failed without explicit Robot Webservice credentials. The changes introduce new Terraform variables for secure credential management and automated vSwitch configuration, alongside updates to the HCCM to properly manage Robot nodes. Comprehensive documentation has been added to guide users through the setup, ensuring a robust and well-supported hybrid cloud environment.
> 
> ### Highlights
> * **Hetzner Robot Integration**: Enables seamless integration of Hetzner Robot dedicated servers into Kubernetes clusters managed by hcloud-cloud-controller-manager (HCCM).
> * **Secure Credential Management**: Introduces dedicated Terraform variables (robot_user, robot_password) for securely handling Robot Webservice credentials, ensuring HCCM functions correctly when robot.enabled is true.
> * **Automated vSwitch Configuration**: Adds Terraform resources to automatically configure vSwitch subnets and expose routes, facilitating private network communication between Cloud and Robot nodes.
> * **Enhanced Documentation**: Provides comprehensive, updated instructions in docs/add-robot-server.md covering network setup, HCCM configuration, and k3s agent deployment for Robot nodes.
> * **CSI Driver Compatibility**: Implements node affinity rules to prevent Hetzner CSI pods from being scheduled on Robot nodes, addressing compatibility limitations with Hetzner Cloud Volumes.
> 
> Changelog
> * **docs/add-robot-server.md**
>   
>   * Restructured the guide for better clarity and flow, focusing on prerequisites, cluster-to-vSwitch connection, Robot-to-vSwitch connection, network verification, and k3s agent configuration.
[...]

PR #1931, opened 2025-10-10
Add test setup for hetzner testing

vsalomaki:
Add ability to run basic tests in Hetzner. This includes just setting up cluster using kube.tf.example "as-is" and then destroying it.

It doesn't run the tests if the required variables are not defined in Github. But I am not entirely sure that it waits for the maintainer/resp approval before running the test.

### DIFF
--- a/.github/workflows/hetzner-test.yaml
+++ b/.github/workflows/hetzner-test.yaml
@@ -1,14 +1,16 @@
-# Prerequisities:
-# Create Hcloud context and set value to secrets.HCLOUD_CONTEXT
-# Create snapshots per README in the context and set values to secrets.SNAPSHOT_ID_ARM and secrets.SNAPSHOT_ID_X86
-# Set HCLOUD_TOKEN to the Hetzner Cloud API token
-# Set TEST_IN_HETZNER to 1
-
 name: Test in Hetzner
 
 on:
   pull_request:
-    types: [opened, synchronize]
+    types: [opened, synchronize, reopened]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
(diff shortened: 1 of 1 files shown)


vsalomaki:
This is first version and in reference to https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/1916#issuecomment-3388955405

mysticaltech:
## 🚧 Deferring for Discussion

Thank you for this contribution! Having real Hetzner cluster testing would be incredibly valuable for catching issues before they reach users.

However, I'd like to discuss a few concerns before merging:

### Concerns

1. **Approval Workflow for External PRs**
   - The workflow uses `environment: hetzner-test` which is good for gating
   - However, I want to verify: Does this definitely wait for maintainer approval before external contributors' PRs trigger the tests?
   - We need to ensure malicious PRs can't trigger resource creation

2. **Cost Implications**
   - Each test run creates actual Hetzner resources (servers, load balancers, etc.)
   - What's the estimated cost per test run?
   - Should we limit test runs to specific PRs (e.g., only those labeled with `needs-hetzner-test`)?

3. **Cleanup on Failure**
   - The `if: ${{ always() }}` on destroy is good
   - But what happens if the workflow is cancelled mid-way?
   - Should we add a scheduled cleanup job to catch orphaned resources?

4. **Security Considerations**
   - The HCLOUD_TOKEN has full API access
   - Should we create a separate project/token with limited permissions for testing?

### Questions

- Have you tested the approval gate with an external contributor's PR?
- What's the typical cost for a single test run?
- Would you be open to adding a label-based trigger instead of running on all PRs?

Let's discuss these points and then we can move forward!

---
*Reviewed for v2.20.0 release candidates - deferred pending discussion*

PR #1935, opened 2025-10-17
fix: providers is blocking external load balancer

sfroment:
When providers is at this place and you let the rest being default you'll end up in a loop waiting for the load balancer

```
Waiting for load-balancer to get an IP...
```


mysticaltech:
@sfroment Please clarify. It's not clear what's being done here.

PR #1936, opened 2025-10-19
feat: default new nodes to Leap Micro (keep MicroOS for upgrades)

pat-s:
## Summary

This PR moves the stable default OS for *new* nodepools/nodes to **openSUSE Leap Micro** while keeping **MicroOS as the default for existing nodepools on upgrade** (to avoid unexpected node recreation).

Key points:
- `os` can be set on:
  - `control_plane_nodepools[*].os`
  - `agent_nodepools[*].os` and per-node overrides via `agent_nodepools[*].nodes[*].os`
  - `autoscaler_nodepools[*].os`
  Values: `leapmicro` or `microos`.
- Upgrade-safe defaulting:
  - If a nodepool already exists in the cluster, default stays `microos` unless it is already labeled `kube-hetzner/os=leapmicro`.
  - If a nodepool is new, default is `leapmicro`.
- Image lookups are conditional on the required OS/arch, so a brand-new Leap Micro cluster does **not** require MicroOS snapshots to exist.

## SELinux (Leap Micro)

Leap Micro is stricter by default. We apply the required allowances via a small custom policy and a `k8s-selinux-policy.service` systemd unit that runs before `k3s` (ConditionSecurity=selinux, idempotent marker file).

## Packer

Adds `packer-template/hcloud-leapmicro-snapshots.pkr.hcl` (alongside MicroOS) and updates `scripts/create.sh` to build Leap Micro snapshots.

## Testing

- Upgrade test: deployed a MicroOS cluster from `v2.19.1`, then switched module source to this PR branch and ran `terraform plan/apply`:
  - no destroys / no replaces; only in-place label updates (`kube-hetzner/os=microos`)
- Fresh cluster: deployed a new cluster that defaults to Leap Micro:
  - all nodes `Ready` with `OS-IMAGE=openSUSE Leap Micro 6.1`
  - `kured`, `system-upgrade-controller`, `hcloud CCM/CSI`, `traefik`, `cert-manager` all healthy
  - confirmed Terraform only reads `data.hcloud_image.leapmicro_*` when the cluster uses Leap Micro (no MicroOS image lookup)

## Notes

If you want to migrate an existing nodepool from MicroOS -> Leap Micro, set `os = "leapmicro"` for that nodepool and replace nodes one-by-one (expected).


mysticaltech:
@pat-s Appreciate it. But please let's remove microos support, otherwise let me know why you think it's needed? My understanding is that new nodes will just require leapmicro, which is good enough. Also please make sure that gemini is happy. You can use "/gemini review" to re-review. Thanks!

mysticaltech:
@pat-s I merged Stani's branch into a local branch, the reason is that I want him in the commit history, as he started all this and spent a lot of time. If you can work with this it would be amazing please. 

For SELinux, we support Udica, don't forget.

Please address the merge conflicts, and all the reviews and then I will merge into this local branch and proceed to merging into master myself after testing it all. Thanks

mysticaltech:
@codex review please

mysticaltech:
Most importantly @pat-s , we move all of it to leapmicro, but microos nodes should keep on working, there's no reason they shouldn't work. Selinux, again we support Udica, please explore that.

pat-s:
> But please let's remove microos support, otherwise let me know why you think it's needed? My understanding is that new nodes will just require leapmicro, which is good enough. Also please make sure that gemini is happy. You can use "/gemini review" to re-review. Thanks!

Hmmm. Not a fan of this change right now (and in general). Why not let the user decide which OS they want to use? Both are totally fine and can be changed through a simple variable which in the end just uses a different packer template.
This also sets the stage for possible additional user-based contributions to make more OS images work with this module.

Besides I am still skeptical that a one-off migration will work: as said, nodes must be replaced one-by-one. 
If we just set all refs to `leapmicro` and run apply, all nodes will be replaced at the same time.

Somewhat unrelated side-note: I am just partly happy having to deal with two AIs now in the review. While I am not against AI and use it myself for coding assistance, I think there's a lot of value for human review > AI review. Even if there's a mix, it would be great to not have a 95%/5% in favor of AI.

>  I merged Stani's branch into a local branch, the reason is that I want him in the commit history, as he started all this and spent a lot of time. If you can work with this it would be amazing please.

This branch was branched off from Stanis initial branch. I think rebasing will again will just create a huge mess.
IF you want to take this on, fine (I understand the rationale) but I am not so sure its "that easy" without ending up in merge hell. And I don't have resources left for this at the moment.

> For SELinux, we support Udica, don't forget.

Hm, this is new to me (did I overlook something from previous discussions?). I gotta say I don't have resources left for another refactoring which is not straight forward and guaranteed to work.
My suggestion would be to do this in a follow-up PR and not delay this initial ground work PR further.

---

Thinking about this reply for some minutes, I have to conclude that I was hoping to not get into a "large rebase" state again but rather having found a good spot when placing this PR without any conflicts. Coming back after creation, I am somewhat overwhelmed with 20+ AI comments + an (unexpected) large refactor request.

I think at this point I have to draw a line and not continue here. Please feel free to continue and get this merged in the way you envision it. I don't care much about contribution history of mine. At this point right now I think that's the most effective (and maybe only) way to get this merged.

mikeywuu:
Hi @mysticaltech, seeing that development is continuing on this one, I wanted to ask if you/the project would be open to supporting additional OS flavors, like Ubuntu for example? I assume some people, myself included, would prefer to use a single OS for all their infrastructure. The flexibility to add support for one's preferred OS would be a nice feature in my opinion.

mysticaltech:
@mikeywuu Only considering LeapMicro cause it's just MicroOS but way more robust. It's important that the nodes are tight and the workflows we have in place work out of the box, other OS are another deal entirely, sorry.

PR #1944, opened 2025-10-27
Fix curl command port for etcd version check

subseq:
Updated the port number in the curl command for etcd version retrieval. 

New port found at: https://gist.github.com/superseb/0c06164eef5a097c66e810fe91a9d408


mysticaltech:
Merged in commit f5084d3. Thank you for catching the etcd port - K3s uses 2382, not the standard 2379.

PR #1959, opened 2025-11-18
Fix cloud-init schema validation errors in autoscaler template

superali:
Fixes cloud-init schema validation errors that prevent autoscaled nodes from registering with the cluster.

Issues Fixed:
- Remove 'debug: True' (not allowed in cloud-init v25+)
- Quote '-6' flag as string in runcmd arrays (required by strict YAML schema)

Impact:
- Cloud-init validation was failing
- k3s-agent was never getting installed
- Autoscaled nodes never registered with cluster

Testing:
- Tested on Hetzner Cloud with k3s cluster
- Verified nodes register successfully after fix


mysticaltech:
Thanks @superali

phm07:
Encountering this issue as well, would be nice to have a release with this fix soon 🙂

PR #1961, opened 2025-11-19
Fix some confusion points regarding using user kustomization

Navidda:
Following the examples (Deploying Simple Resources) lead to this error #1711 .

I did the following to make it work:

Remove the `.tpl` in the `kustomization.yaml`
Updated the ConfigMap key in README to match the previous code example.


mysticaltech (approved):
Good one, thanks

PR #1971, opened 2025-12-01
feat: Allow for networks smaller than 16 bits

paulblum00:
This change removes the hardcoding of subnets to at least `/24` size and the amount of subnets to 256. This allows for networks smaller than `/16` to be used.

I'm not entirely sure about the changes in `agents.tf` and `control_planes.tf`. I couldn't figure out how the exact offset of 101 came to be and especially why we don't just start counting private IPs from the end of the range. It seems like the control plane LB is created at the address index `n-2`. The comment [here](https://github.com/mysticaltech/terraform-hcloud-kube-hetzner/blob/master/control_planes.tf#L37) implies that LBs will be created from the start of the IP range. It would then seem logical to have VMs created at `n-3`, `n-4`, ... . The commit introducing the 101 offset and the comment linked above, is 3 years old and it would be great if someone could explain why this is needed, if it can be replaced this way and if not whether we can find a solution that doesn't involve hardcoding a subnet size.

I intentionally designed all changes in a way that makes them completely backwards compatible (unless I overlooked something).

### DIFF
--- a/agents.tf
+++ b/agents.tf
@@ -35,7 +35,7 @@ module "agents" {
   disable_ipv6                     = each.value.disable_ipv6
   ssh_bastion                      = local.ssh_bastion
   network_id                       = data.hcloud_network.k3s.id
-  private_ipv4                     = cidrhost(hcloud_network_subnet.agent[[for i, v in var.agent_nodepools : i if v.name == each.value.nodepool_name][0]].ip_range, each.value.index + (local.network_size > 16 ? 101 : floor(local.subnet_size * 0.4))
+  private_ipv4                     = cidrhost(hcloud_network_subnet.agent[[for i, v in var.agent_nodepools : i if v.name == each.value.nodepool_name][0]].ip_range, each.value.index + (local.network_size >= 16 ? 101 : floor(pow(local.subnet_size, 2) * 0.4)))
 
   labels = merge(local.labels, local.labels_agent_node)
 
--- a/control_planes.tf
+++ b/control_planes.tf
@@ -38,7 +38,7 @@ module "control_planes" {
 
   # We leave some room so 100 eventual Hetzner LBs that can be created perfectly safely
   # It leaves the subnet with 254 x 254 - 100 = 64416 IPs to use, so probably enough.
-  private_ipv4 = cidrhost(hcloud_network_subnet.control_plane[[for i, v in var.control_plane_nodepools : i if v.name == each.value.nodepool_name][0]].ip_range, each.value.index + (local.network_size > 16 ? 101 : floor(local.subnet_size * 0.4))
+  private_ipv4 = cidrhost(hcloud_network_subnet.control_plane[[for i, v in var.control_plane_nodepools : i if v.name == each.value.nodepool_name][0]].ip_range, each.value.index + (local.network_size >= 16 ? 101 : floor(pow(local.subnet_size, 2) * 0.4)))
 
   labels = merge(local.labels, local.labels_control_plane_node)
 


mysticaltech:
@paulblum00 First half is for agent nodepools, second half starting from end is for control plane nodepools. Stuff like LBs we push to a zone where we know won't be populated by nodepools. That's basically the logic.

It's very unlikely that this PR is backward compatible, if so please explain, because the way I see it, it will redo all the networks and that's a huge breaking change to running clusters upon upgrade, unless I'm mistaken.

Why do you need IP ranges smaller than 16 bits per se? I'm genuinely curious. The plan for the next major version is to allow multiple network to get around the current limitation of 100 nodes max per network. So subnets are good logical isolation, but they're working well for now. The real limitation is the max 100 nodes per network one.

paulblum00:
@mysticaltech It seems like I forgotten to make 1 final change (at least I think so, it was 2 months ago). I have now pushed a final change and in this state I'm pretty sure that it's backwards compatible. The way it works is essentially that I took the hardcoded value and turned it into a variable `subnet_amount`, making it default to the previously hardcoded value of 256.

You can proof it by substituting `var.subnet_amount` with the default of 256, seeing that this will turn the code into your current master state. On the private ipv4s I had to use this hacky workaround of checking whether the network size is in the previously hardcoded range, to set the offset to the previously hardcoded value 101.

I've also tested it locally, running a `tofu plan` with the same inputs once on your current master and once on my branch and `git diff`ing the plan output. No diff.

As to why this is needed: I'm actually really surprised that you don't immediately see the need for it. Let me put it like this:

We have a private IP range of 10.x.x.x for all of our internal infrastructure. A very usual choice is to put the highest level of differentiation first. For example maybe we have different products A, B, C, etc. so we reserve the IP range 10.0.x.x for product A, 10.1.x.x for product B, etc. At this point we're already running into a problem. If say Product A should have a kube-hetzner cluster, this would already force us to use the entire 10.0.x.x range for this cluster. That's usually not feasible at all. Usually we would have at least a second level of differentiation maybe 10.0.0.x for k8s, 10.0.1.x for VMs, etc.

Of course you could also have these 2 levels of differentiation on the first 8 bits, splitting them into 2 x 4 bits, but even the literal notation for IP addresses suggests splitting after 8 bits. A lot of companies (ours included) will already have their networking set up in a way that would make it impossible to incorporate the current state of kube-hetzner, because they would need to redo their entire IP address plan. Not even to mention that it might not even be feasible for some, because 65536 addresses for a single cluster is just way too many.



<details>
  <summary>Inputs I've used for local testing (also tried it with a network cidr '/12')</summary>

```
inputs = {
  hcloud_token = "redacted"
  ssh_private_key = "mock"
  ssh_public_key = "mock"
  cluster_name= "k8s-mon-p-fsn-001"
  control_plane_nodepools = [{
      name= "k8s-controlplane-fsn1",
      server_type= "ccx13",
      location= "fsn1",
      count= 2,
      labels= [],
      taints= [],
      disable_ipv4= true,
      disable_ipv6= true,
      }]
  agent_nodepools= [{
      name= "k8s-agent-fsn1",
      server_type= "cx33",
      location= "fsn1",
      count= 2,
      labels= [],
      taints= [],
      disable_ipv4= true,
      disable_ipv6= true,
      }]
  autoscaler_nodepools = [{
    name= "k8s-agent-fsn1",
    server_type= "cx33",
    location= "fsn1",
    min_nodes= 0,
    max_nodes= 6,
    }]
  network_region= "eu-central"
  ingress_controller= "none"
  enable_metrics_server= true
  automatically_upgrade_k3s= false
  automatically_upgrade_os= false
  initial_k3s_channel= "v1.33"
  use_cluster_name_in_node_name= false
  cni_plugin= "cilium"
  cilium_version= "v1.18.4"
  kured_version= "1.20.0"
  enable_cert_manager= false
  load_balancer_type= "lb11"
  load_balancer_location= "fsn1"
  use_control_plane_lb= true
  autoscaler_disable_ipv4= true
  autoscaler_disable_ipv6= true
  network_ipv4_cidr= "10.16.1.0/16"
  subnet_amount= 256
  nat_router_subnet_index= 10
}

```

</details>

PR #2011, opened 2026-01-16
feat: add configurable timezone for packer MicroOS snapshots

nikolaszimmermann:
Add a timezone variable to set the system timezone on the snapshot. The default is UTC, but can be overridden with any valid timezone identifier (e.g., Europe/Madrid, America/New_York).


mysticaltech:
Good one, thanks @nikolaszimmermann

PR #2014, opened 2026-01-19
Disable root access to the nat-router when enable_sudo = true

sjoerdmulder:
Implementing what the comment in kube.tf.example already said, disabling root login when enable_sudo = true on the NAT router


sjoerdmulder:
@mysticaltech can you check?

mysticaltech:
Merged in commit f5084d3. Thank you! Root SSH is now disabled when enable_sudo=true on the NAT router.

PR #2015, opened 2026-01-20
Use NAT-router for control plane access too

sjoerdmulder:
Enables external kubectl access when using `control_plane_lb_enable_public_interface = false` by automatically forwarding port 6443 on the NAT router to the control plane LB's private IP.

## Changes
* NAT Router: Automatically configures iptables DNAT/MASQUERADE rules when control plane LB has no public interface allowing using the Hetzner Firewall to restrict access to the control plane APIs.
* Kubeconfig: Uses NAT router's public IP as the server address when CP LB is private-only
* TLS Certificates: Adds NAT router IP to K3s API server certificate SANs for proper TLS validation
* Documentation: Updated README, llms.md, and variable descriptions


sjoerdmulder:
@mysticaltech can you check?

PR #2046, opened 2026-02-02
Provide bastion host if no nat_router is provided

markus-seidl:
Provides a default bastion host that can be set. Very useful if you have an existing nat_router and want to use that. If you have an existing nat_router the current code just isn't able to use a bastion host.

### DIFF
--- a/control_planes.tf
+++ b/control_planes.tf
@@ -165,6 +165,7 @@ locals {
           local.control_plane_endpoint_host,
           module.control_planes[k].ipv4_address != "" ? module.control_planes[k].ipv4_address : null,
           module.control_planes[k].ipv6_address != "" ? module.control_planes[k].ipv6_address : null,
+          var.kubeconfig_server_address != "" ? var.kubeconfig_server_address : null,
           try(one(module.control_planes[k].network).ip, null)
         ]),
       var.additional_tls_sans)
--- a/locals.tf
+++ b/locals.tf
@@ -359,7 +359,7 @@ locals {
     bastion_port        = var.ssh_port
     bastion_user        = "nat-router"
     bastion_private_key = var.ssh_private_key
-    } : {
+    } : var.default_bastion != null ? var.default_bastion : {
     bastion_host        = null
     bastion_port        = null
     bastion_user        = null
--- a/locals.tf
+++ b/locals.tf
@@ -359,7 +359,7 @@ locals {
     bastion_port        = var.ssh_port
     bastion_user        = "nat-router"
     bastion_private_key = var.ssh_private_key
-    } : var.default_bastion != null ? var.default_bastion : {
+    } : var.optional_bastion_host != null ? var.optional_bastion_host : {
     bastion_host        = null
     bastion_port        = null
     bastion_user        = null
--- a/locals.tf
+++ b/locals.tf
@@ -354,17 +354,21 @@ locals {
 
   use_nat_router = var.nat_router != null
 
-  ssh_bastion = local.use_nat_router ? {
-    bastion_host        = hcloud_server.nat_router[0].ipv4_address
-    bastion_port        = var.ssh_port
-    bastion_user        = "nat-router"
-    bastion_private_key = var.ssh_private_key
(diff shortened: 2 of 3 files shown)


mysticaltech:
Hey @markus-seidl, thanks for the contribution! 👋

I've reviewed the changes and the code looks solid - clean implementation using `coalesce()` for the fallback chain. A few questions before we move forward:

Could you describe your setup a bit more? I want to make sure I understand the scenario correctly. From the code, it looks like `optional_bastion_host` would be used when:
- You're running nodes on a private network (no public IPs)
- You're NOT using the module's `nat_router` feature
- You have your own pre-existing bastion/jump host

Is that right? The PR description mentions "existing nat_router" but the code prioritizes the module's nat_router over `optional_bastion_host`, so I want to clarify the intended workflow.

**Separate concerns:**
The `kubeconfig_server_address` TLS SAN fix for non-LB mode looks like a valid bug fix, but it seems unrelated to the bastion feature. Could you tell me how that fix came about in this PR please?

markus-seidl:
True, that looks mixed up. I used the branch to setup a new cluster with a pre-existing nat-router and I'm using this branch to fix (?) any issues I encounter
The 2.19.1 release allows to also route the control plane via the nat router (at least if you are using the one provided by the module), but I couldn't find a way to also register the IP in the certificate. Is there another way? This would allow to skip the control_plane_lb (which is expensive) for these setups.

https://github.com/mysticaltech/terraform-hcloud-kube-hetzner/pull/2015/changes#diff-cb331320b24dda8de75b6d57ae60fdae54f503c012a2e1faf339fe8ac556c0b4R148

That is the change when using control_plane_lb_enable_public_interface . Also to me it was weird that the public name `kubeconfig_server_address` was not included in the certificate , but I could also missing things.

markus-seidl:
My kube.tf as reference

```
module "kube-hetzner" {
  providers = {
    hcloud = hcloud
  }
  source = "git::https://github.com/markus-seidl/terraform-hcloud-kube-hetzner.git?ref=feature/provide-default-bastion"
  hcloud_token = var.KARI_HCLOUD_TOKEN

  cluster_name = "cluster-v0"

  # --- Networking ---
  network_region          = "eu-central"
  enable_klipper_metal_lb = true
  load_balancer_type      = null

  # --- SSH ---
  ssh_public_key             = var.K3S_PUBLIC_KEY
  ssh_private_key            = var.K3S_PRIVATE_KEY
  ssh_additional_public_keys = var.SSH_ADDITIONAL_KEYS

  automatically_upgrade_os  = false # only possible with HA
  automatically_upgrade_k3s = true

  # --- Control Plane (minimum: 1 node) ---
  control_plane_nodepools = [
    {
      name         = "cp"
      server_type  = "cx33"
      location     = "hel1"
      labels       = [],
      taints       = [], # "dedicated=control-plane:NoSchedule"
      count        = 1,
      disable_ipv4 = true,
      disable_ipv6 = true,
    }
  ]

  # --- Workers ---
  agent_nodepools = [
    {
      name         = "workers-small",
      server_type  = "cx33"
      location     = "hel1",
      labels       = [],
      taints       = [],
      count        = 1,
      disable_ipv4 = true,
      disable_ipv6 = true,
    },
  ]

  enable_longhorn = true

  use_control_plane_lb                     = false
  control_plane_lb_enable_public_interface = false
  kubeconfig_server_address                = var.private_nat_router_ip4

  optional_bastion_host = {
    bastion_host        = var.private_nat_router_ip4
    bastion_port        = "22"
    bastion_user        = "nat-router"
    bastion_private_key = var.K3S_PRIVATE_KEY
  }
  existing_network_id = [data.hcloud_network.base_network.id]

  network_ipv4_cidr = "10.0.0.0/9"
  flannel_backend   = "wireguard-native"

  enable_local_storage = true

}
```

mysticaltech:
Thanks appreciate the details @markus-seidl, will reconsider for next release.

markus-seidl:
After some tests, I can confirm that these two changes make external/pre-existing nat instances on par with ones created by this module.
Of course you need to configure your nat instance so it can work as a bastion host and the NATing of 6443 (kubernetes port).

PR #2052, opened 2026-02-10
feat: add upgrade scheduling window support for k3s plans

svetch:
Add system_upgrade_schedule_window variable to configure maintenance
windows for automated k3s upgrades via system-upgrade-controller's
window spec (v0.15.0+). When set, upgrade jobs are only created within
the specified time window.

Includes validation to reject empty window configurations.


mysticaltech:
Thanks @svetch

PR #2160, opened 2026-03-05
fix: re-upload extra manifests when control plane is replaced or kustomization changes

bkero:
## Problem

`terraform_data.kustomization_user` triggers only on `manifest_sha1` (content of each individual `.yaml.tpl` file). This causes two failure scenarios:

**Scenario 1 — Control plane rebuilt**: When the first control plane server is replaced, `/var/user_kustomize/` is wiped. On the next `terraform apply`, unchanged files have the same SHA1 so no re-upload occurs. `kustomization_user_deploy` then runs `kubectl apply -k` against an empty directory and fails with:
```
error: accumulating resources: no such file or directory: /var/user_kustomize/<file>.yaml
```

**Scenario 2 — File referenced later**: A `.yaml.tpl` file is added to `extra-manifests/` and applied (uploading the file), but is only added to `kustomization.yaml.tpl` in a subsequent apply. If the server was rebuilt in between, the file is gone from the remote but the SHA1 trigger doesn't detect this.

## Fix

Two additional triggers on each `kustomization_user` instance:

- `control_plane_id`: `terraform_data.first_control_plane.id` changes when the server is replaced, forcing all files to be re-uploaded.
- `kustomization_sha1`: hash of `kustomization.yaml.tpl`. When a new file reference is added to the kustomization, all files are re-uploaded — ensuring the remote is always consistent with what kustomize will try to apply.

## Testing

Reproduced scenario 1 by manually clearing `/var/user_kustomize/` and running `terraform apply`. Without the fix, `kubectl apply -k` fails. With the fix, files are re-uploaded and apply succeeds.


bkero (commented):
## PR Review: #2160

**Title:** fix: re-upload extra manifests when control plane is replaced or kustomization changes
**Author:** @bkero
**Files:** 1 file changed (+3/-1)

### Risk Assessment

| Factor | Value | Risk |
|--------|-------|------|
| Author tenure | ~16 years (since 2009) | 🟢 LOW |
| Prior contributions | 1 prior PR | 🟡 MEDIUM |
| Files changed | 1 file | 🟢 LOW |
| Lines changed | +3/-1 | 🟢 LOW |
| Security-critical files | No (`kustomization_user.tf` is not in the high-risk list) | 🟢 LOW |
| External dependencies | No | 🟢 LOW |

**Overall Risk:** 🟢 LOW

### Security Review

- [x] No hardcoded credentials or tokens
- [x] No suspicious external URLs
- [x] No obfuscated code
- [x] Changes match stated purpose — three trigger keys, nothing more

### Code Analysis

The change adds two triggers to `terraform_data.kustomization_user`:

**`control_plane_id = terraform_data.first_control_plane.id`**
Correct reference — `terraform_data.first_control_plane` is defined in `init.tf` and gets a new ID when the server is replaced, which wipes `/var/user_kustomize/`.

**`kustomization_sha1 = sha1(templatefile("${var.extra_kustomize_folder}/kustomization.yaml.tpl", ...))`**
Correct approach — when `kustomization.yaml.tpl` gains a new resource reference, all `kustomization_user` instances are replaced, guaranteeing the remote is consistent with what kustomize will apply.

**Style:** The existing `"${sha1(...)}"` string-interpolation wrapper was dropped in favour of the bare `sha1(...)` call, which is the preferred Terraform style. Minor improvement.

**One edge case worth noting:** `kustomization_sha1` hardcodes the filename `kustomization.yaml.tpl`. If a user configures `extra_kustomize_folder` to point at a directory that contains `.yaml.tpl` files but no `kustomization.yaml.tpl`, the `templatefile()` call will error. However, `kubectl apply -k` already requires a `kustomization.yaml` on the remote, so `kustomization.yaml.tpl` is an existing implicit requirement of the module. No new failure mode is introduced.

### Backward Compatibility

- [x] No variables removed or renamed
- [x] No default changes
- [x] No resource naming changes
- [x] No Hetzner infrastructure is recreated

**One-time effect on upgrade:** On the first `terraform apply` after adopting this change, all `kustomization_user` instances will be replaced (triggers change), causing a full re-upload of all extra manifests and a re-run of `kubectl apply -k`. This is safe and idempotent — it is also exactly the desired behaviour the fix is intended to produce.

### Release Classification

**Type:** PATCH
**Reason:** Pure bug fix. No new features, no variable changes, no state migrations, no infrastructure recreation. The one-time re-upload on upgrade is a side-effect of the fix, not a breaking change.

### External AI Verification

| Reviewer | Verdict | Key Finding |
|----------|---------|-------------|
| Claude | ✅ | Correct root cause analysis; fix is minimal and sound |
| Gemini | ✅ | Summary confirms intent matches implementation |
| Codex | N/A | Not available in this environment |

**Consensus:** Claude and Gemini agree the change is correct and safe.

### Recommendation

**Action:** APPROVE
**Notes:** The fix is correct, minimal, and addresses the root cause. The `kustomization_sha1` trigger elegantly closes the gap where files uploaded in an earlier apply cycle go missing after the kustomization references them.

mysticaltech:
Thank you @bkero. I preserved this PR's history in #2185, but replaced the stale diff with the v3-safe implementation: user kustomizations now re-upload/redeploy when the first control plane changes, and the upload/deploy triggers were hardened during review. Closing this source PR in favor of the consolidated staging PR.

PR #2162, opened 2026-03-05
Fix zram.service After=multi-user.target deadlocks autoscaler cloud-init 

### LINKED ISSUE #2161: [Bug]: autoscaler nodes with zram_size hang in cloud-init and never join the cluster
### Description

**Affected versions**: 2.19.0, 2.19.1 (introduced in 7cd5843, shipped in v2.19.0)

### Symptoms

When `zram_size` is set on an autoscaler nodepool, newly provisioned nodes boot but never join the k3s cluster. The autoscaler eventually times out and deletes them. No k3s installation takes place.

### Root cause

There is a circular systemd ordering dependency in `templates/autoscaler-cloudinit.yaml.tpl`.

The `zram.service` unit is written with `After=multi-user.target`:

```ini
[Unit]
Description=Swap with zram
After=multi-user.target
```

This unit is activated during `runcmd` via `systemctl enable --now zram.service`. However, `runcmd` runs inside `cloud-final.service`, and `multi-user.target` waits for `cloud-final.service` to complete before it becomes active. This creates a deadlock:

```
cloud-final.service (running runcmd)
  → systemctl enable --now zram.service
    → zram.service waits for multi-user.target (After=)
      → multi-user.target waits for cloud-final.service
        → deadlock - node hangs forever
```

The node never progresses past this point, so k3s is never installed and the node never joins the cluster.

**Why regular nodepools are unaffected**: regular nodes set up zram via Terraform `remote-exec` over SSH after the node has fully booted - at that point `multi-user.target` is already active and there is no deadlock.

### Verification

On a hung autoscaler node, the blocked process is visible:

```
cloud-final.service: running systemctl enable --now zram.service (PID stuck waiting)
```

Killing that PID allows `runcmd` to continue. k3s installs successfully and the node joins the cluster within minutes - confirming the deadlock is the sole blocker.

### Kube.tf file

```terraform
module "k3s-cluster" {
  version = "2.19.1"

  # ... other config ...

  autoscaler_nodepools = [
    {
      name        = "autoscaled-compute"
      server_type = "cx33"
      location    = "hel1"
      min_nodes   = 0
      max_nodes   = 3
      labels      = {}
      taints      = []
      swap_size   = "1G"
      zram_size   = "1G"  # <-- triggers the deadlock
      kubelet_args = []
    }
  ]

  autoscaler_disable_ipv4 = true
  autoscaler_disable_ipv6 = true
}
```

### Screenshots

_No response_

### Platform

Mac

acschm1d:
## Problem

When `zram_size` is set on an autoscaler nodepool, the generated cloud-init includes this sequence in `runcmd`:

```ini
[Unit]
Description=Swap with zram
After=multi-user.target        # <-- the problem

[Install]
WantedBy=multi-user.target
```

followed by:

```bash
systemctl enable --now zram.service
```

This creates an **infinite deadlock** on autoscaler nodes:

1. `cloud-final.service` is running the `runcmd`
2. `runcmd` calls `systemctl enable --now zram.service`
3. systemd queues the start job - but `zram.service` has `After=multi-user.target`, so it won't start until `multi-user.target` is active
4. `multi-user.target` is waiting for `cloud-final.service` to complete
5. `cloud-final.service` is blocked waiting for the `systemctl` call to return

The node hangs until the cluster-autoscaler's `HCLOUD_SERVER_CREATION_TIMEOUT` expires and deletes the server. **Autoscaler nodes with `zram_size` set can never join the cluster.**

This does not affect regular agent/control-plane nodes because they use a Terraform `remote-exec` provisioner over SSH, which runs after boot when `multi-user.target` is already active.

Verified by observing `systemctl list-jobs` on a stuck autoscaler node:

```
JOB  UNIT                   TYPE   STATE
167  multi-user.target      start  waiting
1635 zram.service           start  waiting    # waiting for multi-user.target
307  cloud-final.service    start  running    # blocked on systemctl
```

## Fix

Remove `After=multi-user.target` from the zram service unit in the autoscaler cloud-init template. The `WantedBy=multi-user.target` install directive is kept intact, so the service still starts correctly on subsequent boots as part of `multi-user.target`. Without the `After=` ordering constraint, systemd can start the service immediately when requested from within `cloud-final.service`.

## Testing

Reproduced and fixed on a live autoscaler node (kube-hetzner v2.19.1, OpenSUSE MicroOS, hel1 datacenter). After manually breaking the deadlock, cloud-init completed and the k3s agent joined the cluster successfully.

Fixes https://github.com/mysticaltech/terraform-hcloud-kube-hetzner/issues/2161

### DIFF
--- a/templates/autoscaler-cloudinit.yaml.tpl
+++ b/templates/autoscaler-cloudinit.yaml.tpl
@@ -140,7 +140,6 @@ ${cloudinit_runcmd_common}
   cat <<'  EOF' > /etc/systemd/system/zram.service
   [Unit]
   Description=Swap with zram
-  After=multi-user.target
 
   [Service]
   Type=oneshot


mysticaltech:
Thanks @acschm1d, this fix is included in the v2.19.3 patch train here: #2183.

The autoscaler zram cloud-init deadlock fix is preserved in the branch history and kept as a patch-safe change for the current major version. Once #2183 lands, this PR can be closed as included.

PR #2174, opened 2026-04-09
fix: trigger longhorn filesystem resize on volume size change

amalysh:
## Summary

- Adds `longhorn_volume_size` to `configure_longhorn_volume` triggers so that increasing a volume size automatically grows the filesystem
- Without this fix, changing `longhorn_volume_size` resizes the Hetzner block volume via API but the node keeps seeing the old filesystem size

## Details

The `configure_longhorn_volume` resource already runs `resize2fs` (ext4) or `xfs_growfs` (xfs) and all its inline commands are idempotent (`mkdir -p`, `mountpoint -q || mount`, `awk || echo >>`), so re-triggering on size change is safe.

## Test plan

- [ ] `terraform validate` passes
- [ ] `terraform plan` with an increased `longhorn_volume_size` shows `configure_longhorn_volume` replacement
- [ ] After apply, verify filesystem reflects the new size (`df -h`)

### DIFF
--- a/agents.tf
+++ b/agents.tf
@@ -198,10 +198,11 @@ resource "terraform_data" "configure_longhorn_volume" {
   for_each = { for k, v in local.agent_nodes : k => v if((v.longhorn_volume_size >= 10) && (v.longhorn_volume_size <= 10240) && var.enable_longhorn) }
 
   triggers_replace = {
-    agent_id = module.agents[each.key].id
+    agent_id             = module.agents[each.key].id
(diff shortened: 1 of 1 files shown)


mysticaltech:
Thanks @amalysh, this Longhorn resize trigger fix is included in the v2.19.3 patch train here: #2183.

I preserved your contribution in the branch history and folded it into the fuller Longhorn volume fix for #2180: filesystem UUID fstab entries, stale mount-path cleanup, volume/path/fstype/size triggers, and ext4 vs xfs resize handling. Once #2183 lands, this PR can be closed as included.

Issue #2045, opened 2026-02-02
[Bug]: Switching flannel backend or CNI plugin does not remove existing interface


erendogan51:
### Description

While trying the new [flannel backend option ](https://github.com/mysticaltech/terraform-hcloud-kube-hetzner/commit/b774d074e48dad23f8f33913d54fa7379388a886) feature i encountered the situation where the existing network interface is not removed causing routing to break because both the new and existing interface use the same ip network address.

Furthermore, switching from flannel to cilium seems also broken because of the same reason.

Not sure if its relevant but when i set the flannel-backend option to wireguard-native, the resulting flannel config still states only `wireguard`

### Kube.tf file

```terraform
{
  enable_wireguard = true

  # Override the flannel backend directly (takes precedence over enable_wireguard).
  # Valid values: vxlan (default), host-gw, wireguard-native
  # Use wireguard-native for Robot nodes with vSwitch to avoid MTU auto-calculation issues.
  # See https://docs.k3s.io/networking/basic-network-options for details.
  flannel_backend = "wireguard-native"
}
```

### Screenshots

flannel wireguard iface:
<img width="1303" height="120" alt="Image" src="https://github.com/user-attachments/assets/e1c7b5c3-31e2-479f-b93a-14a9f06053ac" />

flannel vxlan iface:
<img width="1293" height="166" alt="Image" src="https://github.com/user-attachments/assets/8bf9340d-a1d8-4d89-b142-b56d100d92d3" />

cillium wireguard iface:
<img width="1279" height="66" alt="Image" src="https://github.com/user-attachments/assets/bb8dde4e-7448-4d5a-adf1-448db74e0ead" />

flannel config:

<img width="1155" height="326" alt="Image" src="https://github.com/user-attachments/assets/4028c5f5-fd6f-4377-bfe9-1be9d7a98368" />

### Platform

linux

mysticaltech:
Hi @erendogan51, thanks for the detailed report!

This is actually expected behavior rather than a bug - let me explain:

## CNI/Backend Changes Require Node Restarts

**Changing the CNI plugin (flannel → cilium) or flannel backend (vxlan → wireguard-native) on an existing cluster is not a "hot-swap" operation.** The K3s maintainers have explicitly stated that [CNI changes post-provisioning are not supported](https://github.com/k3s-io/k3s/discussions/8101).

When you change these settings:
[...]

Issue #2023, opened 2026-01-27
[Bug]: k3s Error: unable to find interface eth1: route ip+net: no such network interface


MaximilianMeister:
### Description

After upgrading from `2.17.4` to `2.18.5` one of my control plane nodes went down, so i ssh'ed there to check the `k3s.service` and it was in a crash loop with:

`Error: unable to find interface eth1: route ip+net: no such network interface`

I tried rebooting the machine but that didn't help. Error was still the same.

To fix the missing `eth1` interface I ran `/etc/cloud/rename_interface.sh` manually.

On all the other nodes in the cluster (control-plane/agents) there is also no `eth1` but the `k3s.service` and `k3s-agent.service` on agents are still up and running. My guess is that if i restart the service or machine it is going to show the same error.

Any idea what happened here? I suppose `/etc/cloud/rename_interface.sh` runs only once on installation, because a reboot of the broken control-plane hasn't helped.

Is it required (and safe) to just run it again on all the other nodes?

Thanks for any hints!

### Kube.tf file

```terraform
module "kube-hetzner" {
  providers = {
    hcloud = hcloud
  }
  source = "kube-hetzner/kube-hetzner/hcloud"
  version = "2.18.5"
  ...
}
```

### Platform

Linux

steache:
@MaximilianMeister we had the same issue, in the end it was faster to upgrade to 2.18.4 & remake all the nodes from scratch

we were switching all nodes to the new/cheaper node types, so it was not a big deal :) now it stays correctly after upgrade/reboot

MaximilianMeister:
@steache thanks for the hint! Nevertheless I'd like to avoid swapping the nodes to the new flavors for now, as it'll require extra work for me in the aftermath in my setup because the IP's are going to change with the new VM's. I already had some Hetzner server IPs with bad reputation (on blacklists) which caused trouble with one of my services on the worker nodes so I'm always happy if I don't need to change IPs.

Would be great to have an alternative "fix" to this issue, than recreating the nodes.

Even if it's just someone that can confirm running `/etc/cloud/rename_interface.sh` once again manually on all nodes fixes the issue permanently.

mysticaltech:
This issue occurs because **cloud-init only runs on first boot**. When upgrading from 2.17.4 → 2.18.5, existing nodes don't get the updated `rename_interface.sh` script.

**What happened:**
[...]

Issue #1984, opened 2025-12-23
[Bug]: can't reconcile network-related changes with what's in the kube.tf file (help needed!)


asemarian:
### Description

Howdy 👋 

I recently upgraded the module from v1.2.0 to v2.18.4, closely following the instructions outlined [here](https://github.com/mysticaltech/terraform-hcloud-kube-hetzner/tree/v2.18.4?tab=readme-ov-file#upgrading-the-module). Following a failed `apply` during the transition (caused by placement group capacity limits), the cluster became unreachable due to inconsistent private IP assignments. I recovered the cluster by manually updating node IPs in `/etc/rancher/k3s/config.yaml`.

At the moment, the cluster is back online and the current state is reconciled with the infrastructure. However, running `terraform plan` consistently shows a destructive update in-place for two agent nodes that I do not want to take place:

```shell
Terraform will perform the following actions:

  # module.kube-hetzner.module.agents["8-0-agent-nbg1-16gb-cpx41"].hcloud_server.server will be updated in-place
  ~ resource "hcloud_server" "server" {
        id                         = "104828801"
        name                       = "k3s-agent-nbg1-16gb-cpx41-wyb"
        # (22 unchanged attributes hidden)

      - network {
          - alias_ips   = [] -> null
          - ip          = "10.255.0.7" -> null
          - mac_address = "86:00:00:89:a6:03" -> null
          - network_id  = 3724997 -> null
        }
      + network {
          + alias_ips   = []
          + ip          = "10.8.0.101"
          + mac_address = (known after apply)
          + network_id  = 3724997
        }

        # (1 unchanged block hidden)
    }

  # module.kube-hetzner.module.agents["9-0-agent-fsn1-16gb-cx42"].hcloud_server.server will be updated in-place
  ~ resource "hcloud_server" "server" {
        id                         = "103954962"
        name                       = "k3s-agent-fsn1-16gb-cx42-nbd"
        # (22 unchanged attributes hidden)

      - network {
          - alias_ips   = [] -> null
          - ip          = "10.255.0.9" -> null
          - mac_address = "86:00:00:89:a6:19" -> null
          - network_id  = 3724997 -> null
        }
      + network {
          + alias_ips   = []
          + ip          = "10.9.0.101"
          + mac_address = (known after apply)
          + network_id  = 3724997
        }

        # (1 unchanged block hidden)
    }

Plan: 0 to add, 2 to change, 0 to destroy.
``` 

[...]

Issue #1928, opened 2025-10-09
[Bug]: 'Unable to find interface eth1' after upgrading from 2.17.4 to 2.18.3


morinow:
### Description

I was trying to upgrade my existing cluster from module version `2.17.4` to `2.18.3` and the control-plane is not working anymore.
Logs from `journalctl` of my first control-plane node:
```
Oct 09 14:43:01 mrinow-control-plane-small-fsn1-kwl k3s[1211]: time="2025-10-09T14:43:01Z" level=info msg="Waiting to retrieve agent configuration; server is not ready: unable to find interface eth1: route ip+net: no such network interface"
Oct 09 14:43:11 mrinow-control-plane-small-fsn1-kwl k3s[1211]: time="2025-10-09T14:43:11Z" level=info msg="Waiting to retrieve agent configuration; server is not ready: unable to find interface eth1: route ip+net: no such network interface"
Oct 09 14:43:18 mrinow-control-plane-small-fsn1-kwl k3s[1211]: time="2025-10-09T14:43:18Z" level=info msg="Waiting to retrieve agent configuration; server is not ready: unable to find interface eth1: route ip+net: no such network interface"
^C
```

K3s seems to be configured to use interface `eth1` however my server only has these interfaces:
```bash
mrinow-control-plane-small-fsn1-kwl:~ # ip address
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    ...
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
    ...
3: enp7s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1450 qdisc pfifo_fast state UP group default qlen 1000
    ...
```

I saw some commands within the module that were renaming the interface to `eth1`. Is it possible that this command failed? Is this module even supposed to be compatible with clusters created on 2.17.x?

### Kube.tf file

```terraform
locals {
  # Build the subnets under 10.200.x.x so that VPN can forward all traffic to that network
  network_ipv4_cidr = "10.200.0.0/16"   # max. 65.536 IPs for cluster, 16.384 IPs for nodes, LB, etc.
  service_ipv4_cidr = "10.200.64.0/18"  # max. 16.384 IPs for services
[...]

PR #2182, opened 2026-04-23
fix: re-run rename_interface.sh on every boot via systemd oneshot

jtomaszewski:
## Problem

cloud-init runs \`/etc/cloud/rename_interface.sh\` once at first boot, which freezes the private NIC's current MAC into \`/etc/udev/rules.d/70-persistent-net.rules\` and renames the interface to \`eth1\`. If Hetzner later reassigns that MAC — which happens on private NIC detach/reattach, network reconfiguration, or when MicroOS \`transactional-update\` wipes the overlay — the frozen udev rule no longer matches. On the next boot \`eth1\` never appears and k3s/rke2 fails with:

\`\`\`
unable to find interface eth1
\`\`\`

This has been reported in several issues: #2023, #1928, #1873, #1984, #2045, #1871, #1462, and is a close relative of discussion #1287. The common workaround users reach for is setting \`flannel-iface = "enp7s0"\` in \`*_custom_config\`, which loses the original "rename to eth1" invariant and breaks on any future NIC name change.

Prior PRs #1564 and #1960 made the first-boot rename more resilient (failure-tolerant + retries), but do not help when the MAC changes **after** provisioning — the first-boot run has already succeeded and will never fire again.

## Fix

Install a systemd oneshot (\`kh-rename-interface.service\`) via cloud-init \`write_files\`, ordered \`Before=\` the k3s/rke2 units, that re-runs the existing \`/etc/cloud/rename_interface.sh\` on every boot. The script is already idempotent — it early-exits when \`eth1\` is present — so re-running it is cheap. When the iface is missing or the udev rule is stale, it refreshes the rule with the current MAC and renames, self-healing the node before k3s/rke2 starts.

- No new variables, no breaking change.
- Two touch points in \`locals.tf\`:
  - \`cloudinit_write_files_common\`: the new unit file.
  - \`cloudinit_runcmd_common\`: \`systemctl daemon-reload\` + \`systemctl enable kh-rename-interface.service\` (not \`start\` — first-boot rename still happens via the existing remote-exec, and starting while \`eth1\` already exists is a no-op).

## Production evidence

I hit this on 2026-04-23 on a long-running single-control-plane + single-agent cluster: a routine \`tofu apply\` triggered MicroOS to reboot both nodes, Hetzner had reassigned both private NIC MACs since the original provision, \`eth1\` didn't come up on either node, and k3s refused to start. ~2.5h outage resolved by SSH-installing exactly this unit on both nodes and restarting k3s. I've been running the same unit in production for ~24h (both CP and agent), and codified it in my downstream fork for future nodes.

## Testing

- \`tofu fmt -check -recursive\` passes.
- \`tofu validate\` passes.
- Manual boot verification on two real Hetzner nodes (control plane + agent) — unit runs \`Before=k3s.service\`, rename script logs success, \`eth1\` present, k3s starts cleanly. Survives \`reboot\`.

## Related

- Fixes / self-heals around: #2023, #1928, #1873, #1984, #2045, #1871, #1462
- Related discussion: #1287
- Prior first-boot-resilience PRs (not superseded): #1564, #1960

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -2011,28 +2011,52 @@ cloudinit_write_files_common = <<EOT
     systemctl restart NetworkManager
   permissions: "0744"
 
-# Systemd oneshot that re-runs rename_interface.sh on every boot. The
-# cloud-init run above only fires at first boot and freezes the current NIC
-# MAC into /etc/udev/rules.d/70-persistent-net.rules. If Hetzner later
-# reassigns that MAC (NIC detach/reattach, network reconfig, MicroOS
-# transactional-update wiping the overlay, etc.) the udev rule no longer
-# matches, eth1 never appears, and k3s/rke2 fails with
-# `unable to find interface eth1`. The rename script is idempotent
-# (early-exits when eth1 already exists), so re-running it each boot is
-# cheap and self-heals the node.
+# Lightweight wrapper invoked on every boot by kh-rename-interface.service.
+# Fast-paths the common case (eth1 already present and the udev rule's MAC
+# matches the current MAC) so subsequent boots pay no measurable cost. Only
+# when the rule is stale (e.g. Hetzner reassigned the private NIC's MAC) do
+# we fall through to the full rename_interface.sh — which handles detection,
+# udev rule rewrite, and NetworkManager restart.
+- path: /etc/cloud/rename_interface_boot.sh
+  content: |
+    #!/bin/bash
+    set -eu
+
+    UDEV_RULE=/etc/udev/rules.d/70-persistent-net.rules
+
+    if ip link show eth1 >/dev/null 2>&1; then
+      MAC=$(cat /sys/class/net/eth1/address)
+      if [ -r "$UDEV_RULE" ] && grep -q "ATTR{address}==\"$MAC\"" "$UDEV_RULE"; then
+        exit 0
+      fi
+    fi
+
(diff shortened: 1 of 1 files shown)


jtomaszewski:
Thanks @gemini-code-assist, both points are fair. Addressed in 82e40ae:

1. **NM ordering** — dropped \`NetworkManager.service\` from \`Before=\`, added \`After=NetworkManager.service\`, and dropped \`DefaultDependencies=no\` (we now want normal ordering). The rename script can use \`nmcli\` freely.

2. **Boot delay** — added a small \`rename_interface_boot.sh\` wrapper invoked by the unit. It fast-paths when \`eth1\` exists **and** the udev rule's \`ATTR{address}\` matches \`eth1\`'s current MAC, so steady-state boots pay no measurable cost. Only when the rule is actually stale (the failure mode this PR targets) does it fall through to the full \`rename_interface.sh\`.

I left the first-boot remote-exec call to \`rename_interface.sh\` untouched so upgrade paths are unaffected.

mysticaltech:
Thank you @jtomaszewski. I vetted this and integrated it into #2185 for the v3 train while preserving your original commits/history, with a small follow-up fix so the eth1 udev rule is refreshed even when the interface is already renamed. Closing this source PR so the queue points at the consolidated staging PR.

Issue #1259, opened 2024-03-01
[Feature Request]: allow the repo to work with either x86 or arm images only


roberto-sebastiano:
### Description

Hello,

**NEED:**
As a maintaner of a cluster of arm only nodes, I would like the possiblity to have the repo working with only arm images.
I tried commeting the x86 part in the packer file, but later **terraform plan** gives error about:

```
│ Error: no image found matching the selection
│
│   with module.kube-hetzner.data.hcloud_image.microos_x86_snapshot,
│   on .terraform/modules/kube-hetzner/main.tf line 6, in data "hcloud_image" "microos_x86_snapshot":
│    6: data "hcloud_image" "microos_x86_snapshot" {

```
This is true even if I pinpoint the same arm image like
  microos_x86_snapshot_id = "123456" 
  microos_arm_snapshot_id = "123456"
  where 123456 is present on hetzner cloud as a packer snapshot

**PROPOSED SOLUTION:**
something in the line of (see count):

**main.tf**
```
data "hcloud_image" "microos_x86_snapshot" {
  count = var.microos_x86_snapshot_id == null ? 0 : 1

  with_selector     = "microos-snapshot=yes"
  with_architecture = "x86"
  most_recent       = true
}
```

probably this would need to be patched in other files as well, I can provide a PR if needed
❤️ 

Roberto Sebastiano

mysticaltech:
@roberto-sebastiano You can have your cluster 100% ARM, it's what I use actually, no problem at all. The snapshot does not cost anything, don't worry about it.

mysticaltech:
@roberto-sebastiano Why would you build each time? I haven't built mine in the last 6 months.

Also, please PR welcome if you find a way. The problem is that packer and terraform are not connected, so that was the simpler solution we could find. You may want to customize the code to work for you just with ARM, but it's just not easily generalizable.

PR #2206, opened 2026-05-29
fix: make MicroOS snapshot lookup architecture-aware

dimonb:
## What

Make MicroOS snapshot lookup architecture-aware.

Today both `data.hcloud_image.microos_x86_snapshot` and `data.hcloud_image.microos_arm_snapshot` are declared unconditionally. That means an x86-only cluster still fails during plan if no ARM MicroOS snapshot exists, and an ARM-only cluster still fails if no x86 MicroOS snapshot exists. This PR only looks up the snapshot architecture that is actually needed by configured control-plane, agent, or autoscaler nodepools, unless the corresponding explicit `microos_*_snapshot_id` input is already provided.

This keeps the existing behavior for mixed-architecture clusters: if both x86 and ARM nodepools are configured, both snapshots are still required.

## Why

This is the issue shape described in #1259, and it also shows up for the opposite direction in x86-only clusters when ARM/CAX capacity is unavailable. In that case the ARM snapshot cannot be built, but Terraform still fails before it can plan an x86-only cluster because the unused ARM image data source is evaluated.

The failure looks like:

```text
Error: no image found matching the selection

  with data.hcloud_image.microos_arm_snapshot,
  on main.tf line 12, in data "hcloud_image" "microos_arm_snapshot":
  12: data "hcloud_image" "microos_arm_snapshot" {
```

## Validation

- `terraform fmt -recursive -check`
- `git diff --check`
- `terraform init -backend=false -input=false -no-color`
- `terraform validate -no-color`

`terraform validate` still reports the existing unrelated `hcloud_primary_ip.assignee_type` warning.

### DIFF
--- a/agents.tf
+++ b/agents.tf
@@ -8,7 +8,7 @@ module "agents" {
   for_each = local.agent_nodes
 
   name                             = "${var.use_cluster_name_in_node_name ? "${var.cluster_name}-" : ""}${each.value.nodepool_name}${try(each.value.node_name_suffix, "")}"
-  microos_snapshot_id              = substr(each.value.server_type, 0, 3) == "cax" ? data.hcloud_image.microos_arm_snapshot.id : data.hcloud_image.microos_x86_snapshot.id
+  microos_snapshot_id              = substr(each.value.server_type, 0, 3) == "cax" ? local.microos_arm_snapshot_id : local.microos_x86_snapshot_id
   base_domain                      = var.base_domain
   ssh_keys                         = length(var.ssh_hcloud_key_label) > 0 ? concat([local.hcloud_ssh_key_id], data.hcloud_ssh_keys.keys_by_selector[0].ssh_keys.*.id) : [local.hcloud_ssh_key_id]
   ssh_port                         = var.ssh_port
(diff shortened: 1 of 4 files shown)


0lini:
It seems like hetzner currently does not allow creation of arm servers, so this is required for creating new clusters.

dimonb:
> It seems like hetzner currently does not allow creation of arm servers, so this is required for creating new clusters.

yes, that was the reason )

0lini:
> > It seems like hetzner currently does not allow creation of arm servers, so this is required for creating new clusters.
> 
> yes, that was the reason )



> > It seems like hetzner currently does not allow creation of arm servers, so this is required for creating new clusters.
> 
> yes, that was the reason )

Im now running on v3, it already includes a fix for this in form of   enabled_architectures

fr-ivasic:
> > It seems like hetzner currently does not allow creation of arm servers, so this is required for creating new clusters.
> 
> yes, that was the reason )

Most likely it will stay that way, I don't believe Hetzner will come back offering so cheap instances like `CX` or `CAX` for ARM64. Proper flexibility to choose instance is very helpful.

mysticaltech:
Accepted. I ported the architecture-aware Leap Micro snapshot lookup into the v2.20.0 candidate #2208 and staging at 14144f0.

The integrated version also removes a stale autoscaler dependency on the old x86-only snapshot data source.

Closing this PR as superseded by the integrated branch. Thank you.

PR #2211, opened 2026-06-07
fix: install Gateway API CRDs before Traefik when gateway provider is enabled

Juoper:
When traefik_provider_kubernetes_gateway_enabled is true, the Traefik Helm chart creates GatewayClass and Gateway resources that require the gateway.networking.k8s.io/v1 CRDs to already exist. Without them the helm-install-traefik job crashes with 'no matches for kind GatewayClass', the Traefik service is never created, and the 'Waiting for load-balancer to get an IP' loop times out, failing the entire apply.

Adds the Gateway API standard-install manifest to the main kustomization resources, ordered before the ingress controller entry so the CRDs are established before the Traefik HelmChart job runs. A gateway_api_version variable (default v1.2.1) allows pinning the CRD version.

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -209,6 +209,7 @@ locals {
       ],
       var.hetzner_ccm_use_helm ? ["hcloud-ccm-helm.yaml"] : ["ccm-networks.yaml"],
       var.disable_hetzner_csi ? [] : ["hcloud-csi.yaml"],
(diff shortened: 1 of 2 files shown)


mysticaltech:
Thanks @Juoper, the root cause is real and this is patch-worthy. I folded the safe part into #2212 for v2.20.1, with the extra release hardening needed before merging: a validated `gateway_api_version`, docs/example updates, changelog entry, and local Terraform/OpenTofu/kube-test gates.

I am closing this PR as superseded by #2212 rather than merging this branch as-is.

PR #2218, opened 2026-06-22
feat: add configurable kubectl apply options for user kustomization

orlovmyk:
### Problem

When deploying large CRDs (e.g. Argo CD `applicationsets.argoproj.io`), client-side `kubectl apply` can fail with:
```
metadata.annotations: Too long: may not be more than 262144 bytes
```

This happens because kubectl stores the full manifest in the `kubectl.kubernetes.io/last-applied-configuration` annotation, which has a 262144-byte limit. **Server-side apply** avoids that.

### Solution

The hardcoded apply command in `kustomization_user.tf`:

```hcl
kubectl apply -k /var/user_kustomize/ --wait=true
```
is now:

```hcl
kubectl apply -k /var/user_kustomize/ ${var.kustomize_apply_options}
```

New variable kustomize_apply_options (default: "--wait=true") — existing behavior is unchanged unless overridden.

NON-AI MY COMMENT

you can try to install this or this:
  - https://github.com/cert-manager/cert-manager/releases/download/v1.16.2/cert-manager.crds.yaml
  - https://github.com/argoproj/argo-cd/manifests/crds?ref=v3.4.4

### DIFF
--- a/README.md
+++ b/README.md
@@ -421,9 +421,10 @@ Change folder name with `extra_kustomize_folder`. See [example](examples/kustomi
 <details>
 <summary><strong>Custom post-install actions (ArgoCD, etc.)</strong></summary>
 
-For CRD-dependent applications:
+For CRD-dependent applications (use `kustomize_apply_options` with `--server-side` when CRDs hit the annotation size limit):
 
 ```tf
+kustomize_apply_options = "--server-side --field-manager=kube-hetzner --force-conflicts --wait=true"
 extra_kustomize_deployment_commands = <<-EOT
   kubectl -n argocd wait --for condition=established --timeout=120s crd/appprojects.argoproj.io
   kubectl -n argocd wait --for condition=established --timeout=120s crd/applications.argoproj.io
--- a/docs/llms.md
+++ b/docs/llms.md
@@ -2332,9 +2332,17 @@ Locked and loaded! Let's continue the detailed exploration.
 **Section 2.21: Kustomize and Post-Deployment Operations**
 
 ```terraform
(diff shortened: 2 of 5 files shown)


mysticaltech:
Thanks — the default preserves current behavior and the server-side-apply use case is legit, but this needs rework before merge:

1. **Conflicts with the just-merged #2225**, which split `kubectl apply -k` into its own provisioner. Please rebase and apply the options to the isolated apply provisioner only.
2. **Raw shell interpolation**: the value is injected verbatim into a remote root shell. Operator-controlled, but still avoidable — prefer `list(string)` with strict flag-token validation (or at minimum `nullable=false` + rejection of shell metacharacters/newlines).
3. **Missing trigger**: changing only `kustomize_apply_options` won't re-run the deploy (the resource is replaced only via `terraform_data.kustomization_user`). Add it to the replacement triggers.
4. Add a CHANGELOG entry for a user-facing input.
5. **Routing**: this is a feature, so it should target `staging` (v3), not master during a patch release. Please retarget.

Happy to re-review after the rework.

mysticaltech:
Merged into `master` via our new integrate-and-fix flow — your commit is preserved, with maintainer hardening on top rather than another review round: rebased onto the #2225 provisioner split (options apply to the isolated `kubectl apply -k`), variable is now `list(string)` with strict flag-token validation (shell metacharacters rejected), a `triggers_replace` hash so option-only changes re-run the deploy, docs switched to list syntax, changelog added. Default `["--wait=true"]` keeps existing behavior. Ships in v2.21.0. Thanks @orlovmyk!

PR #2226, opened 2026-07-02
fix(kustomize): don't hide deploy errors when the connection uses a sensitive value

knuurr:
> **Stacked on #2225** (the fail-loud fix). Please review/merge that first;
> this PR's net change is only the connection un-masking (locals + the deploy `connection`).

## Problem

tofu/terraform hides **all** output of a `remote-exec` provisioner when anything in its config or
`connection` is a sensitive value. In `kustomization_user_deploy` the `connection` uses
`var.ssh_private_key`. If that value is sensitive, every line of the deploy — including
`kubectl apply -k` errors — prints only as:

```
(output suppressed due to sensitive value in config)
```

So even with the fail-loud fix you get a hard failure but no idea what went wrong. (If your key
isn't sensitive, nothing changes — your output was never hidden.)

## Fix

Strip the sensitive mark from the deploy connection's SSH-key fields with `try(nonsensitive(x), x)`.
Why it's safe:

- these fields are only used to open the SSH session — they are never printed to stdout (the
  connection log shows `Private key: true`, not the key itself);
- provisioner `connection` config is not written to state (a `terraform_data` persists only
  `id/input/output/triggers_replace`; no `connection` blocks appear in state);
- sensitive marks change what is *displayed*, not what is *stored* — so this affects log output only;
- real secrets in `extra_kustomize_deployment_commands` stay hidden — they run in their own
  provisioner (from the fail-loud PR).

## Why I ran into this

I pass a sensitive variable into the module and it flows into the deploy `connection`. That turns
the whole `remote-exec` sensitive, so when `kubectl apply -k` failed I got nothing but
`(output suppressed…)` and couldn't tell what was wrong. Un-masking the connection was the only way
to read the actual error.

## Tradeoff — open to alternatives

This un-masks the connection unconditionally. I couldn't find a clean per-flag toggle because of how
marks propagate: a conditional `cond ? a : b` taints the result if *either* branch is sensitive, and
`try()` strips marks entirely — so a `debug ? nonsensitive(x) : x` gate ends up always sensitive.
Happy to gate it behind a variable or scope it differently if you'd prefer.

### DIFF
--- a/kustomization_user.tf
+++ b/kustomization_user.tf
@@ -61,13 +61,27 @@ resource "terraform_data" "kustomization_user_deploy" {
   }
 
   # Remove templates after rendering, and apply changes.
+  # `kubectl apply -k` runs in its own provisioner so that a non-zero exit fails this
+  # step directly. Previously it shared one remote-exec script with
+  # extra_kustomize_deployment_commands; since remote-exec has no `set -e` and the exit
+  # code is that of the LAST command, a failed apply was masked whenever a trailing
+  # command succeeded — tofu then reported success while nothing reconciled.
   provisioner "remote-exec" {
     # Debugging: "sh -c 'for file in $(find /var/user_kustomize -type f -name \"*.yaml\" | sort -n); do echo \"\n### Template $${file}.tpl after rendering:\" && cat $${file}; done'",
-    inline = compact([
+    inline = [
       "rm -f /var/user_kustomize/**/*.yaml.tpl",
       "echo 'Applying user kustomization...'",
       "kubectl apply -k /var/user_kustomize/ --wait=true",
-      var.extra_kustomize_deployment_commands
+    ]
+  }
+
+  # User-supplied post-apply commands run only after a successful apply. The leading
+  # "true" keeps inline non-empty when the variable is unset (compact() would otherwise
+  # produce an empty list).
+  provisioner "remote-exec" {
+    inline = compact([
+      "true",
+      var.extra_kustomize_deployment_commands,
     ])
   }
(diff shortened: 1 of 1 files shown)


mysticaltech:
Thanks for the careful writeup — but after a source-level review of Terraform (v1.10.1, v1.15.0) and OpenTofu (v1.11.6), we're declining this one.

**The premise doesn't hold for supported versions**: in `internal/terraform/node_resource_abstract_instance.go` (and the OpenTofu equivalent), connection info is evaluated and unmarked separately, and provisioner output suppression is driven only by sensitive marks on the **provisioner config**, not the `connection` block — with comments explicitly stating suppression should not apply to connection values. So unmasking the connection's SSH keys does not buy the claimed visibility.

What most likely bit you: a sensitive value flowing into the provisioner config itself (e.g. via `extra_kustomize_deployment_commands` or rendered template inputs). That's exactly what #2225 fixes — the apply now runs in its own provisioner, so its output stays visible even when the user-commands provisioner is suppressed. #2225 is merged for v2.20.1.

**Security assessment**: while we don't see a normal plan/state/stdout path that would print the key in this exact diff (terraform_data stores only input/output/triggers_replace; the SSH communicator logs `Private key: true`, not material), `nonsensitive()` deliberately removes UI redaction from raw private-key values. That adds secret-handling blast radius without demonstrated benefit.

If you can produce a minimal repro on a supported Terraform/OpenTofu version where a sensitive value used **only** in `connection` suppresses output, please share the transcript and we'll gladly revisit. Closing for now.

PR #2240, opened 2026-07-13
fix(agents): pin explicit private_ipv4 for primary-network agents

### LINKED ISSUE #2239: v3: agent nodes always get IP from last-created subnet, ignoring subnet_ip_range - regression from v2
**Module version:** `v3.0.0`

**Environment:** OpenTofu 1.11.3, Hetzner Cloud, `network_subnet_mode = "per_nodepool"`, `cni_plugin = "calico"`

**OS Image:** OpenSUSE Leap Micro 6.2 (kube-hetzner Packer snapshot)

---

## Background

The module attaches servers using only `network_id`, never `subnet_id`. Without an explicit `private_ipv4`, Hetzner picks the last-created subnet and assigns an IP from its range. In `per_nodepool` mode, the module creates the agent subnet (`10.0.0.0/16`) first and the control plane subnet (`10.255.0.0/16`) last. Because v3.0.0 no longer sets an explicit `private_ipv4` for agent nodes, Hetzner assigns all agent nodes an IP from the control plane subnet which is the opposite of what `subnet_ip_range` intends.

This behavior is documented in the [hcloud Terraform provider documentation](https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs/resources/server):

> "When used alone without `subnet_id`, the server will be attached to the last subnet (ordered by `ip_range`), which may be unpredictable."

And confirmed as a known API limitation in [hcloud provider issue #265](https://github.com/hetznercloud/terraform-provider-hcloud/issues/265):

> "This happens because the API does not support 'attaching to a subnet', it only supports attaching to a Network with an optional IP. This IP would be used to determine the subnet the server is put into."

---

## Root Cause

**v2.21.0 `agents.tf` line 40** - explicit private IP computed from the agent subnet:

```hcl
private_ipv4 = cidrhost(
  hcloud_network_subnet.agent[[for i, v in var.agent_nodepools : i if v.name == each.value.nodepool_name][0]].ip_range,
  each.value.index + (local.network_size >= 16 ? 101 : floor(pow(local.subnet_size, 2) * 0.4))
)
```

**v3.0.0 `agents.tf` line 85** - explicit private IP removed:

```hcl
private_ipv4 = null
```

Both `local.network_size` and `local.subnet_size` are still defined in v3.0.0 `locals.tf` (lines 224 and 227), so restoring the v2 expression is a drop-in fix with no missing dependencies.

---

## Reproduction
**Minimal `kube.tf`:**
```hcl
module "kube-hetzner" {
  version = "3.0.0"

  cluster_name = "cluster-a"
  base_domain  = "cluster-a.example.com"

  network_subnet_mode = "per_nodepool"

  control_plane_nodepools = [
    {
      name        = "control-plane"
      server_type = "cx23"
      location    = "fsn1"
      count       = 1
      labels      = []
      taints      = []
    }
  ]

  agent_nodepools = [
    {
      name            = "worker"
      server_type     = "cpx32"
      location        = "fsn1"
      count           = 2
      subnet_ip_range = "10.0.0.0/16"
      labels          = []
      taints          = []
    }
  ]

  ssh_public_key  = var.ssh_public_key
  ssh_private_key = var.ssh_private_key
}
```

---

## Expected Behavior

Agent nodes receive private IPs from their designated subnet (`10.0.0.x`) as defined by `subnet_ip_range = "10.0.0.0/16"`.

---

## Actual Behavior

**`hcloud server list` on v3.0.0 cluster  - all nodes including agents end up in the control plane subnet (`10.255.0.0/16`) despite `subnet_ip_range = "10.0.0.0/16"` set for agents:**

```
ID          NAME                            PRIVATE NET
111111111   cluster-a-control-plane-aaa     10.255.0.1 (cluster-a)
111111112   cluster-a-control-plane-bbb     10.255.0.3 (cluster-a)
111111113   cluster-a-control-plane-ccc     10.255.0.5 (cluster-a)
111111114   cluster-a-worker-aaa            10.255.0.2 (cluster-a)  ← should be 10.0.x.x
111111115   cluster-a-worker-bbb            10.255.0.4 (cluster-a)  ← should be 10.0.x.x
```

**`hcloud server list` on v2.20.0 cluster - agents are correctly placed in their designated subnet (`10.1.0.0/16`):**
```
222222221   cluster-b-control-plane-aaa     10.255.0.101 (cluster-b)  ← CP subnet ✅
222222222   cluster-b-control-plane-bbb     10.255.0.102 (cluster-b)  ← CP subnet ✅
222222223   cluster-b-worker-aaa            10.1.0.101 (cluster-b)    ← agent subnet ✅
222222224   cluster-b-worker-bbb            10.1.0.102 (cluster-b)    ← agent subnet ✅
```

---

## Verified Fix

We tested the fix by patching `private_ipv4` in `agents.tf` and redeploying the cluster.

**`kubectl get nodes -o wide` after fix:**

```
NAME                         STATUS   INTERNAL-IP   EXTERNAL-IP
cluster-a-control-plane-jal  Ready    10.255.0.1    <public-ip>
cluster-a-control-plane-qdj  Ready    10.255.0.3    <public-ip>
cluster-a-control-plane-qhl  Ready    10.255.0.2    <public-ip>
cluster-a-worker-eeg         Ready    10.0.0.101    <public-ip>  ← ✅ correct subnet
cluster-a-worker-jqe         Ready    10.0.0.102    <public-ip>   ← ✅ correct subnet
```

**`hcloud server list` after fix:**

```
ID          NAME                         PRIVATE NET
150234815   cluster-a-control-plane-jal  10.255.0.1 (cluster-a)   ✅
150234816   cluster-a-control-plane-qdj  10.255.0.3 (cluster-a)   ✅
150234817   cluster-a-control-plane-qhl  10.255.0.2 (cluster-a)   ✅
150234818   cluster-a-worker-eeg         10.0.0.101 (cluster-a)   ✅
150234819   cluster-a-worker-jqe         10.0.0.102 (cluster-a)   ✅
```

---

## Impact

- `subnet_ip_range` per agent nodepool has **no effect** on actual private IP assignment in v3
- IaC configuration does not reflect real infrastructure state
- Hetzner subnet isolation between control plane and agent nodes does not work in v3
- Terraform state file does not record `subnet_id` for nodes, making the inconsistency invisible without checking the Hetzner API directly
- Clusters migrating from v2 to v3 with `per_nodepool` and explicit `subnet_ip_range` silently lose subnet isolation

---

## Proposed Fix

Restore `cidrhost()` private IP assignment from v2 in `agents.tf` line 85:

```hcl
private_ipv4 = cidrhost(
  hcloud_network_subnet.agent[[for i, v in var.agent_nodepools : i if v.name == each.value.nodepool_name][0]].ip_range,
  each.value.index + (local.network_size >= 16 ? 101 : floor(pow(local.subnet_size, 2) * 0.4))
)
```

Both `local.network_size` and `local.subnet_size` are already defined in v3.0.0 `locals.tf`. Therefore, this is a one-line change with no additional dependencies.

boy51:
## Problem

`v3.0.0` sets `private_ipv4 = null` for agent nodes (`agents.tf`). The hcloud API attaches a server to a **Network** with an *optional* IP; it cannot attach to a specific subnet (hcloud provider limitation, [hetznercloud/terraform-provider-hcloud#265](https://github.com/hetznercloud/terraform-provider-hcloud/issues/265)). With no explicit IP, the server is assigned an address from the **last-created subnet** (ordered by `ip_range`) — which in `per_nodepool` mode is the control-plane subnet. Agent nodes therefore ignore their nodepool's `subnet_ip_range` and all land in the control-plane range.

Fixes #2239.

## Fix

Restore the v2 behavior: derive an explicit `private_ipv4` from the nodepool's agent subnet — reusing the exact subnet selection already present one line above in `ipv4_subnet_id` — offset by the node index (same `local.network_size` / `local.subnet_size` formula as v2). Guarded to primary-network nodes (`network_id == 0`); external-network nodepools keep API auto-assignment.

```hcl
private_ipv4 = each.value.network_id == 0 ? cidrhost(
  hcloud_network_subnet.agent[local.use_per_nodepool_subnets ? [for i, v in var.agent_nodepools : i if v.name == each.value.nodepool_name][0] : 0].ip_range,
  each.value.index + (local.network_size >= 16 ? 101 : floor(pow(local.subnet_size, 2) * 0.4))
) : null
```

## Notes

- Control-plane nodes carry the same `private_ipv4 = null` (`control_planes.tf`) but currently resolve to the correct range because the control-plane subnet sorts last. Left unchanged here to keep the fix minimal and scoped to #2239 — happy to add symmetric control-plane pinning if you'd prefer it in the same PR.
- `terraform fmt` clean; `terraform validate` passes on the module.
- The plan-matrix smoke test reads live provider data sources (needs a real Hetzner token), so I couldn't run it locally; CI should exercise it.

### DIFF
--- a/agents.tf
+++ b/agents.tf
@@ -82,7 +82,16 @@ module "agents" {
   network_id                    = local.agent_primary_network_id_by_node[each.key]
   primary_network_key           = each.value.network_id
   extra_network_ids             = local.agent_effective_extra_network_ids_by_node[each.key]
-  private_ipv4                  = null
+  # Pin an explicit private IPv4 for primary-network agents. The hcloud API
+  # attaches a server to a Network with an optional IP, not to a subnet; with
+  # no IP it lands in the last-created subnet (the control-plane range),
+  # ignoring the nodepool's subnet_ip_range (issue #2239). Mirror the
+  # ipv4_subnet_id selection above and offset by node index, matching v2.
+  # External-network nodepools (network_id != 0) keep API auto-assignment.
+  private_ipv4 = each.value.network_id == 0 ? cidrhost(
+    hcloud_network_subnet.agent[local.use_per_nodepool_subnets ? [for i, v in var.agent_nodepools : i if v.name == each.value.nodepool_name][0] : 0].ip_range,
+    each.value.index + (local.network_size >= 16 ? 101 : floor(pow(local.subnet_size, 2) * 0.4))
+  ) : null
 
   labels = merge(local.labels, local.labels_agent_node, each.value.hcloud_labels, { "kube-hetzner/os" = each.value.os })
 


mysticaltech:
Merged — thank you @boy51, this was an excellent PR on the trickiest kind of territory (IP allocation). Your root-cause was exactly right, the v2 formula restore was byte-exact (we proved v2 → v3.0.1 upgrades stay IP-no-ops), and the `network_id == 0` guard was correct against the internal normalization. On top of your commit we added: a collision-free dense cross-pool index for `network_subnet_mode = "shared"` (pool-local indexes could collide there), a plan-time capacity guard for small shared subnets, a render-harness contract locking the allocation expressions, and docs. Your commit is preserved as authored — you'll be in the v3.0.1 contributors.

We also verified the whole story **live**: deployed a v3.0.0 cluster (reproduced both agents landing in `10.255.0.x`), applied the fix, and watched them move to `10.0.0.101` / `10.1.0.101` — in-place NIC reattach, no server replacement, all nodes back `Ready` without manual intervention, follow-up plan converged clean. Ships in **v3.0.1** shortly. 🙏

PR #2242, opened 2026-07-19
fix: registries_config type error on plan when non-empty

### LINKED ISSUE #2241: [Bug]: Non-empty registries_config fails plan/apply with "Inconsistent conditional result types" (v3.0.0+, Terraform & OpenTofu)
### Description

Since **v3.0.0**, setting a non-empty `registries_config` makes `plan`/`apply` fail with `Error: Inconsistent conditional result types` — under **both** Terraform and OpenTofu (byte-identical error). This blocks the documented v2 → v3 in-place migration for anyone carrying a `k3s_registries` → `registries_config` value (private registries, pull-through mirrors, ...).

The failing expression is `locals.tf` line 199 (v3.0.1 and current `master`):

```hcl
registries_config_user = trimspace(var.registries_config) == "" ? {} : yamldecode(var.registries_config)
```

This is defined HCL behavior rather than an interpreter bug: a conditional's two result values must unify to a single type. `{}` has type `object({})`, while `yamldecode()` of the user's YAML yields a concrete type like `object({ mirrors = ..., configs = ... })`; object types with different attribute sets cannot unify, so evaluation fails. Terraform and OpenTofu share the same type system (cty), hence the identical failure.

Why it slipped through: the expression is new in v3.0.0 — v2.x passed `var.k3s_registries` through undecoded (`base64encode(var.k3s_registries)` into cloud-init), so any YAML shape worked; and with the default empty `registries_config` the failing branch is never evaluated, so plans without registries (including the v3 release-evidence upgrade) never hit it. `validate` also passes (variable value unknown there); it only surfaces at `plan`/`apply`.

**Standalone repro** (no providers, no module download — the expression copied verbatim):

```hcl
variable "registries_config" {
  type    = string
  default = <<-EOT
    mirrors:
      "registry.k8s.io":
        endpoint:
          - "http://10.43.0.10:5000"
    configs:
      "10.43.0.10:5000":
        tls:
          insecure_skip_verify: true
  EOT
}

locals {
  registries_config_user = trimspace(var.registries_config) == "" ? {} : yamldecode(var.registries_config)
}
```

`tofu plan` / `terraform plan` →

```
Error: Inconsistent conditional result types

The true and false result expressions must have consistent types. The 'false'
value includes object attribute "configs", which is absent in the 'true'
value.
```

**Proposed fix** — move the conditional inside `yamldecode()` so both branches are strings, which unify trivially; semantics are identical (empty/whitespace input still yields `{}`, downstream `try(local.registries_config_user.mirrors, null)` / `length(keys(...))` unaffected):

```hcl
registries_config_user = yamldecode(trimspace(var.registries_config) == "" ? "{}" : var.registries_config)
```

Verified against both empty and non-empty inputs on OpenTofu 1.12.1, and used (as a local patch) to complete a real v2.21.0 → v3.0.1 in-place migration of a live cluster — the standard upgrade path then applied cleanly. Note `... ? yamldecode("{}") : yamldecode(var.registries_config)` is **not** sufficient — the branches then have concrete, still-inconsistent object types and fail the same way. `try(yamldecode(var.registries_config), {})` would also work but silently swallows YAML syntax errors, so the string-conditional form seems preferable.

PR with the one-line fix incoming.

### Kube.tf file

```hcl
# Minimal relevant portion — any non-empty value triggers it:
module "kube-hetzner" {
  version = "3.0.1"
  # ...

  registries_config = <<-EOT
    mirrors:
      "registry.k8s.io":
        endpoint:
          - "http://10.43.0.10:5000"
    configs:
      "10.43.0.10:5000":
        tls:
          insecure_skip_verify: true
  EOT
}
```

### Screenshots

```
Error: Inconsistent conditional result types

  on .terraform/modules/kube-hetzner/locals.tf line 199, in locals:
 199:   registries_config_user = trimspace(var.registries_config) == "" ? {} : yamldecode(var.registries_config)
    ├────────────────
    │ var.registries_config is "mirrors:\n  \"registry.k8s.io\":\n    endpoint:\n      - \"http://10.43.0.10:5000\"\nconfigs:\n  \"10.43.0.10:5000\":\n    tls:\n      insecure_skip_verify: true\n"

The true and false result expressions must have consistent types. The 'false'
value includes object attribute "configs", which is absent in the 'true'
value.
```

### Platform

Linux (OpenTofu v1.12.1 and Terraform, identical result)

prochac:
Fixes #2241

## Problem

Since v3.0.0, any non-empty `registries_config` fails `plan`/`apply` on both Terraform and OpenTofu:

```
Error: Inconsistent conditional result types

The true and false result expressions must have consistent types. The 'false'
value includes object attribute "configs", which is absent in the 'true' value.
```

`locals.tf`'s `cond ? {} : yamldecode(var.registries_config)` requires both conditional branches to unify to one type, and `object({})` cannot unify with the concrete object type produced by `yamldecode()` of the user's YAML. This is defined HCL/cty behavior, identical in both tools, and blocks the v2 → v3 in-place migration for anyone carrying a `k3s_registries` → `registries_config` value. (v2 never decoded the string, which is why it only surfaces in v3.)

## Fix

Move the conditional inside `yamldecode()` so both branches are strings, which unify trivially:

```hcl
registries_config_user = yamldecode(trimspace(var.registries_config) == "" ? "{}" : var.registries_config)
```

Semantics are unchanged: empty/whitespace input still yields `{}` (`yamldecode("{}")` is an empty object), non-empty input decodes exactly as before, and the downstream `try(local.registries_config_user.mirrors, null)` / `length(keys(...))` usages are unaffected. Malformed YAML still errors loudly (unlike a `try(yamldecode(...), {})` variant, which would silently swallow it).

## Verification

- Standalone repro of the expression (in #2241): fails before, plans cleanly after, for both empty and non-empty inputs, on OpenTofu 1.12.1 and Terraform.
- Used as a local patch to complete a real v2.21.0 → v3.0.1 in-place migration of a live 3-node MicroOS cluster with a non-empty registries mirror config: the standard upgrade path then applied with the documented no-destroy plan gate passing and the re-encoded `registries.yaml` rolling out cleanly (semantic no-op, k3s service restarts as documented).
- `tofu fmt -check` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -196,7 +196,10 @@ locals {
   # Check if the user has set custom DNS servers.
   has_dns_servers = length(var.dns_servers) > 0
 
-  registries_config_user = trimspace(var.registries_config) == "" ? {} : yamldecode(var.registries_config)
+  # The conditional lives inside yamldecode() so both branches are strings:
+  # `cond ? {} : yamldecode(...)` fails type unification (object({}) vs the
+  # decoded object type) for any non-empty registries_config.
+  registries_config_user = yamldecode(trimspace(var.registries_config) == "" ? "{}" : var.registries_config)
   embedded_registry_mirror_registries = var.embedded_registry_mirror.enabled ? [
     for registry in var.embedded_registry_mirror.registries : registry
   ] : []


mysticaltech:
Thank you @prochac. I pulled your `registries_config` type fix into the integrated v3.1.0 intake candidate, with authorship preserved: https://github.com/mysticaltech/terraform-hcloud-kube-hetzner/pull/2253. The candidate also has the full Terraform/OpenTofu/render/contract/smoke gates recorded there.

mysticaltech:
Thank you @prochac. Your fix was integrated through #2253 and is now on `master`; the original commit authorship is preserved so you get release contributor credit. I added maintainer-side release hardening and broader validation around it before merging the v3.1.0 intake branch.

PR #2170, opened 2026-03-28
feat: add dual-stack IPv6 networking support

mgazza:
## Summary

Adds optional dual-stack IPv6 support to K3s clusters on Hetzner Cloud. When the new `cluster_ipv6_cidr` and `service_ipv6_cidr` variables are set, the module configures the full stack for dual-stack networking. Leaving them empty (the default) preserves existing single-stack IPv4 behavior with zero changes.

### What it does

- Configures K3s with dual-stack `--cluster-cidr`, `--service-cidr`, and `--node-ip` flags
- Enables Cilium IPv6 (`ipv6.enabled: true`, `requireIPv6PodCIDR: true`)
- Sets `HCLOUD_INSTANCES_ADDRESS_FAMILY=dualstack` on Hetzner CCM so it advertises both IPv4 and IPv6 node addresses (ref: [hcloud-cloud-controller-manager#803](https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/803))
- Adds CCM toleration for `node.cilium.io/agent-not-ready` taint to break the chicken-and-egg deadlock between Cilium (needs IPv6 node addresses from CCM) and CCM (can't schedule because Cilium taint blocks it)
- Increases bootstrap wait timeouts (system-upgrade-controller 900s→1800s, namespace deployments 300s→600s) to accommodate Cilium restart loops during dual-stack CCM convergence

### Usage

```hcl
module "kube-hetzner" {
  source = "kube-hetzner/kube-hetzner/hcloud"

  # Existing IPv4 config (unchanged)
  cluster_ipv4_cidr = "10.42.0.0/16"
  service_ipv4_cidr = "10.43.0.0/16"

  # NEW: Enable dual-stack by setting IPv6 CIDRs
  cluster_ipv6_cidr = "fd00:42::/48"
  service_ipv6_cidr = "fd00:43::/112"

  # ... rest of config
}
```

### Testing

Tested with a 2-node cluster (1 control-plane cx23, 1 agent cx23) in Hetzner Cloud nbg1 with Cilium CNI on k3s v1.34.5:

- Clean `terraform apply` from scratch completes without intervention (29 resources)
- All pods reach Running state (Cilium, CCM, CSI, CoreDNS, system-upgrade-controller, kured)
- Both nodes get dual-stack pod CIDRs (e.g., `10.42.0.0/24` + `fd00:42::/64`)
- Both nodes get IPv6 ExternalIP addresses from Hetzner CCM
[...]

PR #2245, opened 2026-07-23
feat: dual-stack node-ip for control-plane and agent nodes

### LINKED ISSUE #2244: [Bug]: cluster_ipv6_cidr / service_ipv6_cidr are unusable — control-plane node-ip stays IPv4-only, k3s refuses to start
## Description

Setting `cluster_ipv6_cidr` and `service_ipv6_cidr` on a standard cluster
(`node_transport_mode = "hetzner_private"`, `multinetwork_mode = "disabled"`)
renders dual-stack `cluster-cidr` and `service-cidr`, but leaves `node-ip`
IPv4-only. k3s then aborts on startup:

```
level=fatal msg="Error: cluster-cidr: [10.42.0.0/16 fd00:42::/56] and node-ip: [10.255.0.101], must share the same IP version (IPv4, IPv6 or dual-stack)"
```

The module's config-update provisioner catches this, restores the previous
`config.yaml` from backup and restarts k3s, so the cluster stays healthy. But the
two variables cannot be used at all on this topology.

**Root cause:** control-plane `node-ip` is hardcoded to the private IPv4 —
`control_planes.tf:382` (k3s), `control_planes.tf:441` (RKE2), and `init.tf:112` /
`init.tf:281` (first control plane). A dual-stack `node-ip` is only built for
**agents**, and only under `multinetwork_overlay_enabled` (`agents.tf:146,172`),
i.e. the experimental `cilium_public_overlay` preview. On the standard
private-network topology no node advertises an IPv6 `node-ip`, so the dual-stack
CIDRs are unusable. Enabling `cilium_public_overlay` hits the same fatal, because
it adds `node-external-ip` (a different flag), not `node-ip`.

**Why it exists (half-landed feature):** #2111 added the variables; #2170
implemented the `--node-ip` wiring but was closed unmerged (its subnet rewrite was
too risky for v3 intake); #2185 carried forward the Cilium/CCM pieces but contains
no `node-ip` change (`gh pr diff 2185 | grep node-ip` is empty). No open issue/PR
currently tracks the remainder.

A PR with the minimal fix (the `node-ip` portion of #2170, rebased onto v3,
without the subnet rewrite) is ready and live-tested.

## Kube.tf file

```terraform
module "kube-hetzner" {
  version = "3.0.1"

  providers    = { hcloud = hcloud }

  ssh_public_key  = file("~/.ssh/id_ed25519.pub")
  ssh_private_key = null

  network_region = "us-west"

  control_plane_nodepools = [
    { name = "control-plane", server_type = "cpx31", location = "hil", labels = [], taints = [], count = 3 }
  ]
  agent_nodepools = []

  cni_plugin = "cilium"

  cluster_ipv6_cidr = "fd00:42::/56"
  service_ipv6_cidr = "fd00:43::/112"
}
```

## Screenshots

k3s journal on the first control plane during `terraform apply`:

```
level=fatal msg="Error: cluster-cidr: [10.42.0.0/16 fd00:42::/56] and node-ip: [10.255.0.101], must share the same IP version (IPv4, IPv6 or dual-stack)"
```

Reproduces on module v3.0.1 and on current `master` (the `node-ip` lines are
unchanged there). k3s v1.33.13+k3s1, Cilium tunnel mode, openSUSE MicroOS, single
Hetzner Network, 3 control planes, no agents, no NAT router.

## Platform

Linux

bkero:
## Summary

Makes `cluster_ipv6_cidr` / `service_ipv6_cidr` actually usable on the standard
`hetzner_private` topology by advertising a dual-stack `node-ip`.

Fixes #2244.

Today those variables render dual-stack `cluster-cidr` and `service-cidr` while
`node-ip` stays IPv4-only, so k3s/RKE2 refuse to start:

```
cluster-cidr: [10.42.0.0/16 fd00:42::/56] and node-ip: [10.255.0.101],
must share the same IP version (IPv4, IPv6 or dual-stack)
```

## What changed

- New `control_plane_node_ip_by_node` / `agent_node_ip_by_node` locals that append
  the node's public IPv6 when `local.cluster_has_ipv6` is set, and pass through
  `private_ipv4_address` otherwise.
- Used at all six `node-ip` sites: `control_planes.tf:382,441`, `init.tf:112,281`,
  `agents.tf:146,172`. The agent `multinetwork_overlay_enabled` branch is untouched.
- A validation precondition requiring public IPv6 on all control-plane and agent
  nodes when the IPv6 CIDRs are set, so a missing address fails with a clear
  message instead of the opaque k3s error above.

Hetzner Cloud Networks are IPv4-only, so the public address is the only IPv6
available for `node-ip`. `advertise-address` is deliberately left as the private
IPv4 — it is single-valued and the apiserver advertisement stays on the private
network.

Diff: 5 files, +52/-6.

## Relationship to #2170

This is the `node-ip` portion of #2170, rebased onto the v3 locals
(`cluster_has_ipv6` rather than that PR's `enable_dualstack`), and **without the
subnet rewrite** that was the stated reason #2170 was not merged. It also drops
that PR's cloud-init runtime IP-detection, because the addresses are already known
to Terraform for control planes and static agents. Credit to @mgazza for the
original work in #2170.

Scope is intentionally narrow: no autoscaler changes, no subnet changes, no Cilium
or CCM changes (those already landed via #2185).

## Default behavior

Unchanged. `cluster_has_ipv6` is false unless `cluster_ipv6_cidr` is set, so every
existing cluster renders exactly the same `node-ip` as before. The new validation
only triggers when the IPv6 CIDRs are set.

## Testing

- `terraform fmt -check -recursive` — clean
- `terraform init -backend=false && terraform validate` — passes
- Single-stack path unchanged: with `cluster_ipv6_cidr` unset the locals return
  `private_ipv4_address` exactly as before.

### Live-tested on a real cluster

Applied to a production v3.0.1 cluster (k3s v1.33.13+k3s1, Cilium tunnel mode, 3
control planes, single Hetzner Network, MicroOS). Before the patch the same config
aborted with the fatal error; with it, all three control planes restarted k3s
cleanly and rendered:

```
"cluster-cidr": "10.42.0.0/16,fd00:42::/56"
"service-cidr": "10.43.0.0/16,fd00:43::/112"
"node-ip":      "10.255.0.101,2a01:4ff:1f0:dbe3::1"
```

Nodes returned Ready, etcd healthy, Cilium picked up `enable-ipv6=true`, and pods
received dual-stack addresses (e.g. `10.42.x.y` + `fd00:42::…`). Cross-node
IPv4/IPv6 pod networking and service DNS verified working.

### Migration caveat for existing clusters (documented, not a code issue)

`node.spec.podCIDRs` is assigned at node registration and is immutable
(`spec.podCIDRs: Forbidden: node updates may not change podCIDR except from "" to
valid`). A **new** cluster registers nodes against the dual-stack `cluster-cidr`
and gets both pod CIDRs. An **existing** cluster additionally needs its nodes
re-registered so the controller-manager reallocates `podCIDRs`; until then Cilium
waits on `required IPv6 PodCIDR not available`. This is inherent to enabling
dual-stack on a live cluster, not something this patch can or should solve — it
just needs a note in the docs (happy to add one).

## Open questions

1. `advertise-address` left as private IPv4 — correct, or should it follow?
2. Autoscaler nodepools are out of scope here; they need runtime detection since
   addresses aren't known at plan time (#2170 used a cloud-init hook). Same PR or
   separate?
3. Requiring public IPv6 on all nodes — acceptable, or should there be an escape
   hatch for per-nodepool `enable_public_ipv6 = false`?

### DIFF
--- a/agents.tf
+++ b/agents.tf
@@ -143,7 +143,7 @@ locals {
       # Kubelet arg precedence (last wins): local.kubelet_arg < global_kubelet_args < agent_kubelet_args < v.kubelet_args
       kubelet-arg   = local.agent_effective_kubelet_args_by_node[k]
       flannel-iface = local.flannel_iface
-      node-ip       = local.multinetwork_overlay_enabled ? join(",", compact([local.multinetwork_transport_ipv4_enabled ? module.agents[k].ipv4_address : null, local.multinetwork_transport_ipv6_enabled ? module.agents[k].ipv6_address : null])) : module.agents[k].private_ipv4_address
+      node-ip       = local.multinetwork_overlay_enabled ? join(",", compact([local.multinetwork_transport_ipv4_enabled ? module.agents[k].ipv4_address : null, local.multinetwork_transport_ipv6_enabled ? module.agents[k].ipv6_address : null])) : local.agent_node_ip_by_node[k]
       node-label    = v.labels
       node-taint    = v.taints
     },
@@ -169,7 +169,7 @@ locals {
       token     = local.cluster_token
       # Kubelet arg precedence (last wins): local.kubelet_arg < global_kubelet_args < agent_kubelet_args < v.kubelet_args
       kubelet-arg = local.agent_effective_kubelet_args_by_node[k]
-      node-ip     = local.multinetwork_overlay_enabled ? join(",", compact([local.multinetwork_transport_ipv4_enabled ? module.agents[k].ipv4_address : null, local.multinetwork_transport_ipv6_enabled ? module.agents[k].ipv6_address : null])) : module.agents[k].private_ipv4_address
+      node-ip     = local.multinetwork_overlay_enabled ? join(",", compact([local.multinetwork_transport_ipv4_enabled ? module.agents[k].ipv4_address : null, local.multinetwork_transport_ipv6_enabled ? module.agents[k].ipv6_address : null])) : local.agent_node_ip_by_node[k]
       node-label  = v.labels
       node-taint  = v.taints
     },
--- a/control_planes.tf
+++ b/control_planes.tf
@@ -379,7 +379,7 @@ locals {
       kubelet-arg                 = local.control_plane_effective_kubelet_args_by_node[k]
       kube-apiserver-arg          = concat(local.kube_apiserver_arg, var.enable_secrets_encryption ? ["encryption-provider-config=${local.secrets_encryption_config_file}"] : [])
       kube-controller-manager-arg = local.kube_controller_manager_arg
-      node-ip                     = module.control_planes[k].private_ipv4_address
+      node-ip                     = local.control_plane_node_ip_by_node[k]
       advertise-address           = module.control_planes[k].private_ipv4_address
       node-label                  = v.labels
       node-taint                  = v.taints
@@ -438,7 +438,7 @@ locals {
       kube-apiserver-arg          = concat(local.kube_apiserver_arg, var.enable_secrets_encryption ? ["encryption-provider-config=${local.secrets_encryption_config_file}"] : [])
       kube-controller-manager-arg = local.kube_controller_manager_arg
       flannel-iface               = local.flannel_iface
-      node-ip                     = module.control_planes[k].private_ipv4_address
+      node-ip                     = local.control_plane_node_ip_by_node[k]
       advertise-address           = module.control_planes[k].private_ipv4_address
       node-label                  = v.labels
       node-taint                  = v.taints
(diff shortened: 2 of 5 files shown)


mysticaltech:
Thank you @bkero. Your static dual-stack `node-ip` work is included in the integrated v3.1.0 intake candidate, with authorship preserved: https://github.com/mysticaltech/terraform-hcloud-kube-hetzner/pull/2253. I added release-gate hardening around the unsafe combinations we found during review: standard-path active autoscaler, NAT router, Robot nodes, Tailscale transport, IPv6-only standard path, Cilium native routing, missing public IPv6, and public-overlay family mismatches now fail at plan time instead of booting badly.

mysticaltech:
Thank you @bkero. The safe static-node dual-stack `node-ip` portion was incorporated through #2253 and is now on `master`; your commit authorship is preserved in the release history. I kept the wider subnet/topology rewrite out of this release train and added plan-time guards for the combinations that are not live-proven yet.

PR #2251, opened 2026-07-31
fix(microos): grant sshd read access to shadow_t so pubkey auth works under SELinux enforcing

### LINKED ISSUE #2250: Cluster Autoscaler-provisioned MicroOS nodes lose SSH access after cloud-init runs (v3, latest CA image + latest MicroOS packer snapshot)
### Description

Manually creating a server directly from our MicroOS packer snapshot works fine — root SSH via pubkey succeeds immediately, no lockout. But when the **Cluster Autoscaler** provisions a node into an `autoscaler_nodepools` group (i.e. it fires a new machine and applies its own cloud-init), SSH to that node breaks. Once broken, it stays broken — every subsequent `terraform apply` fails against that node because Terraform can no longer SSH to it either.

This is new behavior with the latest v3 module + latest Cluster Autoscaler image + latest MicroOS packer snapshot. Same packer snapshot, manually booted (no CA, no cloud-init from CA), does not exhibit the problem. Packer template is the stock/default `hcloud-microos-snapshots.pkr.hcl` from this repo, unmodified.

We've only tested MicroOS so far, not Leap Micro — can't say whether Leap Micro nodepools are affected by the same issue.

### Environment

- Module version: `v3.0.1`
- Kubernetes distribution: k3s (`v1.31.14+k3s1`)
- Packer template: MicroOS (`hcloud-microos-snapshots.pkr.hcl`, default/unmodified, current upstream defaults)
- Autoscaler nodepool server type: `cpx22`
- `cluster_autoscaler_extra_args` includes `--enforce-node-group-min-size=true`, `--scale-down-unneeded-time=3m`, `--scale-down-delay-after-add=1m`

### Steps to reproduce

1. Build a MicroOS snapshot with the current upstream packer template.
2. Boot a server directly from that snapshot (outside of kube-hetzner/CA) → confirm SSH pubkey auth works.
3. Deploy a cluster with an `autoscaler_nodepools` entry using that snapshot.
4. Trigger the Cluster Autoscaler to scale up (e.g. schedule a pod that doesn't fit).
5. Wait for CA to provision the new node and apply its cloud-init.
6. Attempt SSH to the new node with the configured key.

### Expected

SSH to the CA-provisioned node succeeds, same as a manually-booted node from the same snapshot.

### Actual

SSH is rejected (`Permission denied (publickey,keyboard-interactive)` — a clean rejection, not `MaxAuthTries`/agent key-spray noise). The node is otherwise healthy and running. Terraform can no longer reach it, so any subsequent `apply` that touches that nodepool fails at the SSH-provisioning step.

### Suspected cause

We believe this isolates the bug specifically to **Cluster Autoscaler's own cloud-init payload**, as distinct from the cloud-init Terraform applies to statically-managed nodepools. Our working theory:

- MicroOS/transactional-update's `/etc` overlay finalization (`tukit`) can silently restore a locked root shadow entry (`root:!...`) even after the base image's own unlock step runs during the packer build's `transactional-update shell`.
- If Cluster Autoscaler's cloud-init template re-runs a `passwd`/`lock_passwd` step (e.g. via `cloud-init`'s default `lock_passwd: true` behavior) that a Terraform-provisioned node's cloud-init does not, that would explain why manually-booted and Terraform-provisioned nodes are unaffected while CA nodes are.

### Additional context

Happy to provide full `cloud-init.log`/`shadow` state from an affected node via rescue-mode inspection if useful.

urashidmalik:
## Summary

Fixes #2250 — SSH pubkey auth denied on Cluster Autoscaler-fired MicroOS nodes.

### Root cause

The stock `selinux-policy`/`selinux-policy-targeted` `ssh` module ships no `allow` rule granting `sshd_t`/`sshd_auth_t`/`sshd_session_t` read access to `shadow_t`. Under SELinux enforcing mode, once OpenSSH's split privsep daemons (`sshd-auth`, `sshd-session` — OpenSSH 9.8+) need to read `/etc/shadow`, the read is denied. This surfaces as sshd logging "account is locked" for a perfectly valid, unlocked account, with a matching AVC denial for a `shadow_t` read (`sesearch --allow -s sshd_session_t -t shadow_t -c file -p read` returns nothing on an unpatched snapshot).

A manually booted node from the snapshot can appear unaffected depending on timing/session type, but Cluster Autoscaler-fired nodes hit this immediately and permanently, because CA nodes get no Terraform-side SSH retry/remediation the way static nodepools do (`terraform_data.initial_readiness` / `ssh_authorized_keys` only apply to `hcloud_server` resources Terraform manages directly — CA's own `--cloud-provider=hetzner` binary provisions nodes out-of-band).

### Fix

`hcloud-microos-snapshots.pkr.hcl` now bakes a small local SELinux policy module (`sshd_shadow_read`) into the snapshot during `install_packages`, granting exactly the missing `getattr`/`open`/`read` rules on `shadow_t` for `sshd_t`, `sshd_auth_t`, and `sshd_session_t`. `clean_up` now also verifies the module is active (`sesearch --allow ...`) and fails the packer build (`FATAL`) if it isn't, so a broken fix can't silently ship in a snapshot.

### Verification

- Rebuilt the MicroOS x86 snapshot with this change; the build-time `FATAL` guardrail passed.
- Re-applied Terraform so Cluster Autoscaler's node group config referenced the new snapshot.
- Deleted a live CA-managed node and let the autoscaler fire a replacement from the new snapshot — clean SSH pubkey auth.
- Repeated the cycle a second time (delete + let CA refire) to rule out a one-off race — clean SSH pubkey auth again, on a different node with a different IP.

Both cycles produced a clean `Authentications that can continue: publickey` success with no denial, confirmed via `ssh -vvv`.

### Scope

Only touches `hcloud-microos-snapshots.pkr.hcl` (the Leap Micro template does not use this SELinux policy build and is unaffected) and adds a matching `CHANGELOG.md` entry under `[Unreleased]`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-_No unreleased changes._
+### 🐛 Bug Fixes
+
+- **MicroOS: SSH pubkey auth denied on Cluster Autoscaler-fired nodes** - The stock `selinux-policy`/`selinux-policy-targeted` `ssh` module ships no `allow` rule granting `sshd_t`/`sshd_auth_t`/`sshd_session_t` read access to `shadow_t`, so every login is denied under SELinux enforcing mode once OpenSSH's split privsep daemons (`sshd-auth`, `sshd-session`, OpenSSH 9.8+) need to read `/etc/shadow` — visible as sshd logging "account is locked" for a valid, unlocked account, with a matching AVC denial for a `shadow_t` read. Cluster Autoscaler-fired nodes get no Terraform-side SSH retry/remediation the way static nodepools do, so they surface the gap immediately and permanently, while a manually booted node from the same snapshot can look fine. `hcloud-microos-snapshots.pkr.hcl` now bakes a small local SELinux policy module (`sshd_shadow_read`) closing the gap, with a build-time verification step that fails the packer build if the module isn't active in the snapshot. Verified live: rebuilt the MicroOS x86 snapshot, then forced two independent Cluster Autoscaler node replacement cycles — both new nodes authenticated via SSH pubkey cleanly. Fixes #2250.
 
 ---
 
--- a/packer-template/hcloud-microos-snapshots.pkr.hcl
+++ b/packer-template/hcloud-microos-snapshots.pkr.hcl
@@ -88,13 +88,34 @@ locals {
   # Only install kernel-longterm if selected; kernel-default is already in the base image
   kernel_package_list = var.kernel_type == "longterm" ? ["kernel-longterm"] : []
 
-  needed_packages = join(" ", concat(local.kernel_package_list, ["restorecond", "policycoreutils", "policycoreutils-python-utils", "setools-console", "audit", "bind-utils", "wireguard-tools", "fuse", "open-iscsi", "nfs-client", "xfsprogs", "cryptsetup", "lvm2", "git", "cifs-utils", "bash-completion", "mtr", "tcpdump", "udica", "qemu-guest-agent"], var.packages_to_install))
+  needed_packages = join(" ", concat(local.kernel_package_list, ["restorecond", "policycoreutils", "policycoreutils-python-utils", "setools-console", "checkpolicy", "audit", "bind-utils", "wireguard-tools", "fuse", "open-iscsi", "nfs-client", "xfsprogs", "cryptsetup", "lvm2", "git", "cifs-utils", "bash-completion", "mtr", "tcpdump", "udica", "qemu-guest-agent"], var.packages_to_install))
 
   # Read sysctl config if file path is provided, otherwise empty (base64 encoded for safe transfer)
   sysctl_config_content = var.sysctl_config_file != "" ? base64encode(file(var.sysctl_config_file)) : ""
 
   # Commands to write sysctl config if provided (decode base64)
   sysctl_commands = local.sysctl_config_content != "" ? "echo '${local.sysctl_config_content}' | base64 -d > /etc/sysctl.d/99-custom.conf" : ""
+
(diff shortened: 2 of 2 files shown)


mysticaltech:
Thank you @urashidmalik. I investigated this seriously while preparing the v3.1.0 intake. I did not merge the SELinux shadow-read policy change because the evidence points to a safer lower-privilege root cause fix: restoring openSUSE's vendor `UsePAM yes` path instead of granting sshd direct shadow reads. That fix is in the integrated candidate here: https://github.com/mysticaltech/terraform-hcloud-kube-hetzner/pull/2253. If you can retest against that candidate and still see AVC denials, please share the fresh `ausearch -m avc -ts boot` output and we can revisit the policy with concrete evidence.

mysticaltech:
Thank you @urashidmalik for the SELinux policy PR. I am keeping this open for now rather than merging it blind: #2253 has landed a lower-privilege SSH/PAM fix that may address the same failure mode without granting `sshd_t` read access to `shadow_t`. The next step is to retest your autoscaler-created MicroOS node case against current `master` / the v3.1.0 release candidate and only keep this policy patch if the AVC evidence still points to `shadow_t` after that fix.

mysticaltech:
Thank you @urashidmalik for the careful investigation, the proposed policy, and the live reproduction work. The exact autoscaler retest now shows that the broader SELinux rule is unnecessary after #2252/#2253: key-only SSH succeeds under Enforcing mode without the local module and without any `shadow_t` AVC.

Because this PR would grant SSH domains direct read access to `/etc/shadow` and add a build dependency/local policy where the vendor PAM path now works, we are closing it as superseded rather than merging it. Your report was valuable and led directly to the lower-privilege fix and this dedicated release-candidate autoscaler test. Thank you for helping us harden v3.1.0.

urashidmalik:
Glad we could help, lets go for lower-privilege fix 🚀

PR #2263, opened 2026-08-10
fix(rke2): restore SELinux context for encryption-config

antony-jr:
Ensure /etc/rancher/rke2/encryption-config.yaml receives the correct SELinux context after being moved from /tmp.

Without relabeling, the file can retain the user_tmp_t context, causing kube-apiserver running under rke2_service_t to fail with permission denied when loading the encryption provider configuration.

This prevents first control-plane bootstrap from hanging when SELinux is enforcing and enable_secrets_encryption is enabled.

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -641,7 +641,7 @@ locals {
       # move the config file into place and adjust permissions
       "[ -f /tmp/config.yaml ] && mv /tmp/config.yaml /etc/rancher/rke2/config.yaml",
       "chmod 0600 /etc/rancher/rke2/config.yaml",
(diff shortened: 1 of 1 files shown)


mysticaltech:
Thank you, @antony-jr. Your SELinux diagnosis and original fix are now merged into `master` for v3.2.0, with your commit and authorship preserved.

We built on it by making the whole Secrets EncryptionConfiguration path root-only and role-aware, rejecting unsafe in-place key replacement or disablement, and adding cleanup and rollback on failure. The final MicroOS/RKE2 canary exercised enforcing SELinux on both roles, preserved the labels across real agent and control-plane reboots, kept RKE2 in `container_runtime_t` and CSI in `spc_t`, reported matching encryption hashes, and produced zero AVCs. Your focused contribution was the starting point for that complete repair.

